PR #1, opened 2021-06-03
Init API defs

njhale:
Add initial go structs defining rukpak's core APIs:

- [x] `Bundle`
- [x] `ProvisionerClass`
- [ ] `Instance`

See [the strawman](https://hackmd.io/hppR60SiRGKPcqKQuByvVg?view#RukPak) for the high level design being followed.


api/v1alpha1/bundle_types.go
```diff
@@ -0,0 +1,80 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type BundleSourceType string
+
+// +union
+type BundleSource struct {
+	// +unionDiscriminator
+	// +optional
+	Type BundleSourceType
```
ecordell, line 29: What does the `type` tell us that we can't derive from the `class`? Do you have examples in mind?

api/v1alpha1/bundle_types.go
```diff
@@ -0,0 +1,80 @@
...
+
+	// +optional
+	Ref string
```
ecordell, line 32: At least some of the strawmen had this as a list of `refs` (for e.g. a set of kustomize layers, or a helm chart + a config, etc)

If this is a single string instead of an array, is there another idea for addressing "bundles" that require multiple inputs like that?

ecordell (commented):
the early bits are looking good!

api/v1alpha1/bundle_types.go
njhale: The idea was to make room for sources of bundle content that we couldn't express as a URI-like reference; e.g. embedded yaml or maybe volume mounts.

api/v1alpha1/bundle_types.go
timflannagan: IIRC the rationale earlier today was to start with something on a smaller scale and scale up to an array of references once we get a feel for how these APIs will interact and be consumed to get a couple of prototypes out the door quicker.

ecordell (approved):

PR #3, opened 2021-07-12
Introduce a sanity github action check

timflannagan:
Introduce the .github/workflows directory and add a action that ensures
that the codegen (manifests/APIs) are up-to-date and the package
formatting (go fmt/goimports) don't produce a diff.

Update the Makefile's format target and also run the goimports command
against the set of Go packages

### DIFF
--- a/.github/workflows/sanity.yaml
+++ b/.github/workflows/sanity.yaml
@@ -1,0 +1,20 @@
+name: sanity
+
+on:
+  push:
+    branches:
+      - '**'
+  pull_request:
+    paths:
+      - '**'
+
+jobs:
+  sanity:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '~1.16'
+      - name: Run verification checks
+        run: make verify
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@
 
 # Dependency directories
 vendor/
+bin/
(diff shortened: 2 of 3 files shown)


.github/workflows/sanity.yaml
```diff
@@ -0,0 +1,22 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '~1.16'
+      - name: Run formatting checks
+        run: make vendor && make format && git diff --exit-code
+      - name: Ensure codegen is up-to-date
+        run:  make generate && make manifests && git diff --exit-code
```
njhale, line 22: can we use the `verify` make target for this instead?

how about adding specific targets that verify each thing we care about; e.g. `codegen`, `manifests`, `goimports`, etc?

.github/workflows/sanity.yaml
```diff
@@ -0,0 +1,22 @@
...
+jobs:
+  sanity:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '~1.16'
+      - name: Run formatting checks
+        run: make vendor && make format && git diff --exit-code
```
njhale, line 20: do we really care about the vendor directory for `rukpak`? I see the `vendor` target calls `tidy`, which we probably do care about -- can we factor it out?

njhale (changes requested):

.github/workflows/sanity.yaml
timflannagan: Nope just an oversight - we can only specify `tidy` for now.

njhale (approved):

PR #4, opened 2021-07-14
Introduce a sample rukpak.io provisioner

timflannagan:
Builds off the types definitions introduced in #1 for the Bundle and ProvisionerClass APIs.

Introduces a sample provisioner that currently handles scratch container images containing arbitrary content in a flattened file system. Those contents are unpacked, persisted to a file system (via PVCs/PVs), and served via a simply file system server.

WIP/POC -- mainly a rough baseline of code I had sitting around locally that might be useful to build off of going forward.


api/v1alpha1/bundle_types.go
```diff
@@ -17,34 +17,65 @@ limitations under the License.
...
+	Conditions []metav1.Condition           `json:"conditions,omitempty"`
+	Unpacked   BundleUnpackStatusType       `json:"unpacked,omitempty"`
+	Volume     *corev1.LocalObjectReference `json:"volume"`
+	URI        string                       `json:"uri,omitempty"`
 }
 
+type BundleUnpackStatusType string
+
+const (
+	BundleUnpacked       BundleUnpackStatusType = "Unpacked"
```
exdx, line 67: I think these status types might be overly specified to the existing olm bundle type -- for example, if the bundle API is used for a helm chart or a git repository, would we call pulling them from remote to a PV on-cluster "unpacking" in the same way as we do for olm based bundles? I'm wondering if there are some more generic terms that we could use to describe this behavior. 

We have "unpack" a bundle versus "install" a helmchart versus "clone" a repository essentially. 

Maybe we could use the term `Contents` as the top level field on the Bundle.Status and then the term `Available` to describe that a bundle has been fully unpacked and is available on cluster. `InProgress` to mean that the content unpacking is in progress. And `NotAvailable` by default when the bundle is created.

cmd/serve/main.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"fmt"
+	"net/http"
+	"os"
+
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+const port = ":8081"
```
exdx, line 12: nit: should this be a constant or a command line argument, like the directory path, defaulting to 8081?

pkg/k8s/provisioner/provisioner_controller.go
```diff
@@ -0,0 +1,128 @@
...
+			return nil, err
+		}
+
+		pvc.SetName(pvcName)
+		pvc.SetNamespace(r.globalNamespace)
+		volumeMode := corev1.PersistentVolumeFilesystem
+
+		pvc.Spec = corev1.PersistentVolumeClaimSpec{
+			VolumeMode:  &volumeMode,
+			AccessModes: []corev1.PersistentVolumeAccessMode{corev1.ReadWriteOnce},
```
exdx, line 115: I know we are still thinking about this, but is there a reason we don't use `ReadOnlyMany` AccessMode [1] for the PVC? Otherwise we can't share the content among multiple nodes. 

[1] https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes

pkg/k8s/provisioner/provisioner_controller.go
```diff
@@ -0,0 +1,128 @@
...
+
+	return nil
+}
+
+// TODO(tflannag): This logic should be abstracted further away as a newVolume
+// type operation.
+// TODO(tflannag): Need a way to avoid run-once operations that cannot be reconciled
+// when state changes, e.g. avoid an InstallPlan-like scenario.
+// TODO(tflannag): Avoid hardcoding the storage requests.
+func (r *Reconciler) createPVC(bundleName string) (*corev1.PersistentVolumeClaim, error) {
```
exdx, line 98: so do we create the PV at all or is it up to the user to allocate a StorageClass or PV that can be claimed via the associated PVC? I like the latter option frankly, if we state that you need a storage solution in-cluster to use rukpak

pkg/k8s/provisioner/provisioner_controller.go
```diff
@@ -0,0 +1,128 @@
...
+
+	return ctrl.Result{}, utilerrors.NewAggregate(errors)
+}
+
+// ensureBundleVolume is a helper method responsible for ensuring the bundle's status
+// references an existing volume that will the content stored in the bundle sources'
+// filesystem.
+func (r *Reconciler) ensureBundleVolume(ctx context.Context, bundleName string) error {
+	fresh := &v1alpha1.Bundle{}
+	if err := r.Client.Get(ctx, getNonNamespacedName(bundleName), fresh); err != nil {
```
exdx, line 65: Is the reason we strip the namespace from the bundle here because each bundle name is unique across the cluster?

exdx: I can see that that is the case based on the cluster-scoped Bundle CRD. I am a little surprised, I know we wanted the Instance CRD to be cluster scoped but maybe namespacing bundles could help with storage and tenancy.

For example, as a user in one namespace, I want to install a Bundle. I won't be able to see Bundles in other namespaces necessarily. When I go to create a Instance from the Bundle, that is cluster-scoped and everyone can see it. Other users installing the same bundle will not be able to create an Instance from the same Bundle since one already exists. But they could install it onto the cluster in a separate mountpath on the PV and use it after the other Instance is removed, for example.

exdx (commented):
Nice! Just some thoughts.

cmd/serve/main.go
timflannagan: I wouldn't worry too much about the current structure of the code. This is me basically pumping out code when I get the chance so it's not the cleanest and should mainly be used as a reference point as a state of the union for this effort :laughing:

pkg/k8s/provisioner/provisioner_controller.go
timflannagan: Short answer: I'm developing locally on a single node cluster (e.g. kind) where it doesn't really matter too much what type of access mode(s) I'm using when creating a PVC, and I'm mainly interested to prototyping the ideas outlined in the e2e strawman.

I think looking forward, we'd like to abstract away the storage backend as much as possible. This could mean unpacked bundle contents are housed in-or-off cluster, as a PVC or shared across ConfigMaps (the latter being a little tricker for introspection of resources prior to approval), a database, or alternatives I haven't yet wrapped my head around. You might be able to make the argument that specification of a storage backend should be decided by the provisioner, maybe even defined as a parameter in a ProvisionerClass resource.

One other quick note, and this was especially true in the past and may not hold as much weight now, is that ReadWriteMany PVCs are typically harder to create as there are fewer storage provisioners that support that type of access mode.

pkg/k8s/provisioner/provisioner_controller.go
timflannagan: I think these are all good point to bring up. In the context of this work, where we haven't built out an Instance API yet, there's no real clear delineation of responsibilities of a Bundle vs. Instance resource as they're muddled together and that logic lives in the Bundle controller currently. As a result of that, the Bundle is cluster-scoped, adn we essentially utilize a global `rukpak` namespace, mainly for ease of local development as we don't have to care about introducing proper RBAC, or worrying about how we'll access resources defined in another namespace. 

This structure is highly likely to change, and outside of the rukpak CRDs, the overall structure is likely contingent on the individual provisioner until we get a better idea of how to consolidate some of these components into a re-usable library to avoid duplicate of effort on a per-provisioner-case.

I still need to continue working out in my head all of the various end-user workflow(s), and haven't gotten a chance to start thinking about how we'd handle authorization/activation of content referenced in individual Bundles (or a series of Bundles that share a label selector), or how a provisioner would create those resources. It's likely that at least one of these APIs will transition to a namespaced-scoped resource.

api/v1alpha1/bundle_types.go
timflannagan: Yeah completely agree - as an aside, everything related to how these status types are handled is up for grabs imo. I definitely like the idea of displaying availability, or approval, maybe these can be custom columns when querying local through kubectl would be a decent UX?

Makefile
```diff
@@ -55,6 +49,40 @@ verify: manifests generate
...
+.PHONY: vendor
+vendor:
+	go mod tidy
+	go mod vendor
+
+.PHONY: build
+build:
+	$(MAKE) bin/provisioner
+	$(MAKE) bin/unpacker
+	$(MAKE) bin/serve
```
njhale, line 76: Is there a reason these are separate make invocations and not preconditions?

api/v1alpha1/bundle_types.go
```diff
@@ -17,34 +17,68 @@ limitations under the License.
...
 type BundleStatus struct {
-	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
-	// Important: Run "make" to regenerate code after modifying this file
-	Conditions []metav1.Condition `json:"conditions,omitempty"`
-
-	// TODO: Reference to volume containing unpacked bundle content
+	Conditions []metav1.Condition           `json:"conditions,omitempty"`
+	Unpacked   BundleUnpackStatusType       `json:"unpacked,omitempty"`
+	Approved   BundleApprovalStatusType     `json:"approved,omitempty"`
+	Volume     *corev1.LocalObjectReference `json:"volume,omitempty"`
```
njhale, line 55: I think this is probably exposing too much about the underlying storage layer. The bundle content can be stored in memory for all a client cares, so long as the URI returns the content. 

The provisioner should probably have a scheme to find the storage it's using without relying on hints in the status.

api/v1alpha1/bundle_types.go
```diff
@@ -17,34 +17,68 @@ limitations under the License.
...
 // BundleStatus defines the observed state of Bundle
 type BundleStatus struct {
-	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
-	// Important: Run "make" to regenerate code after modifying this file
-	Conditions []metav1.Condition `json:"conditions,omitempty"`
-
-	// TODO: Reference to volume containing unpacked bundle content
+	Conditions []metav1.Condition           `json:"conditions,omitempty"`
+	Unpacked   BundleUnpackStatusType       `json:"unpacked,omitempty"`
+	Approved   BundleApprovalStatusType     `json:"approved,omitempty"`
```
njhale, line 54: Should this be `required`?

api/v1alpha1/bundle_types.go
```diff
@@ -17,34 +17,68 @@ limitations under the License.
...
+	Volume     *corev1.LocalObjectReference `json:"volume,omitempty"`
+	URI        string                       `json:"uri,omitempty"`
```
njhale, line 56: I think we may want some more info about a particular URI, like the content digest for comparison purposes. Maybe we should make something like [OCI Content Descriptors](https://github.com/opencontainers/image-spec/blob/main/descriptor.md#oci-content-descriptors) to do this.

pkg/k8s/provisioner/instance_controller.go
```diff
@@ -0,0 +1,256 @@
...
+	predicateProvisionerIDFilter := predicate.NewPredicateFuncs(func(obj client.Object) bool {
+		pc, ok := obj.(*v1alpha1.ProvisionerClass)
+		if !ok {
+			return false
+		}
+		return pc.Spec.Provisioner == ID
+	})
+	err = ctrl.NewControllerManagedBy(mgr).
+		For(&v1alpha1.ProvisionerClass{}, builder.WithPredicates(predicateProvisionerIDFilter)).
+		Watches(&source.Kind{Type: &v1alpha1.Bundle{}}, handler.EnqueueRequestsFromMapFunc(r.bundleHandler)).
```
njhale, line 67: So here's the tricky part: we _also_ need to dynamically add watches for all child resources created from each `Bundle` this provisioner processes. One way to do this is in-process, by spinning up new controllers within controller-runtime. Another way, would be to spin up a new `Deployment` configured to directly track and apply a specific set of resources.

timflannagan: I'm likely heading down using client-go's dynamic informer (and throwing that in it's own goroutine?) or updating the `manageWith`/`SetupWithManager` to simply return a controller.Controller for dynamic watching arbitrary resources. Also, TIL I learned you can return a controller using `Build(...)` instead of always defaulting to `Complete(...)`. Embedding a controller.Controller into something like the `bundleController` struct seems like a poor implementation detail long term but could work trying to get a POC up-and-running.

That Deployment angle is interesting being able to separate that logic from the core controller - what did you have in mind at a high-level?

PR #5, opened 2021-07-27
update README to include latest APIs

exdx:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@
 # Dependency directories
 vendor/
 bin/
+
--- a/README.md
+++ b/README.md
@@ -1,3 +1,101 @@
-# rukpak
+# RukPak
 
-Rukpak runs in a Kubernetes cluster and defines an API for installing cloud native bundle content 
+RukPak runs in a Kubernetes cluster and defines an API for installing cloud native bundle content. 
+
+## Introduction 
+
+RukPak is a pluggable solution for the packaging and distribution of cloud-native content and supports advanced strategies for installation, updates, and policy.
+The project provides a content ecosystem for installing a variety of artifacts, such as Git repositories, Helm charts, OLM bundles, and more onto a Kubernetes cluster. 
+These artifacts can then be managed, scaled, and upgraded in a safe way to enable powerful cluster extensions. 
+
+At its core, RukPak is a small set of APIs, packaged as Kubernetes CustomResourceDefinitions, and controllers that watch for those APIs. These APIs express what 
+content is being installed on-cluster and how to create a running instance of the content. 
+
+## Components 
+
+RukPak is composed of a few primary APIs: `Bundle`, `Instance`, and `ProvisionerClass`
+
+### Bundle
+A `Bundle` represents content that needs to be made available to other consumers in the cluster.
+Much like the contents of a container image need to be pulled and unpacked in order for Pods to start using them, 
+`Bundles` are used to reference content that may need to be pulled and should be unpacked. 
+In this sense, Bundle is a generalization of the image concept, and can be used to represent any type of content.
+
+`Bundles` do nothing on their own - they require a `Provisioner` to unpack and make their content available in-cluster. They can be
+unpacked to any arbitrary storage medium such as a PersistentVolume or ConfigMap. Each `Bundle` has an associated `spec.class` field which 
+indicates the `Provisioner` that should be watching and unpacking that particular bundle type. 
+
+Example Bundle
+```yaml
+apiVersion: core.rukpak.io/v1alpha1
+kind: Bundle
+metadata: 
+  name: example-operator.v0.9.3
+spec:
+  class: rukpack.io/k8s
+  refs:
+  - file://content
+  volumeMounts:
+  - mountPath: /content
+    configMap:
+      name: local 
+      namespace: plumbus
+```
+
+### Instance 
+The `Instance` API points to a Bundle and indicates that it should be “active”. This includes pivoting from older versions of an active bundle. 
+`Instance` may also include an embedded spec for a desired Bundle.
+
+Much like Pods stamp out instances of container images, `Instances` stamp out an instance of Bundles. `Instance` can be seen as a generalization of the Pod concept.
+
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -13,3 +13,6 @@
 
 # Dependency directories
 vendor/
+
+# goland 
+.idea/*
```
njhale, line 18: I'd like to keep the `.gitignore` sparse if we can -- @exdx maybe it's better for this to go in global git settings per contributor?

README.md
```diff
@@ -1,3 +1,102 @@
...
+`Bundles` are used to reference content that may need to be pulled and should be unpacked. 
+In this sense, Bundle is a generalization of the image concept, and can be used to represent any type of content.
+
+`Bundles` do nothing on their own - they require a `Provisioner` to unpack and make their content available in-cluster. They can be
+unpacked to any arbitrary storage medium such as a PersistentVolume or ConfigMap. Each `Bundle` has an associated `spec.class` field which 
+indicates the `Provisioner` that should be watching and unpacking that particular bundle type. 
+
+Example Bundle
+```yaml
+apiVersion: core.RukPak.io/v1alpha1
```
njhale, line 30: nit: `core.rukpak.io/v1alpha1`

exdx: whoops, find and replace did this

README.md
```diff
@@ -1,3 +1,102 @@
...
+RukPak is a pluggable solution for the packaging and distribution of cloud-native content and supports advanced strategies for installation, updates, and policy.
+The project provides a content ecosystem for installing a variety of artifacts, such as Git repositories, Helm charts, OLM bundles, and more onto a Kubernetes cluster. 
+These artifacts can then be managed, scaled, and upgraded in a safe way to enable powerful cluster extensions. 
+
+At its core, RukPak is a small set of APIs, packaged as Kubernetes CustomResourceDefinitions, and controllers that watch for those APIs. These APIs express what 
+content is being installed on-cluster and how to create a running instance of the content. 
+
+## Components 
+
+RukPak is composed of a few primary APIs: `Bundle`, `Instance`, and `Provisioner/ProvisionerClass`
```
timflannagan, line 16: I'd remove the mention of Provisioner here - it seems to imply that it's a concrete type vs. the controller responsible for watching these primary APIs.

README.md
```diff
@@ -1,3 +1,102 @@
-# rukpak
+# RukPak
 
-Rukpak runs in a Kubernetes cluster and defines an API for installing cloud native bundle content 
+RukPak runs in a Kubernetes cluster and defines an API for installing cloud native bundle content.
```
timflannagan, line 3: Any value in calling out the stage of this project (e.g. pre-alpha) and is subject to change?

README.md
```diff
@@ -1,3 +1,102 @@
...
+      - mountPath: /content
+        configMap:
+          name: resolved-654adh-content
+          namespace: olm
+```
+
+### Provisioner/ProvisionerClass
+`ProvisionerClass` defines a configuration for a `Provisioner`. Provisioners are controllers that understand `Instance`, `Bundle`, and `ProvisionerClass` APIs and take action.
+
+Each `Provisioner` has a unique id. For example, the `Provisioner` that ships with OLM and understand operator bundles has the id: operators.coreos.com/operators.
```
timflannagan, line 81: I think we'd like to substitute the OLM language here in favor of a rukpak-agnostic provisioner ID: https://github.com/operator-framework/rukpak/blob/main/pkg/k8s/provisioner/bundle_controller.go#L28.

timflannagan, line 81:
```suggestion
Each `Provisioner` has a unique id. For example, the `Provisioner` that ships with OLM and understands operator bundles has the id: operators.coreos.com/operators.
```

README.md
timflannagan: Might be worth linking the API type (`./api`) definition directory here.

README.md
timflannagan: Let's handle this kind of work in a follow-up/further down the line.

README.md
```diff
@@ -1,3 +1,102 @@
...
+spec:
+  class: olm.bundle
+  refs:
+  - file://content
+  volumeMounts:
+  - mountPath: /content
+    configMap:
+      name: local 
+      namespace: plumbus
+  URI: my-bundle.svc.local/content
```
timflannagan, line 43: This field will likely be exposed outside of the spec, and likely as a field that gets populated post-creation in the status field.

README.md
```diff
@@ -1,3 +1,102 @@
...
+```
+
+### Instance 
+The `Instance` API points to a Bundle and indicates that it should be “active”. This includes pivoting from older versions of an active bundle. 
+`Instance` may also include an embedded spec for a desired Bundle.
+
+Much like Pods stamp out instances of container images, `Instances` stamp out an instance of Bundles. `Instance` can be seen as a generalization of the Pod concept.
+
+The specifics of how an `Instance` makes changes to a cluster based on a referenced `Bundle` is defined by the 
+`ProvisionerClass` and the `Provisioner` that is configured to handle that `ProvisionerClass`.
```
timflannagan, line 53: @njhale Should we remove an mention of the Instance API until we flesh that API out more and document once that's been implemented, or just document it now as what we perceive as a future API?

timflannagan (commented):
We'll need to rebase this PR after I updated the .gitignore field in a previous PR.

I think it's fine to keep the majority of these changes and iterate as we build out this project further. I'd like to strip away any OLM-specific mentions, and just focus on rukpak here. 

I had a couple of comments that I'd like to see addressed (and/or get clarification on) before we merge this, but thanks for taking a stab at this task.

README.md
timflannagan: We can ignore linking the API directory for now - but I'd like to remove the mention of `Provisioner/ProvisionerClass` and just have `ProvisionerClass` documented here before merging this.

README.md
exdx: Yes that's correct, it shouldn't be on the spec

timflannagan (approved):

njhale (approved):
This looks g2g for now

In the future, I'd like:
- less buzzwords; e.g. "cloud-native"
- more focus on the fact that `rukpak` defines an interface and not an implementation

/lgtm

PR #7, opened 2021-08-30
Skeleton - k8s provisioner

njhale:
Adds a skeleton for the "plain" k8s bundle provisioner.

This includes:

- no-op Bundle and BundleClass controllers (wired to main())
- build, generate, and lint `make` targets
- multistage `Dockerfile`


provisioner/k8s/main.go
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	"os"
+
+	"github.com/operator-framework/rukpak/provisioner/k8s/cmd"
+)
+
+func main() {
+	if err := cmd.Execute(); err != nil {
+		os.Exit(1)
```
timflannagan, line 11: Log the error as well to stderr?

api/v1alpha1/types.go
```diff
@@ -0,0 +1,118 @@
...
+// BundleClassList contains a list of BundleClass
+type BundleClassList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []BundleClass `json:"items"`
+}
+
+// Source locates content to stage.
+// TODO(njhale): consider runtime.RawExtension as a way to let provisioners define their own schemes for sourcing content;
+// e.g. inline kubernetes manifests, resource references, etc.
```
timflannagan, line 60: That's an interesting thought - (nit) can we deflate this TODO so it doesn't also appear when rendering manifests. Open an RFE issue?

api/v1alpha1/types.go
```diff
@@ -0,0 +1,118 @@
...
+
+type ProvisionerID string
+
+// BundleClassSpec defines the desired state of BundleClass
+type BundleClassSpec struct {
+	Provisioner ProvisionerID `json:"provisioner"`
+
+	// +optional
+	Parameters map[string]string `json:"parameters,omitempty"`
+}
```
timflannagan, line 32: Minor detail, and something I'm not sure we really care about given we're very early in the dev stages of this project, but if we're interested in mirroring the StorageClass API, we may want to move these fields outside of the Spec and have everything live in the top-level BundleClass structure.

api/v1alpha1/types.go
```diff
@@ -0,0 +1,118 @@
...
+type Source string
+
+// BundleSpec defines the desired state of Bundle
+type BundleSpec struct {
+	// Class specifies the name of the provisioner that should manage the Bundle.
+	// +optional
+	Class string `json:"class,omitempty"`
+
+	// Sources locates all remote content to be staged by the Bundle.
+	Sources []Source `json:"sources"`
```
timflannagan, line 70: How do we feel about sticking with a single Source for now and increasing the scope when a concrete provisioner has been implemented?

provisioner/k8s/controller/bundle_controller.go
```diff
@@ -0,0 +1,49 @@
...
+
+	in := &v1alpha1.Bundle{}
+	if err := b.Get(ctx, req.NamespacedName, in); err != nil {
+		if apierrors.IsNotFound(err) {
+			// If the instance is not found, we're likely reconciling because of a DELETE event.
+			return reconcile.Result{}, nil
+		}
+
+		log.Error(err, "Error requesting Operator")
+		return reconcile.Result{Requeue: true}, nil
```
timflannagan, line 45: minor nit/nothing blocking: could also just deflate this to `client.IgnoreNotFound(err)`.

Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+COPY . .
+
+RUN make build
+
+FROM gcr.io/distroless/static:debug
+
+WORKDIR /
+COPY --from=builder /build/bin/k8s .
+
+EXPOSE 8080
```
timflannagan, line 13: It wasn't immediately clear to me why we'd need to expose 8080 here - exposing controller runtime's Prometheus metrics?

api/v1alpha1/types.go
njhale: I feel you. Unfortunately, the conventions for CRD-defined types _always_ include a `spec` field. I'm not sure if doing so would break other tooling (e.g. codegen, on-cluster validations, etc.).

api/v1alpha1/types.go
njhale: aye aye

Dockerfile
njhale: yup, I believe they're enabled by default in controller-runtime.

api/v1alpha1/types.go
timflannagan: That's a good point and something that I hadn't considered. Let's hold off for now then and follow-up on this later if we feel the need to revisit.

api/v1alpha1/types.go
njhale: Sure -- I could also figure out how to use [git-notes](https://git-scm.com/docs/git-notes).

api/v1alpha1/types.go
```diff
@@ -0,0 +1,116 @@
...
+// Source locates content to stage.
+type Source string
+
+// BundleSpec defines the desired state of Bundle
+type BundleSpec struct {
+	// Class specifies the name of the provisioner that should manage the Bundle.
+	// +optional
+	Class string `json:"class,omitempty"`
+
+	// Sources locates all remote content to be staged by the Bundle.
```
timflannagan, line 67: This could also be local content but this description is fine with me for now until we expand the provisioner's scope.

provisioner/k8s/cmd/run.go
```diff
@@ -0,0 +1,43 @@
...
+	k8scontroller "github.com/operator-framework/rukpak/provisioner/k8s/controller"
+)
+
+func init() {
+	rootCmd.AddCommand(runCmd)
+}
+
+var runCmd = &cobra.Command{
+	Use:   "run",
+	Short: "run the provisioner",
```
timflannagan, line 18: This doesn't have to be in this PR but can we update this command at some point to avoid printing the command usage when encountering an error (`SilenceUsage: true`).

njhale: done

api/v1alpha1/types.go
njhale: Yeah, I was thinking that we could use RawExtension and let provisioners ducktype this field too.

Let's stick with this and see where it falls down (we are in `v1alpha1` after all).

PR #14, opened 2021-08-30
Introduce barebones build and unit testing GH actions

timflannagan:


.github/workflows/build.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '~1.17'
+      - run: make build
```
timflannagan, line 19: @njhale I think this is fine for now, at least we can determine quickly whether we'd regressed in being able to compile the provisioner binary, but we may want to move towards a build and unit testing workflow where we actually build the container image -> publish locally -> unit tests are run using that container image?

njhale: > unit tests are run using that container image?

we normally create sparse images that don't contain source/test files. we could always compile the test binary, but I don't really see a reason to do that.

running `make unit-test` against the source in an action seems most natural here

Makefile
```diff
@@ -38,7 +38,7 @@ generate: ## Generate code and manifests
 	$(Q)go generate ./...
 
 # Static tests.
-.PHONY: test test-unit verify
+.PHONY: test test-unit verify build bin/k8s
```
njhale, line 41: `bin/k8s` is a real target though.

timflannagan: Sure, I'm only interested in ensuring this target always gets run, so maybe .FORCE would be more applicable here. I typically sway towards using .PHONY as it's more explicit imo. I don't have a strong opinion either way.

timflannagan: I realize I'm not entirely sure whether .FORCE would be relevant here - the `make bin/k8s` target creates local `./bin/k8s` executable, so adding that as a .PHONY target would make more sense. What we could pivot towards instead is creating a `clean` target that removes this file beforehand, and mark that target as a prerequisite to the `bin/k8s` target. I'd guess the .PHONY implementation is a bit more efficient but it's likely negligible at best.

PR #18, opened 2021-09-03
Makefile,manifests: Add skeleton of default rukpak resources

timflannagan:
Update the Makefile and add an `install` target that creates the
rukpak-specific resources.

Update the manifests root directory and add manifests for a `rukpak`
namespace and the requisite RBAC.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,6 +48,10 @@ test-unit: ## Run the unit tests
 verify: tidy generate format
 	git diff --exit-code
 
+install: generate
+	kubectl apply -f manifests
+	kubectl apply -f provisioner/k8s/manifests
+
 # Binary builds
 GO_BUILD := $(Q)go build
 
--- a/provisioner/k8s/manifests/namespace.yaml
+++ b/provisioner/k8s/manifests/namespace.yaml
@@ -1,0 +1,4 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: k8s-provisioner
--- a/provisioner/k8s/manifests/provisionerclass.yaml
+++ b/provisioner/k8s/manifests/provisionerclass.yaml
@@ -1,0 +1,6 @@
+apiVersion: core.rukpak.io/v1alpha1
+kind: ProvisionerClass
+metadata:
+  name: default
+spec:
+  provisioner: rukpak.io/k8s
--- a/provisioner/k8s/manifests/rbac.yaml
+++ b/provisioner/k8s/manifests/rbac.yaml
@@ -1,0 +1,59 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: k8s-provisioner-admin
+  namespace: k8s-provisioner
+rules:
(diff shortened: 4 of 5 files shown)


manifests/namespace.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+kind: Namespace
+metdata:
+  name: rukpak
```
njhale, line 4: Would it make more sense to have manifests -- other than the CRDs -- for the specific provisioner implementations themselves; e.g.

```yaml
kind: Namespace
metadata:
    name: k8s-provisioner
```

Makefile
```diff
@@ -48,6 +48,9 @@ test-unit: ## Run the unit tests
 verify: tidy generate format
 	git diff --exit-code
 
+install: generate
+	kubectl apply -f manifests
```
njhale, line 52: does this install a single provisioner? all provisioners?

njhale (changes requested):

manifests/namespace.yaml
timflannagan: Yeah, I think that's a good point but we could also just a quick refactoring of the current structure further down the line when we start trying to bridge out to other provisioner implementation - I'm fine with updating though. Maybe moving them to a nested directory under the ./provisioner directory?

Makefile
timflannagan: I think in the current landscape it would be the CRDs + the requisite static manifests to install the k8s bundle provisioner? And in the future, package provisioners individually outside of the root manifest directory?

ecordell: if the provisioner can run locally too, this could be something like:

```
go run ./provisioner/k8s ./deploy
```

manifests/namespace.yaml
timflannagan: @njhale Ended up updating and moved the k8s provisioner-specific YAML resources to ./provisioner/k8s/manifests.

Makefile
timflannagan: Yeah that's an interesting take - might be worth just taking that route for the time being until we get something more substantial (e.g. an actual k8s provisioner deployment that gets installed, which is not present in these PR changes). Should be an easy addition the nested k8s/provisioner/README installation instructions once those gets added.

njhale: Now that the manifests directory has moved, I think we need to either point directly to `provisioner/k8s/manifests` here or add a symlink to it in the top-level `manifests` directory.

njhale (changes requested):
Looking good!

Makefile
njhale: > go run ./provisioner/k8s ./deploy

@ecordell I like this, and I think it implies that `./deploy` is a directory containing a "plain k8s bundle".

Seems like a solid feature-request and long term direction for this.

timflannagan: Let's just point directly to that directory path for now and iterate on the structure as we get more of this project built out.

njhale (approved):
/lgtm

PR #19, opened 2021-09-03
Introduce a bundle unpacking component 

timflannagan:
Update the provisioner directory and creates an unpacker package that's
responsible for pulling down and extracting referenced bundle contents.
An individual Bundle custom resource is pulled down, where the image
layers are extracted and fed into a tar reader. When reading the tar
contents, ensure that only the configured (through a CLI flag) manifest
directory is written to a configured ConfigMap resource.

Add a Makefile target for building the unpacker executable.


provisioner/unpacker/main.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+	if cm.Data == nil {
+		cm.Data = map[string]string{}
+	}
+
+	ref, err := name.ParseReference(string(bundle.Spec.Source))
+	if err != nil {
+		return err
+	}
+	fmt.Println(ref)
```
timflannagan, line 102: Note: if we're interesting in chasing down this route, we'd likely need to add support for configuring authentication. There are a couple of different implementations this library provides, which is nice, but it's not immediately clear whether it's compatible with things like manifest list images, and obviously we wouldn't be pulling through CRI.

provisioner/unpacker/main.go
```diff
@@ -0,0 +1,144 @@
...
+	cm := &corev1.ConfigMap{}
+	if err := client.Get(ctx, types.NamespacedName{
+		Name:      bundle.GetName(),
+		Namespace: o.StorageNamespace,
+	}, cm); err != nil {
+		return err
+	}
+	if cm.Data == nil {
+		cm.Data = map[string]string{}
+	}
```
timflannagan, line 96: Open to suggestions on whether this implementation is acceptable: we're essentially hardcoding that we can find the Bundle's underlying storage (i.e. a ConfigMap in this case) via the Bundle's metadata.Name, and all storage backends are housed in a shared namespace (e.g. the `k8s-provisioner` namespace).

provisioner/unpacker/main.go
```diff
@@ -0,0 +1,144 @@
...
+		return err
+	}
+
+	bundle := &v1alpha1.Bundle{}
+	if err := client.Get(ctx, types.NamespacedName{Name: o.BundleName}, bundle); err != nil {
+		return err
+	}
+
+	cm := &corev1.ConfigMap{}
+	if err := client.Get(ctx, types.NamespacedName{
```
exdx, line 88: Here we get a configmap, but we don't check to see if it has content in it before attempting to update it later on. Does the unpacker ever care about overwriting some data already in storage?

provisioner/unpacker/main.go
```diff
@@ -0,0 +1,144 @@
...
+		Name:      bundle.GetName(),
+		Namespace: o.StorageNamespace,
+	}, cm); err != nil {
+		return err
```
exdx, line 92: If the configmap is not on-cluster already and we get a NotFound error, should we create one here?

provisioner/unpacker/main.go
```diff
@@ -0,0 +1,144 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	controllerruntime "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/operator-framework/rukpak/api/v1alpha1"
+)
+
+type unpackOptions struct {
+	BundleName       string
+	StorageNamespace string
```
exdx, line 30: Maybe this is premature optimization but I think we could add `StorageType` or something here that would represent that this unpacker unpacks to a ConfigMap specifically. Or simply call it ConfigMapUnpacker.

provisioner/unpacker/main.go
```diff
@@ -0,0 +1,144 @@
...
+
+	bundle := &v1alpha1.Bundle{}
+	if err := client.Get(ctx, types.NamespacedName{Name: o.BundleName}, bundle); err != nil {
+		return err
+	}
+
+	cm := &corev1.ConfigMap{}
+	if err := client.Get(ctx, types.NamespacedName{
+		Name:      bundle.GetName(),
+		Namespace: o.StorageNamespace,
```
exdx, line 90: I think hardcoding the name is ok based on the bundle name, but I believe the storage backends will probably be uniquely separated across namespaces. They probably won't all housed in one namespace. We should think about a good place to express that to the unpacker.

provisioner/unpacker/main.go
timflannagan: Agreed, I was thinking the same thing and I've been playing around something locally that attempts to fire off Jobs based on the type of Bundle source being reconciled. I landed on avoiding trying to make this Job look like fully fleshed out helm chart template, but I think that's a reasonable path forward long-term, once more Bundle sources are supported.

provisioner/unpacker/main.go
timflannagan: > Does the unpacker ever care about overwriting some data already in storage?

I've typically landed on the Bundle's source reference acts as the source-of-truth vs. what's present on-cluster, so I'm not super worried about overwriting the on-cluster representation if the end goal is a declarative installation. There's likely a caching layer that could be introduced to alleviate re-processing already present resources on-cluster, but I think that's something we can investigate further down the line as we work towards getting something worker vs. pre-maturely optimizing.

provisioner/unpacker/main.go
```diff
@@ -0,0 +1,144 @@
...
+		if !strings.Contains(header.Name, o.ManifestDir) {
+			continue
+		}
+		logger.Infof("processing the %s file", header.Name)
+
+		content, err := ioutil.ReadAll(t)
+		if err != nil {
+			return fmt.Errorf("failed to read from the current %s file: %v", header.Name, err)
+		}
+		cm.Data[filepath.Base(header.Name)] = string(content)
```
akihikokuroda, line 131: I wonder if Data is better than BinaryData.

PR #42, opened 2022-02-21
Update the rukpak API definitions to match the joelanford/kuberpak APIs

timflannagan:
Update the api/v1alpha1/bundle_types.go, copy over the
joelanford/kuberpak type definitions, and re-run code generation.

Introduce the api/v1alpha1/bundleinstance_types.go file, which is
copied over from joelanford/kuberpak. Update the code generation using
the `make generate` command, and commit the code generate changes and
the newly generated BundleInstance CRD file.


api/v1alpha1/bundle_types.go
```diff
@@ -18,48 +18,73 @@ package v1alpha1
...
+	PhaseUnpacked  = "Unpacked"
+)
 
 // BundleSpec defines the desired state of Bundle
 type BundleSpec struct {
-	// Class specifies the name of the provisioner that should manage the Bundle.
-	// +optional
-	Class string `json:"class,omitempty"`
+	// ProvisionerClassName sets the name of the provisioner that should reconcile this BundleInstance.
+	ProvisionerClassName string `json:"provisionerClassName"`
```
timflannagan, line 43: Note: Joe's implementation has Bundle's specifying the unique provisioner ID (e.g. `kuberpak.io/registry+v1`) vs. referencing the metadata.Name of an existing ProvisionerClass resource, which has the unique provisioner ID referenced in it's spec. I think this implementation is fine for the time being as we continue to iterate and prototype, and we can re-introduce (or re-evaluate) the value of this ProvisionerClass API and make that breaking change before GA.

api/v1alpha1/bundle_types.go
```diff
@@ -18,48 +18,73 @@ package v1alpha1
...
+	Objects []BundleObject `json:"objects,omitempty"`
+}
 
-	Conditions []metav1.Condition `json:"conditions,omitempty"`
+type BundleObject struct {
+	Group     string `json:"group"`
+	Version   string `json:"version"`
+	Kind      string `json:"kind"`
+	Name      string `json:"name"`
+	Namespace string `json:"namespace"`
```
timflannagan, line 79: Another note: we had talked about whether this implementation detail also works with Bundles that contain contents that aren't GVK-based, e.g. a Bundle that contains OPA rules, and whether those use cases make sense to support. I think this is fine to keep in the short term, and we can re-evaluate over time as we continue to iterate, and make any breaking changes before GA.

api/v1alpha1/bundle_types.go
```diff
@@ -18,48 +18,73 @@ package v1alpha1
...
-	// +optional
-	Class string `json:"class,omitempty"`
+	// ProvisionerClassName sets the name of the provisioner that should reconcile this BundleInstance.
+	ProvisionerClassName string `json:"provisionerClassName"`
 
-	// Source locates all remote content to be staged by the Bundle.
-	Source Source `json:"source"`
-}
+	// Image is the bundle image that backs the content of this bundle.
+	Image string `json:"image"`
```
exdx, line 46: This field has changed from a more nebulous `Source` to a concrete `Image` field referring to container images only. Maybe we can revisit this change once other content sources becomes supported?

I kind of like the more generic term but agree with keeping things simple for now.

api/v1alpha1/bundle_types.go
```diff
@@ -18,48 +18,73 @@ package v1alpha1
...
-// Content surfaces staged content for clients to access.
-type Content struct {
-	runtime.RawExtension `json:"-"`
+	// ImagePullSecrets is a list of pull secrets to have available to
+	// pull the referenced image.
+	ImagePullSecrets []ImagePullSecret `json:"imagePullSecrets,omitempty"`
 }
 
-type ProvisionerID string
+type ImagePullSecret struct {
```
exdx, line 53: As discussed in our conversation this is not really secure: it allows the provisioner to potentially copy secrets from any namespace into it's own namespace. Maybe we should just drop support for private bundle images for now?

api/v1alpha1/bundle_types.go
timflannagan: Yep - I missed calling this out. Looking at the e2e strawman, it looks like we would like to support multiple `sources` over time, but we decided on keeping the API fairly scoped in the short term, which means the following:

- We only support a single Bundle source;
- We only support a remote container image Bundle source;

With that in mind, we can simply land on a spec.Image field in the short term, and work towards a model that supports multiple, different Bundle sources depending on the individual provisioner and what that provisioner supports under-the-hood.

api/v1alpha1/bundle_types.go
timflannagan: Good catch - removed this in 3aec843fd8fe1f0f73d847869d6447d960e81b62

exdx (approved):

PR #48, opened 2022-02-22
feat: update kuberpak provisioner to registryv1

### LINKED ISSUE #45: Rename the kuberpak provisioner
Goal: rename the kuberpak provisioner to something more closely resembles the registry+v1 bundle format.

AC:
- Update the provisioner/kuberpak repository and rename that.
- Update any package imports that reference kuberpak.
- Update the root Makefile and change the target that builds the kuberpak controller binary

exdx:
Closes #45 

This PR updates the recently introduced kuberpak provisioner to the registryv1 provisioner as the naming convention is that the provisioner should be named after the bundle type. It also updates references from kuberpak.io to core.rukpak.io indicating the new home of the provisioner, where all default provisioners are considered part of the core.rukpak.io api group.


exdx:
Branch looks wrong -- closing.

provisioner/README.md
```diff
@@ -3,4 +3,4 @@
 The following subdirectories contain concrete Provisioner implementations.
 
 - [k8s](k8s/README.md) - provisions "plain" k8s bundles
-- [kuberpak](kuberpak/README.md) - provisioners "registry+v1" OLM bundles
+- [registryv1](registryv1/README.md) - provisions "registry+v1" OLM bundles
```
timflannagan, line 6: whoops - nice catch.

provisioner/registryv1/main.go
```diff
@@ -108,7 +108,7 @@ func main() {
 	}
 
 	// TODO: derive pod namespace from the pod that this process is running in.
-	ns := "kuberpak-system"
+	ns := "registryv1-system"
```
timflannagan, line 111: Hmm... this reads weird? Maybe something like `rukpak-system` might make more sense in the short term while we figure out how multiple provisioners will be structured on-cluster?

provisioner/registryv1/controllers/bundle_controller.go
```diff
@@ -199,7 +199,7 @@ func (r *BundleReconciler) ensureUnpackPod(ctx context.Context, bundle *olmv1alp
 	pod.SetNamespace(r.PodNamespace)
 
 	return util.CreateOrRecreate(ctx, r.Client, pod, func() error {
-		pod.SetLabels(map[string]string{"kuberpak.io/owner-name": bundle.Name})
+		pod.SetLabels(map[string]string{"core.rukpak.io/owner-name": bundle.Name})
```
timflannagan, line 202: This can be done in a follow-up, but I think we should make this `core.rukpak.io` domain a constant, or that entire label a constant so we reduce the number of places we'd need to manually update it if we decide to go in another direction. Thoughts?

doc/bundles.md
```diff
@@ -0,0 +1,36 @@
+## Bundles
```
timflannagan, line 1: Can we pull this out into it's own commit, or it's own PR entirely?

exdx: Yes, sounds good, it should be in another PR altogether. I had some other changes locally that made their way through by accident

provisioner/registryv1/controllers/bundle_controller.go
exdx: Yes, we definitely need a consistent labeling strategy and have the label live in one place in the codebase. Good follow-up change.

PR #50, opened 2022-02-22
Update ci

### LINKED ISSUE #10: Bootstrap a CI/CD pipeline
Introduce github actions for the building and testing workflows.

tylerslaton:
Adding a few things here to improve the signaling/DevEx we get from CI: 

1. Grouped some actions together (e2e and unit, for example)
2. Added a .golangci.yaml file and accompanying `lint` action. Also addressed complaints from golangci.
3. Added an e2e action and Ginkgo's needed boilerplate code.
4. Updated the action's `on` call to run on pull request and pushes to main.

Closes #10


.github/workflows/build.yaml
```diff
@@ -1,15 +1,16 @@
 name: build
 
 on:
+  workflow_dispatch:
+  pull_request:
+    paths-ignore:
+    - '**/*.md'
```
timflannagan, line 7: Let's remove the paths-ignore keyword for now as it doesn't play well with branch protection rules.

.github/workflows/sanity.yaml
```diff
@@ -1,12 +1,13 @@
 name: sanity
 
 on:
+  workflow_dispatch:
+  pull_request:
+    paths-ignore:
+    - '**/*.md'
```
timflannagan, line 7: Same thing as above: Let's remove the paths-ignore keyword for now as it doesn't play well with branch protection rules.

test/e2e/e2e_suite_test.go
```diff
@@ -0,0 +1,13 @@
...
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestE2E(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "E2E Suite")
+}
```
timflannagan, line 13: It's not super clear to me whether we'll have a top-level e2e suite vs. a per-provisioner e2e suite, but this is probably fine for now until we get a better feel for the repository.

provisioner/kuberpak/internal/convert/registryv1.go
```diff
@@ -109,12 +109,12 @@ func Convert(in RegistryV1, installNamespace string, targetNamespaces []string)
...
-		roles               []rbacv1.Role
-		roleBindings        []rbacv1.RoleBinding
-		clusterRoles        []rbacv1.ClusterRole
-		clusterRoleBindings []rbacv1.ClusterRoleBinding
-	)
+
+	roles := []rbacv1.Role{}
+	roleBindings := []rbacv1.RoleBinding{}
+	clusterRoles := []rbacv1.ClusterRole{}
+	clusterRoleBindings := []rbacv1.ClusterRoleBinding{}
```
timflannagan, line 116: This was outlined by golangci-lint? The changes here should be identical to the original variable declarations.

test/e2e/e2e_suite_test.go
tylerslaton: 🤔 Yeah that's a good point. I would like to think we'd have a per-provisioner suite now that you mention it. It would be hard to bootstrap a suite into the two existing provisioners and would set a good precedent. Would you me want to move forward on that?

provisioner/kuberpak/controllers/bundleinstance_controller.go
```diff
@@ -131,15 +131,14 @@ func (r *BundleInstanceReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
-				Status:  metav1.ConditionFalse,
-				Reason:  "BundleLookupFailed",
-				Message: err.Error(),
-			})
-			return ctrl.Result{}, err
 		}
+		meta.SetStatusCondition(&bi.Status.Conditions, metav1.Condition{
+			Type:    "Installed",
+			Status:  metav1.ConditionFalse,
+			Reason:  "BundleLookupFailed",
```
timflannagan, line 138: unrelated: We should add these hardcoded reasons to the bundleinstance_types.go file at some point.

provisioner/kuberpak/internal/convert/registryv1.go
tylerslaton: Yeah it was outlined by `prealloc` linter. I was debating just removing that linter for this reason since this felt kinda strange.

provisioner/kuberpak/controllers/bundleinstance_controller.go
tylerslaton: Definitely, that'll particularly become nice when the E2E suite becomes non-boilerplate.

.github/workflows/sanity.yaml
```diff
@@ -18,3 +17,13 @@ jobs:
...
         run: make verify
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '~1.17'
+      - name: Run verification checks
+        run: make lint
```
timflannagan, line 28: Let's use the golangci-lint first-class action for now as it caches the module cache under-the-hood which speeds up linting runs significantly. You can check out https://github.com/operator-framework/operator-lifecycle-manager/blob/master/.github/workflows/sanity.yaml#L23-L25 for a concrete use in OLM.

timflannagan: We can also remove any Makefile targets that are covered by the golangci-lint integration (e.g. anything that runs `gofmt ...` or `goimports ...`, etc.).

.github/workflows/build.yaml
```diff
@@ -1,15 +1,14 @@
...
     branches:
-      - '**'
-  pull_request:
-    paths:
-      - '**'
+    - main
 
 jobs:
-  build:
+  binaries:
```
timflannagan, line 11: This can be tackled as a follow-up, and this is only an issue in the short term until e2e is fleshed out, but it looks like we have a job lower in this build workflow for ensuring the root Dockerfile container image can be successfully built. We should also verify that built image can be run, even if it's just running the `--help` flag for the registry+v1 or k8s provisioners.

provisioner/kuberpak/controllers/bundle_controller.go
```diff
@@ -95,19 +95,21 @@ func (r *BundleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ c
...
+		return ctrl.Result{}, updateStatusUnpackFailing(&u, fmt.Errorf("ensure unpack pod: %w", err))
 	} else if op == controllerutil.OperationResultCreated || op == controllerutil.OperationResultUpdated || pod.DeletionTimestamp != nil {
 		updateStatusUnpackPending(&u)
 		return ctrl.Result{}, nil
 	}
 
 	switch phase := pod.Status.Phase; phase {
 	case corev1.PodPending:
-		return ctrl.Result{}, r.handlePendingPod(&u, pod)
+		r.handlePendingPod(&u, pod)
```
timflannagan, line 106: Hmm interesting - this is another case where golangci-lint linting surprises me. Any idea on which specific linter was complaining about this implementation?

provisioner/kuberpak/internal/convert/registryv1.go
timflannagan: Yeah this feels strange - I think I would be favor of dropping it. WDYT?

provisioner/kuberpak/controllers/bundle_controller.go
tylerslaton: This was `unparam` with this message:

> unparam      (*BundleReconciler).handlePendingPod - result 0 (error) is always nil

I also thought it was quite strange.

provisioner/kuberpak/internal/convert/registryv1.go
tylerslaton: To be fair, this would also satisfy the linter:

```go
var (
	roles        []rbacv1.Role
	roleBindings []rbacv1.RoleBinding
)

clusterRoles := []rbacv1.ClusterRole{}
clusterRoleBindings := []rbacv1.ClusterRoleBinding{}
```

tylerslaton: I'm also good with dropping it, sorry Github comments didn't update for a bit so I didn't see your message.

.golangci.yaml
```diff
@@ -0,0 +1,25 @@
+run:
+  # Default timeout is 1m, up to give more room
+  timeout: 4m
+
+linters:
```
timflannagan, line 5: This can also be done as a follow-up: it might be worth expanding the `importas` linter and specifying a core of common package imports we expect, e.g. `rukpakv1alpha1`, `corev1`, `metav1`, etc. You can find how to do that using [golangci-lint's documentation around the importas linter](https://golangci-lint.run/usage/linters/#importas).

provisioner/kuberpak/controllers/bundle_controller.go
timflannagan: Yeah I've had issues with this linter in the past, but it's possible it's uncovering a true error? In any case, I think it's fine to keep it enabled, and we can revisit reverting back to the original implementation once we get a better testing surface.

timflannagan: Oh I realize it's because we only return nil errors from that function, which means the error return variable can be dropped, which makes more sense now.

timflannagan (approved):
There's a couple of things I'd still like to see get done but I don't want to hold up merging this as having a basic CI/CD pipeline in place while we iterate should be pretty valuable, at least so we don't step on each others toes too much.

PR #51, opened 2022-02-23
Makefile,Dockerfile: Fix the runtime errors in the rukpak container image

timflannagan:
Update the Makefile and ensure that cgo is disabled for both the
registry+v1 and plain bundle provisioners. Update the Dockerfile to
specify an explicit path to the `k8s` executable to silence the "exec
user process caused: no such file or directory" error message when
running this container image locally.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,15 +8,12 @@ help: ## Show this help screen
 	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z0-9_-]+:.*?##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
 
 # Code management
-.PHONY: lint format tidy clean generate
+.PHONY: lint tidy clean generate
 
 PKGS = $(shell go list ./...)
 
 lint: golangci-lint
 	$(Q)$(GOLANGCI_LINT) run
-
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -19,16 +8,13 @@ help: ## Show this help screen
 	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z0-9_-]+:.*?##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
 
 # Code management
-.PHONY: lint format tidy clean generate
```
tylerslaton, line 22: 👍🏻

Makefile
```diff
@@ -63,10 +49,10 @@ GO_BUILD := $(Q)go build
 build: bin/k8s bin/kuberpak
 
 bin/k8s:
-	$(GO_BUILD) -o $@ ./provisioner/k8s
+	CGO_ENABLED=0 go build -o $@ ./provisioner/k8s
```
tylerslaton, line 52: non-blocking: I'd think it would make more sense to just add this to the GO_BUILD variable instead. Thoughts?

tylerslaton (approved):
Looks good! Non-blocking suggestion, but good to go!

Makefile
timflannagan: I guess it felt a bit awkward to use that go build variable as there's some parsing issues with ordering? We should tackle overhauling this Makefile at some point though as it's all over the place right now.

PR #56, opened 2022-02-23
registryv1: Deflate the internal/ and cmd/ directories to root

timflannagan:
Move the nested internal/ and cmd/ directories from the
kuberpak/registry+v1 provisioner to the root directory as a first step
towards creating a shared library for other provisioners to use.

Update the package imports for any package referencing the old nested directory structure.

Add the bundle unpacking component's Dockerfile to the root repository as well
and update the build GHA workflow to build that unpack.Dockerfile during CI runs.


Makefile
```diff
@@ -54,6 +54,10 @@ bin/k8s:
 bin/registryv1:
 	CGO_ENABLED=0 go build -o $@ ./provisioner/registryv1
 
+bin/unpack:
+	CGO_ENABLED=0 go build -o $@ ./cmd/unpack/...
```
exdx, line 58: nit: should all these `bin/` directives use the `GO_BUILD` variable defined above? ie
`CGO_ENABLED=0 $(GO_BUILD) -o $@ ./cmd/unpack/...`

exdx (approved):
Nice, breaking this out into small commits made it easy to review. One small nit otherwise LGTM.

Makefile
timflannagan: I wasn't able to get that working when I was playing around with it yesterday due to parsing issues. When the verbosity flag is enabled, the go build command is prefixed with a `@` symbol, so that expression gets evaluated as `CGO_ENABLED=0 @go build -o ...` which make can't parse. The alternative was to override that GO_BUILD variable everytime it gets used which didn't feel super ideal either. I think we need to overhaul the Makefile at some point in the near future, so maybe we can visit this conversation then?

PR #58, opened 2022-02-23
.github/workflow: Ensure we're caching the go module cache

timflannagan:
Update the build GHA workflow and ensure we're caching the Go module
cache between CI runs. This allows us to have quicker overall CI run
times as we avoid having to constantly populate the Go module cache when
building the provisioner binaries in CI.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -15,7 +15,19 @@ jobs:
       - uses: actions/setup-go@v2
         with:
           go-version: '~1.17'
-      - run: make build
+
+      - name: Cache the go module and build paths
+        uses: actions/cache@v2
+        with:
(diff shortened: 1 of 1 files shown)


timflannagan:
Weird it looks like it's actually taking more time: https://github.com/operator-framework/rukpak/runs/5306853237?check_suite_focus=true?

/hold

timflannagan:
The problem was we weren't caching the go build cache :facepalm: 

/hold cancel

tylerslaton (approved):
Not going to block this PR, but it would be cool to see a cache job that all the other jobs rely on. The reason for this being that we could cache all the dependencies that the other jobs need in a single place and they can all use that cache. Thoughts?

PR #61, opened 2022-02-24
add go-apidiff make target and CI check

### LINKED ISSUE #54: Add go-apidiff to CI
What do we think about running my joelanford/go-apidiff in CI to give us signal about PRs that introduce new APIs or break existing APIs?

joelanford:

Closes #54


.github/workflows/sanity.yaml
```diff
@@ -21,7 +21,26 @@ jobs:
...
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '~1.17'
+      - id: golangci-lint
+        run: echo "::set-output name=version::$(cd hack/tools; go list -m github.com/golangci/golangci-lint | cut -d' ' -f2)"
       - name: Run linting checks
-        uses: "golangci/golangci-lint-action@v2"
+        uses: golangci/golangci-lint-action@v2
+        with:
+          version: ${{ steps.golangci-lint.outputs.version }}
```
joelanford, line 32: Rather than hardcoding to 1.43 here (and risking getting out of sync with what's in our `hack/tools/go.mod`), I thought "let's just read our go.mod and use that!"

timflannagan: Do we need a `needs` keyword here if we're reliant on the output of parsing the go.mod in the previous job step?

timflannagan (approved):
lgtm - It would be nice if we could enable caching for the tool-related binaries so we don't need to always re-populate the module cache every time a CI action is run.

joelanford:
/hold
Trying a few things out.

I agree caching is good. I added auto-caching to the go-apidiff action last week after being inspired by golangci-lint's action, so we could use that and do the same sort of version detection that I added with go-apidiff. (That's the current state of this PR)

Alternatively, we could also just continue using `make go-apidiff` and I could add an explicit caching step in the `go-apidiff` job.

joelanford:
Closing - we don't need this until we're closer to a point where API stability matters.

PR #62, opened 2022-02-24
docs: add provisionerv1 doc

### LINKED ISSUE #60: Add documentation around the plain manifest bundle provisioner
Goal: the plain manifest provisioner was recently introduced as a POC and we've been working on productizing it. We should aim towards having some documentation around how to use it for those that want to play around with it locally, and build on top of this documentation set over time.

AC:

- Add a description of the provisioner
- Outline some simple use cases
- Document how to run the provisioner locally

Stretch:

- Cleanup the root README.md and remove any out-of-date information from the last time we combed over that file.

exdx:

Closes #60 

This PR adds documentation about the new registryv1 provisioner: a summary, use cases, and how to install and use locally.

### DIFF
--- a/provisioner/registryv1/README.md
+++ b/provisioner/registryv1/README.md
@@ -1,0 +1,123 @@
+# registryv1 Provisioner
+
+## Summary
+
+The registryv1 provisioner is a core rukpak provisioner that knows how to interact with `registry+v1` format bundles. 
+The `registry+v1` bundles are the typical OLM bundles that contain a CSV manifest as well as other supported manifest types, such as CRDs and RBAC.
+They contain a manifests and metadata directory and can be built via standard OLM tooling such as the `opm` tool.
+For more information on the `registry+v1` bundle format see the [documentation](https://olm.operatorframework.io/docs/tasks/creating-operator-bundle/#operator-bundle). 
+
+The registryv1 provisioner is able to unpack a given `registry+v1` bundle onto a cluster and then instantiate it, making the
+content of the bundle available in the cluster. It does so by reconciling the `Bundle` and `BundleInstance` types that 
+have the spec.provisionerClassName field set to `core.rukpak.io/registry+v1`. This field must be set to the correct provisioner name 
+in order for the registryv1 provisioner to see and interact with the bundle. 
+
+Since this provisioner is responsible for creating OLM bundles on-cluster, it has a soft dependency on 
+OLM and the existing set of APIs, specifically ClusterServiceVersion and OperatorGroup. Those APIs must exist on the cluster
+in order to instantiate a `registry+v1` bundle successfully. The provisioner creates the CSV provided in the bundle, but delegates
+additional handling of the CSV to OLM itself, therefore OLM should be installed on the cluster alongside this provisioner. 
+
+## Use cases
+
+### Install and apply a specific version of a bundle 
+
+The registryv1 provisioner can install and make available a specific `registry+v1` bundle available in the cluster. This is in contrast to 
+the existing OLM workflow which requires a Subscription and manual approval for the relevant InstallPlan to pin to a specific version.  
+
+Simply create a `Bundle` resource pointing to a specific version of your bundle, and a `BundleInstance` which references that bundle.
+The registryv1 provisioner will unpack the provided Bundle onto the cluster, and eventually make the content available on the cluster. 
+
+```yaml
+apiVersion: core.rukpak.io/v1alpha1
+kind: Bundle
+metadata:
+  name: my-bundle
+spec:
+  image: my-bundle-image@sha256:xyz123
+  provisionerClassName: core.rukpak.io/registry+v1
+---
+apiVersion: core.rukpak.io/v1alpha1
+kind: BundleInstance
+metadata:
+  name: my-bundle-instance
+spec:
+  provisionerClassName: core.rukpak.io/registry+v1
+  bundleName: my-bundle
+``` 
+### Make bundle content available but do not install it
+
+There is a natural separation between sourcing of the content and application of that content via two separate RukPak APIs, `Bundle` and `BundleInstance`. 
+A user can specify a particular `Bundle` to be available in the cluster for inspection before any application of the resources. 
+Given a `Bundle` resource named `my-bundle`, the registryv1 provisioner will pull down and unpack the bundle to a variety of storage backends, such as ConfigMaps or PersistentVolumes.
+
+Currently the ConfigMap backend is supported by the provisioner, and the content can be inspected by viewing the metadata ConfigMap, which holds
+information on all ConfigMaps related to this bundle. 
+
+`kubectl get cm bundle-metadata-my-bundle -o yaml` provides access to this metadata ConfigMap, which then links to all other ConfigMap objects
+where the unpacked data is stored. 
+
+Surfacing the content of a bundle in a more user-friendly way, via a plugin or additional API, is on the RukPak roadmap. 
+
+### Pivot between bundle versions 
+
+The `BundleInstance` API is meant to indicate the version of the bundle that should be active within the cluster. Given two unpacked bundles in the cluster,
+`my-bundle-v0.0.1` and `my-bundle-v0.0.2`, the `spec.bundleName` field of the related `BundleInstance` can be updated from say, the older version,
+to the newer version, and the registryv1 provisioner will be able to successfully create the new content on-cluster and remove the old content.  
+
+## Running locally 
+
+To experiment with the registryv1 provisioner locally, first setup a local cluster. It's recommended to have OLM installed on the cluster as well, 
+which can be done easily by heading to the [releases](https://github.com/operator-framework/operator-lifecycle-manager/releases) section and installing via the install script. 
+
+Once the cluster is setup and OLM is running, take the following steps from the root of the repository:
+* Create the rukpak namespace via `kubectl create ns rukpak-system` by default
+* Run `make bin/registryv1` to create the registryv1 provisioner binary locally
+* Apply the RukPak core CRDs via `kubectl apply -f manifests/`
+* Run `./bin/registryv1` to start the controller 
+
+From there, creating some `Bundles` and `BundleInstance` types to see the provisioner in action. For an example bundle to use, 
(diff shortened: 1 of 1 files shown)


provisioner/registryv1/README.md
```diff
@@ -0,0 +1,123 @@
...
+apiVersion: core.rukpak.io/v1alpha1
+kind: BundleInstance
+metadata:
+  name: combo-system
+spec:
+  provisionerClassName: core.rukpak.io/registry+v1
+  bundleName: combo-0.0.1
+```
+
+From there, check out the combo operator and ensure that the CSV is present on the cluster.
```
exdx, line 123: I like using the combo operator as an example, but when I install on the cluster locally the CSV stays in a `Pending` state. looking at the status it looks like there are some PolicyRules that are missing. This is maybe due to a gap between installing a CSV directly versus via a Subscription to a catalog. Not sure what the next step is. 

```
  - dependents:
    - group: rbac.authorization.k8s.io
      kind: PolicyRule
      message: namespaced rule:{"verbs":["get","list","watch","create","update","patch","delete"],"apiGroups":[""],"resources":["configmaps"]}
      status: NotSatisfied
      version: v1
    - group: rbac.authorization.k8s.io
      kind: PolicyRule
      message: namespaced rule:{"verbs":["get","list","watch","create","update","patch","delete"],"apiGroups":["coordination.k8s.io"],"resources":["leases"]}
      status: NotSatisfied
      version: v1
    - group: rbac.authorization.k8s.io
      kind: PolicyRule
      message: namespaced rule:{"verbs":["create","patch"],"apiGroups":[""],"resources":["events"]}
      status: NotSatisfied
```

Installing a dummy operator instead, something without any permisions or just a set of APIs is possible for the purposes of the README but not as compelling.

provisioner/registryv1/README.md
```diff
@@ -0,0 +1,123 @@
...
+## Summary
+
+The registryv1 provisioner is a core rukpak provisioner that knows how to interact with `registry+v1` format bundles. 
+The `registry+v1` bundles are the typical OLM bundles that contain a CSV manifest as well as other supported manifest types, such as CRDs and RBAC.
+They contain a manifests and metadata directory and can be built via standard OLM tooling such as the `opm` tool.
+For more information on the `registry+v1` bundle format see the [documentation](https://olm.operatorframework.io/docs/tasks/creating-operator-bundle/#operator-bundle). 
+
+The registryv1 provisioner is able to unpack a given `registry+v1` bundle onto a cluster and then instantiate it, making the
+content of the bundle available in the cluster. It does so by reconciling the `Bundle` and `BundleInstance` types that 
+have the spec.provisionerClassName field set to `core.rukpak.io/registry+v1`. This field must be set to the correct provisioner name
```
exdx, line 12: the `spec.provisionerClassName` field should be documented on the main README as well, even if we remove references to the `ProvisionerClass` API from there.

provisioner/registryv1/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Running locally 
+
+To experiment with the registryv1 provisioner locally, first setup a local cluster. It's recommended to have OLM installed on the cluster as well, 
+which can be done easily by heading to the [releases](https://github.com/operator-framework/operator-lifecycle-manager/releases) section and installing via the install script. 
+
+Once the cluster is setup and OLM is running, take the following steps from the root of the repository:
+* Create the rukpak namespace via `kubectl create ns rukpak-system` by default
+* Run `make bin/registryv1` to create the registryv1 provisioner binary locally
+* Apply the RukPak core CRDs via `kubectl apply -f manifests/`
```
exdx, line 75: Alternatively they can use the `make install` directive but that currently also installs the plain-k8s provisioner manifests which are not useful in this context

tylerslaton: This probably points to just adding a `make install-registryv1` target (for each other provisioner as well.

provisioner/registryv1/README.md
tylerslaton: Combo is pretty tailor made for this purpose so we should try and see what we can do to get around that. Is this an issue with Combo's implementation or with the registryv1 provisioner?

exdx: Agreed -- I have to look further but I imagine, given that you said you were able to install it successfully via the traditional flow (add the bundle to a catalog, make a subscription, etc), that it's something to do with the provisioner. Not all of the roles/policyrules that are required for combo to work (if the rules I mentioned above are necessary) are being generated.

provisioner/registryv1/README.md
exdx: Yup, sounds good 👍

provisioner/registryv1/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+There is a natural separation between sourcing of the content and application of that content via two separate RukPak APIs, `Bundle` and `BundleInstance`. 
+A user can specify a particular `Bundle` to be available in the cluster for inspection before any application of the resources. 
+Given a `Bundle` resource named `my-bundle`, the registryv1 provisioner will pull down and unpack the bundle to a variety of storage backends, such as ConfigMaps or PersistentVolumes.
+
+Currently the ConfigMap backend is supported by the provisioner, and the content can be inspected by viewing the metadata ConfigMap, which holds
+information on all ConfigMaps related to this bundle. 
+
+`kubectl get cm bundle-metadata-my-bundle -o yaml` provides access to this metadata ConfigMap, which then links to all other ConfigMap objects
+where the unpacked data is stored.
```
joelanford, line 57: There is a [label](https://github.com/operator-framework/rukpak/blob/7cc5fa45e61c2ef324bdbb5a8cf02bdaa42fc527/internal/storage/storage.go#L142) on the configmaps that contains the bundle's `.metadata.name`. That could be used in a label selector to find all of the configmaps.

One unrelated thing that comes to mind though: we might need to make the bundle configmap label key different than the[ label key used in the bundle instance reconciler](https://github.com/operator-framework/rukpak/blob/7cc5fa45e61c2ef324bdbb5a8cf02bdaa42fc527/provisioner/registryv1/controllers/bundleinstance_controller.go#L391).

provisioner/registryv1/README.md
```diff
@@ -0,0 +1,123 @@
...
+### Pivot between bundle versions 
+
+The `BundleInstance` API is meant to indicate the version of the bundle that should be active within the cluster. Given two unpacked bundles in the cluster,
+`my-bundle-v0.0.1` and `my-bundle-v0.0.2`, the `spec.bundleName` field of the related `BundleInstance` can be updated from say, the older version,
+to the newer version, and the registryv1 provisioner will be able to successfully create the new content on-cluster and remove the old content.  
+
+## Running locally 
+
+To experiment with the registryv1 provisioner locally, first setup a local cluster. It's recommended to have OLM installed on the cluster as well, 
+which can be done easily by heading to the [releases](https://github.com/operator-framework/operator-lifecycle-manager/releases) section and installing via the install script.
```
joelanford, line 70: Add an alternative install method for operator-sdk users?
```suggestion
To experiment with the registryv1 provisioner locally, first setup a local cluster. It's recommended to have OLM installed on the cluster as well, 
which can be done easily by heading to the [releases](https://github.com/operator-framework/operator-lifecycle-manager/releases) section and installing via the install script or by running `operator-sdk olm install`. 
```

provisioner/registryv1/README.md
joelanford: In SDK projects, there is:
1. `make install` (just for APIs) 
2. `make deploy` (APIs and controllers from pre-built docker images, controller has configured RBAC)
3. `make run` (just `go run` the main.go, controller has whatever the developer perms are)
4. `make undeploy (delete everything from `make deploy`)
5. `make uninstall` (delete everything from `make install`)

My typical development flow has two main cycles:
1. Update APIs, run `make install` 
2. Update controllers, run `make run`

The third cycle, which I use much more rarely is when I'm changing RBAC for the controller. In that case, it is necessary to build the images and use `make deploy` so that the controller uses a service account tied to that RBAC.

Does that structure (perhaps with better names or tweaks since we have multiple mains) seem reasonable?

Also, have we considered a single main that runs all of our built-in provisioners? We'd get some memory optimization through shared caches.

provisioner/registryv1/README.md
joelanford: This is definitely the registryv1 provisioner not doing everything that the install plan does. Its the weirdness I was talking about that I hacked around in my demo. I'm pretty sure that the install plan generates and applies these things such that they exist by the time the CSV controller is reconciling the CSV.

exdx: Yes, seems likely. I guess the question now is whether we want to transfer over that existing functionality to registryv1 provisioner? I imagine the answer is yes if we want to have a useful provisioner. We want to avoid rebuilding OLM of course but if its simply a matter of generating necessary permissions on install it should be doable. 

I'm not sure if there are other gaps around cert creation/handling as well or if webhooks will work alright.

provisioner/registryv1/README.md
exdx: That seems like a nice setup -- I know we want to cleanup our Makefile a bit once we setup the core provisioners and this seems like a good structure to follow. Once we actually have deployment manifests for a working provisioner it should be easier to wire all this up. 

> Also, have we considered a single main that runs all of our built-in provisioners? We'd get some memory optimization through shared caches.

That's interesting, I can think of pros and cons to that approach. I think the original idea was for everything to be separate (but encourage shared code use), but maybe worth rethinking.

provisioner/registryv1/README.md
exdx: That label is helpful, I will add that information to the doc.

provisioner/registryv1/README.md
exdx: Sounds good -- I can link to the OLM getting started doc, which mentions the sdk method of installation https://olm.operatorframework.io/docs/getting-started/#installing-olm-in-your-cluster

provisioner/registryv1/README.md
exdx: See #65 for tracking issue

exdx:
Closing as we are going to remove the registryv1 provisioner entirely at this stage. Some of the discussion in this PR was useful and may be a reference going forward.

PR #66, opened 2022-02-27
registryv1: Reset the InstalledBundle status field when the Bundle doesn't exist

timflannagan:
Update the registryv1's BundleInstance controller and reset the
status.InstalledBundleName when the corresponding spec.BundleName does
not exist.

This is easy to produce locally:

- Create a Bundle
- Create a BundleInstance that specifies the previously created Bundle
- Delete the Bundle
- Ensure the BundleInstance's status.InstalledBundleName doesn't get reset

### DIFF
--- a/provisioner/registryv1/controllers/bundleinstance_controller.go
+++ b/provisioner/registryv1/controllers/bundleinstance_controller.go
@@ -56,6 +56,10 @@ import (
 	"github.com/operator-framework/rukpak/internal/storage"
 )
 
+const (
+	registryV1ProvisionerID = "core.rukpak.io/registry+v1"
+)
+
 // BundleInstanceReconciler reconciles a BundleInstance object
 type BundleInstanceReconciler struct {
 	client.Client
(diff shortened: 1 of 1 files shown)


provisioner/registryv1/controllers/bundleinstance_controller.go
```diff
@@ -93,21 +96,19 @@ func (r *BundleInstanceReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 		bi.ObjectMeta.ManagedFields = nil
-		if err := r.Status().Patch(ctx, bi, client.Apply, client.FieldOwner("core.rukpak.io/registry+v1")); err != nil {
+		if err := r.Status().Patch(ctx, bi, client.Apply, client.FieldOwner(registryV1ProvisionerID)); err != nil {
 			l.Error(err, "failed to patch status")
 		}
 	}()
 
 	b := &olmv1alpha1.Bundle{}
 	if err := r.Get(ctx, types.NamespacedName{Name: bi.Spec.BundleName}, b); err != nil {
+		bi.Status.InstalledBundleName = ""
```
timflannagan, line 111: Just a quick note: we return an error whenever we encounter a Bundle lookup failure, which means we consistently requeue until we reach the backoff limit. Does it make more sense to return `client.IgnoreNotFound(err)` so we differential a Bundle that doesn't exist (at least in the context of the current cache element) vs. a different error type.

timflannagan: Another implementation could be catching the delete event and resetting this status.InstalledBundleName field.

exdx, line 111: Should we reset the `Status.InstalledBundleName` only when the error is NotFound? If we reset the field on any error, then the following could occur:
* Bundle is on cluster and installed
* BI controller reconciles and makes progress
* Transient error getting the Bundle occurs early in the control loop
* BI `Status.InstalledBundleName` gets reset 

I guess the next time the bundle is found the status is updated, but I'm wondering if instead it makes more sense to only reset the field if the bundle is not found

timflannagan: Yep, I think that makes sense. I updated in the last force push to only reset when apierrors.IsNotFound(...) is true.

exdx (approved):

PR #70, opened 2022-02-28
k8s: Avoid using OLM's cpb tool for unpacking Bundle contents

timflannagan:
Follow-up to #69 and remove the need to use OLM's cpb tooling to copy Bundle contents from scratch-based bundle container images.

### DIFF
--- a/internal/storage/storage.go
+++ b/internal/storage/storage.go
@@ -139,20 +139,24 @@ func (s *ConfigMaps) buildObject(obj client.Object, owner client.Object) (*corev
 	gvk := obj.GetObjectKind().GroupVersionKind()
 
 	labels := map[string]string{
-		"core.rukpak.io/owner-name":       owner.GetName(),
-		"core.rukpak.io/configmap-type":   "object",
+		"core.rukpak.io/owner-name":     owner.GetName(),
+		"core.rukpak.io/configmap-type": "object",
+	}
+	annotations := map[string]string{
 		"core.rukpak.io/object-group":     gvk.Group,
 		"core.rukpak.io/object-version":   gvk.Version,
 		"core.rukpak.io/object-kind":      gvk.Kind,
 		"core.rukpak.io/object-name":      obj.GetName(),
 		"core.rukpak.io/object-namespace": obj.GetNamespace(),
 	}
+
 	immutable := true
 	cm := &corev1.ConfigMap{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      fmt.Sprintf("%sobject-%s-%s", s.NamePrefix, owner.GetName(), hash[0:8]),
-			Namespace: s.Namespace,
-			Labels:    labels,
+			Name:        fmt.Sprintf("%sobject-%s-%s", s.NamePrefix, owner.GetName(), hash[0:8]),
+			Namespace:   s.Namespace,
+			Labels:      labels,
+			Annotations: annotations,
 		},
 		Immutable: &immutable,
 		Data: map[string]string{
--- a/provisioner/k8s/controllers/bundle_controller.go
+++ b/provisioner/k8s/controllers/bundle_controller.go
@@ -202,34 +202,24 @@ func (r *BundleReconciler) ensureUnpackPod(ctx context.Context, bundle *olmv1alp
 		pod.Spec.AutomountServiceAccountToken = &automountServiceAccountToken
 		pod.Spec.Volumes = []corev1.Volume{
 			{Name: "util", VolumeSource: corev1.VolumeSource{EmptyDir: &corev1.EmptyDirVolumeSource{}}},
-			{Name: "bundle", VolumeSource: corev1.VolumeSource{EmptyDir: &corev1.EmptyDirVolumeSource{}}},
 		}
 		pod.Spec.RestartPolicy = corev1.RestartPolicyNever
-		if len(pod.Spec.InitContainers) != 2 {
-			pod.Spec.InitContainers = make([]corev1.Container, 2)
+		if len(pod.Spec.InitContainers) != 1 {
+			pod.Spec.InitContainers = make([]corev1.Container, 1)
 		}
 		pod.Spec.InitContainers[0].Name = "install-cpb"
-		pod.Spec.InitContainers[0].Image = r.CopyBundleImage
-		pod.Spec.InitContainers[0].Command = []string{"/bin/cp", "-Rv", "/bin/cpb", "/util/cpb"}
+		pod.Spec.InitContainers[0].Image = r.UnpackImage
+		pod.Spec.InitContainers[0].Command = []string{"cp", "-Rv", "/unpack", "/util/unpack"}
 		pod.Spec.InitContainers[0].VolumeMounts = []corev1.VolumeMount{{Name: "util", MountPath: "/util"}}
-
-		pod.Spec.InitContainers[1].Name = "copy-bundle"
-		pod.Spec.InitContainers[1].Image = bundle.Spec.Image
-		pod.Spec.InitContainers[1].ImagePullPolicy = corev1.PullAlways
-		pod.Spec.InitContainers[1].Command = []string{"/util/cpb", "--ignore-annotations-file", "/bundle"}
-		pod.Spec.InitContainers[1].VolumeMounts = []corev1.VolumeMount{
-			{Name: "util", MountPath: "/util"},
-			{Name: "bundle", MountPath: "/bundle"},
-		}
 
 		if len(pod.Spec.Containers) != 1 {
 			pod.Spec.Containers = make([]corev1.Container, 1)
 		}
-		pod.Spec.Containers[0].Name = "unpack-bundle"
-		pod.Spec.Containers[0].Image = r.UnpackImage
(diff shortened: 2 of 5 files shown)


provisioner/k8s/controllers/bundle_controller.go
```diff
@@ -347,7 +337,7 @@ func (r *BundleReconciler) getBundleImageDigest(pod *corev1.Pod) (string, error)
 
 func getObjects(bundleFS fs.FS) ([]client.Object, error) {
 	var objects []client.Object
-	const manifestsDir = "manifests"
+	const manifestsDir = "."
```
tylerslaton, line 340: I would expect this to grab from a label or other defined directory. Should we default to expecting the manifests to be in the root of a plain bundle?

internal/storage/storage.go
```diff
@@ -138,13 +139,18 @@ func (s *ConfigMaps) buildObject(obj client.Object, owner client.Object) (*corev
 	}
 	gvk := obj.GetObjectKind().GroupVersionKind()
 
+	objName := obj.GetName()
+	if strings.Contains(objName, ":") {
```
tylerslaton, line 143: Would it make more sense to just let this fail? I ask because this could probably grow to include the remaining things that it takes to validate a label. Alternatively, this could go into an internal package and house more expanded logic. We may not care to make this too elaborate but just want to check that we want this logic to begin with.

tylerslaton (approved):
Small comments to think about before merging.

internal/storage/storage.go
timflannagan: I think erroring out here is the wrong thing to do as the Bundle contents are technically valid, and the problem is with how we're building up a ConfigMap internally. I guess we could produce a warning-level log, but the value of that seems low, and performing a simple translation so it doesn't violate a metadata.Label validation regex shouldn't increase the maintenance burden all that much.

joelanford: Maybe we just move all of the `object-*` fields to annotations? I don't think these are actually used in selectors anywhere.

provisioner/k8s/controllers/bundle_controller.go
```diff
@@ -202,34 +202,24 @@ func (r *BundleReconciler) ensureUnpackPod(ctx context.Context, bundle *olmv1alp
...
-			{Name: "util", MountPath: "/util"},
-			{Name: "bundle", MountPath: "/bundle"},
-		}
-
 		if len(pod.Spec.Containers) != 1 {
 			pod.Spec.Containers = make([]corev1.Container, 1)
 		}
-		pod.Spec.Containers[0].Name = "unpack-bundle"
-		pod.Spec.Containers[0].Image = r.UnpackImage
+		pod.Spec.Containers[0].Name = "copy-bundle"
```
joelanford, line 218: Nit: maybe just `bundle`?
```suggestion
		pod.Spec.Containers[0].Name = "bundle"
```

provisioner/k8s/controllers/bundle_controller.go
joelanford: My two cents: by the time we're calling this function, `bundleFS` should be organized into a consistent view such that we don't need to parse metadata or lookup labels to interpret it.

I'd go even further and say that our plain manifest bundle format should define `/manifests` as _the_ one and only location for manifests. And we if add metadata eventually (I think we'll need to), then we should have a well-defined structure for that as well (specific directories, filenames, and formats)

unpack.Dockerfile
```diff
@@ -17,7 +17,7 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o unpack ./cmd/unpack/
 
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
-FROM gcr.io/distroless/static:nonroot
+FROM gcr.io/distroless/static:debug
```
joelanford, line 20: Is this change intended to be permanent?

provisioner/k8s/controllers/bundle_controller.go
timflannagan: That's fine with me - we parse the name of the copy-bundle container to get the image digest of the bundle.Spec.Image, so I created a constant to track the name of the container that's responsible for unpacking bundle contents to stdout.

internal/storage/storage.go
```diff
@@ -138,13 +139,18 @@ func (s *ConfigMaps) buildObject(obj client.Object, owner client.Object) (*corev
...
+		objName = strings.ReplaceAll(objName, ":", "-")
+	}
+
 	labels := map[string]string{
 		"core.rukpak.io/owner-name":       owner.GetName(),
```
joelanford, line 148: We currently have "core.rukpak.io/owner-name" used in both bundle storage and bundle instance installation, which could result in some incorrect assumptions if a bundle and bundle instance have the same metadata.name.

I think we need to separate these labels (e.g. `core.rukpak.io/bundle-name` and `core.rukpak.io/bundle-instance-name`)

unpack.Dockerfile
timflannagan: If we're copying over the unpack executable to the bundle.Spec.Image container image, which is scratch-based (i.e. no shell), then we need the unpack executable to have a shell so we can fire off `cp` commands for copying over the unpack binary in an initContainer to the container image that unpacks the bundle.Spec.Image.

provisioner/k8s/controllers/bundle_controller.go
timflannagan: Yeah that was the original proposal a couple of quarters back, and we can even introduce tooling for statically validating the bundle format ahead of time, and ensuring that /manifests is the consistent destination container image directory.

internal/storage/storage.go
timflannagan: I think this is reasonable for now, and solves the immediate issue I was having deploying Bundles locally. We can revisit moving some of them back to labels if we need to label select ConfigMaps that reference a specific namespace, or GVK, etc.

timflannagan: Done in the latest force-push.

internal/storage/storage.go
timflannagan: I'm fine with updating to differentiate between bundle and bundleinstance controller resources, but the only issue I see when poking at this code locally is how to handle the manager's cache if we have two different label selecting mechanisms now. I didn't see anything in the `labels` package for exposing a label selector for objects that have the core.rukpak.io/bundle-name label OR the core.rukpak.io/bundleinstance-name label present.

provisioner/k8s/main.go
```diff
@@ -56,9 +56,11 @@ func main() {
 	var enableLeaderElection bool
 	var probeAddr string
 	var systemNamespace string
+	var unpackImage string
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
-	flag.StringVar(&systemNamespace, "system-namespace", "rukpak-system", "Configures the namespace that gets used to deploy system resources")
+	flag.StringVar(&systemNamespace, "system-namespace", "rukpak-system", "Configures the namespace that gets used to deploy system resources.")
+	flag.StringVar(&unpackImage, "unpack-image", "quay.io/tflannag/unpack:latest", "Configures the container image that gets used to unpack Bundle contents.")
```
timflannagan, line 63: Note: we'll need to create a dedicated container image in the operator-framework quay organization for the bundle unpacking component.

internal/storage/storage.go
timflannagan: I think I'm leaning towards handling this in a follow-up in either case as this PR fixes a couple of things at this point. Thoughts?

unpack.Dockerfile
joelanford: Oh right. Thanks.

joelanford: Maybe add a comment about why we need `:debug` since it isn't obvious just looking at the Dockerfile.

internal/storage/storage.go
joelanford: Follow-up is fine. These labels would be used on separate types, so we should be able to use different KV pairs for `SelectorsByObject` in the `cache.Options`.

timflannagan: Alright sounds good to me - I'll track this locally and try to tackle this tomorrow.

timflannagan:
I'm going to track the open conversations locally vs. creating individual issues for them. I opened https://github.com/operator-framework/rukpak/issues/75 for documenting the plain bundle format.

internal/storage/storage.go
joelanford: Actually, you're totally right. This isn't a selector on `Bundle` and `BundleInstance` objects. This is a selector on the rest of the objects. And I think that's the single `dependentSelector`. Hmm...

joelanford: Added a new `core.rukpak.io/owner-kind` label. And an `In` selector on `Bundle` or `BundleInstance` to find objects with either value. https://github.com/operator-framework/rukpak/pull/76

PR #71, opened 2022-02-28
doc: outline plain-v0 provisioner

### LINKED ISSUE #60: Add documentation around the plain manifest bundle provisioner

exdx:
Closes #60 

Provides documentation on the newly introduced plain-v0 provisioner via the provisioner scoped README

### DIFF
--- a/provisioner/plain-v0/README.md
+++ b/provisioner/plain-v0/README.md
@@ -1,3 +1,202 @@
-# plain-v0
+# Plain-v0 Provisioner
 
-A concrete Provisioner implementation for "plain" k8s bundles.
+## Summary
+
+The `plain-v0` provisioner is a core rukpak provisioner that knows how to interact with `plain+v0` format bundles.
+The `plain+v0` bundles, or plain bundles, are simply container images containing a set of static kubernetes YAML
+manifests in a given directory. For example, below is an example Dockerfile that builds a plain `plain+v0` bundle from a
+directory containing static manifests.
+
+```dockerfile
+FROM scratch
+COPY /manifests /manifests
+```
+
+where the given `manifests` directory contains the kubernetes resources required to deploy an application, for example:
+
+```
+manifests
+├── namespace.yaml
+├── cluster_role.yaml
+├── role.yaml
+├── serviceaccount.yaml
+├── cluster_role_binding.yaml
+├── role_binding.yaml
+└── deployment.yaml
+```
+
+Note: the static manifests must be located in the root-level /manifests directory for the bundle image to be a
+valid `plain+v0` bundle that the provisioner can unpack.
+
+The `plain-v0` provisioner is able to unpack a given `plain+v0` bundle onto a cluster and then instantiate it, making
+the content of the bundle available in the cluster. It does so by reconciling `Bundle` and `BundleInstance` types that
+have the `spec.provisionerClassName` field set to `core.rukpak.io/plain-v0`. This field must be set to the correct
+provisioner name in order for the `plain-v0` provisioner to see and interact with the bundle.
+
+Below is an example of the provisioner reconciliation flow:
+
+```mermaid
+graph TD
+    C[Provisioner]
+    C -->|Watches| D[Bundle]
+    C -->|Watches| E[BundleInstance]
+    D -->|References| F[Content]
+    E -->|Creates/Manages| F[Content]
+```
+
+## Use cases
+
+### Install and apply a specific version of a bundle
+
+The `plain-v0` provisioner can install and make available a specific `plain+v0` bundle in the cluster.
+
+Simply create a `Bundle` resource pointing to a specific version of your bundle, and a `BundleInstance` which references
+that bundle. The `plain-v0` provisioner will unpack the provided Bundle onto the cluster, and eventually make the
+content available on the cluster.
+
+```yaml
+apiVersion: core.rukpak.io/v1alpha1
+kind: Bundle
+metadata:
+  name: my-bundle
+spec:
+  image: my-bundle@sha256:xyz123
+  provisionerClassName: core.rukpak.io/plain-v0
+---
+apiVersion: core.rukpak.io/v1alpha1
+kind: BundleInstance
+metadata:
+  name: my-bundle-instance
+spec:
+  provisionerClassName: core.rukpak.io/plain-v0
+  bundleName: my-bundle
+``` 
+
+First, the Bundle will be in the Pending stage as the provisioner sees it and begins unpacking the referenced content:
+
+```
+NAME          IMAGE                                        PHASE     AGE
+my-bundle     my-bundle@sha256:xyz123                      Pending   3s
+```
+
+Then eventually, as the bundle content is unpacked onto the cluster via the defined storage mechanism, the bundle status
+will be updated to Unpacked, indicating that all its contents have been stored on-cluster.
+
+```
+NAME          IMAGE                                        PHASE      AGE
+my-bundle     my-bundle@sha256:xyz123                      Unpacked   10s
+```
+
+Now that the bundle has been unpacked, the provisioner is able to create the resources in the bundle on the cluster.
+These resources will be owned by the corresponding BundleInstance. Creating the BundleInstance on-cluster results in an
+InstallationSucceeded Phase if the application of resources to the cluster was successful.
+
+```
+NAME                 DESIRED BUNDLE   INSTALLED BUNDLE   INSTALL STATE           AGE
+my-bundle-instance   my-bundle        my-bundle          InstallationSucceeded   11s
+```
+
+### Make bundle content available but do not install it
+
+There is a natural separation between sourcing of the content and application of that content via two separate RukPak
+APIs, `Bundle` and `BundleInstance`. A user can specify a particular `Bundle` to be available in the cluster for
+inspection before any application of the resources. Given a `Bundle` resource named `my-bundle`, the plain-v0
+provisioner will pull down and unpack the bundle to a set of ConfigMaps.
+
+By default, `rukpak-system` is the configured namespace for deploying `plain-v0` provisioner-related system resources.
+
+Surfacing the content of a bundle in a more user-friendly way, via a plugin or additional API, is on the RukPak roadmap.
+
+### Pivoting between bundle versions
+
+The `BundleInstance` API is meant to indicate the version of the bundle that should be active within the cluster. Given
+two unpacked bundles in the cluster,
+`my-bundle-v0.0.1` and `my-bundle-v0.0.2`, the `spec.bundleName` field of the related `BundleInstance` can be updated to
+pivot to a different version. For example by pivoting from an older version, to the newer version, by updating
+the `spec.bundleName` field the `plain-v0` provisioner will create the new bundle content on-cluster and remove the old
+content. The provisioner also continually reconciles the created content via dynamic watches to ensure that all
+resources referenced by the bundle are present on the cluster.
+
+## Running locally
+
+To experiment with the `plain-v0` provisioner locally, first setup a local cluster, or simply
+have [kind](https://kind.sigs.k8s.io/) installed locally.
+
+Once the cluster has been setup, take the following steps:
+
+* Clone the repository via `git clone https://github.com/operator-framework/rukpak`
+* Navigate to the repository via `cd rukpak`
+* Run `make run` to build and deploy the provisioner onto the local cluster.
+
+From there, create some `Bundles` and `BundleInstance` types to see the provisioner in action. For an example bundle to
+use, the [combo operator](https://github.com/operator-framework/combo) is a good candidate.
+
+Create the bundle:
+
+```yaml
+apiVersion: core.rukpak.io/v1alpha1
+kind: Bundle
+metadata:
+  name: combo-0.0.1
+spec:
+  image: quay.io/tyslaton/combo-plain-bundle:v0.0.1
+  provisionerClassName: core.rukpak.io/plain-v0
+```
+
+Check the Bundle status via `kubectl get bundle combo-0.0.1`. Eventually the Bundle should show up as Unpacked.
+
+```
+NAME          IMAGE                                        PHASE      AGE
+combo-0.0.1   quay.io/tyslaton/combo-plain-bundle:v0.0.1   Unpacked   17h
+```
+
+Create the combo `BundleInstance` referencing the combo `Bundle` available in the cluster.
+
+Note: it's perfectly valid to create a `Bundle` and `BundleInstance` referencing that `Bundle` at the same time. There
+is no specific ordering required, which is perfect for GitOps flows, and eventually both the `Bundle`
+and `BundleInstance`
+would succeed.
+
+```yaml
+apiVersion: core.rukpak.io/v1alpha1
+kind: BundleInstance
+metadata:
+  name: combo
+spec:
+  provisionerClassName: core.rukpak.io/plain-v0
+  bundleName: combo-0.0.1
+```
+
+Check the BundleInstance status to ensure that the installation was successful.
+
+```
+NAME    DESIRED BUNDLE   INSTALLED BUNDLE   INSTALL STATE           AGE
+combo   combo-0.0.1      combo-0.0.1        InstallationSucceeded   16h
+```
+
+From there, check out the combo operator deployment and ensure that the operator is present on the cluster.
+
+```shell
+kubectl -n combo get deployments.apps combo-operator -o yaml 
+```
+
+The operator should be successfully installed. To test out how the `plain-v0` provisioner continually reconciles
+BundleInstance resources, try deleting the combo deployment.
+
+```shell
+kubectl -n combo delete deployments.apps combo-operator
+```
+
+Checking for the deployment again, it will be back on the cluster. The provisioner ensures that all resources required
+for the BundleInstance to run are accounted for on-cluster.
+
+To cleanup from the installation, simply remove the BundleInstance from the cluster. This will remove all references
+resources including the deployment, RBAC, and the operator namespace. Then the Bundle can be removed as well, which
+removes any unpacked manifests from the cluster.
+
+```shell
+kubectl delete bundleinstances.core.rukpak.io combo
+kubectl delete bundle combo-0.0.1
+```
+
+The cluster state is now the same as it was prior to installing the operator. 


README.md
```diff
@@ -13,7 +13,7 @@ content is being installed on-cluster and how to create a running instance of th
 
 ## Components 
 
-RukPak is composed of a few primary APIs: `Bundle`, `Instance`, and `ProvisionerClass`
+RukPak is composed of a few primary APIs: `Bundle`, `BundleInstance`, and `Provisioner`
```
timflannagan, line 16: nit: there's no Provisioner API.

README.md
```diff
@@ -22,80 +22,47 @@ Much like the contents of a container image need to be pulled and unpacked in or
...
+  image: quay.io/tyslaton/combo-plain-bundle:v0.0.1
+  provisionerClassName: core.rukpak.io/plain
 ```
 
-### Instance 
-The `Instance` API points to a Bundle and indicates that it should be “active”. This includes pivoting from older versions of an active bundle. 
-`Instance` may also include an embedded spec for a desired Bundle.
+### BundleInstance 
+The `BundleInstance` API points to a Bundle and indicates that it should be “active”. This includes pivoting from older versions of an active bundle. 
+`BundleInstance` may also include an embedded spec for a desired Bundle.
```
timflannagan, line 41: Is it worth mentioning that the embedded Bundle spec won't be present until a future release?

README.md
timflannagan: another nit: trailing period.

provisioner/k8s/README.md
```diff
@@ -1,3 +1,112 @@
...
+apiVersion: core.rukpak.io/v1alpha1
+kind: BundleInstance
+metadata:
+  name: combo-system
+spec:
+  provisionerClassName: core.rukpak.io/plain
+  bundleName: combo-0.0.1
+```
+
+From there, check out the combo operator and ensure that the CSV is present on the cluster.
```
timflannagan, line 112: nit: CSV is present on the cluster?

provisioner/k8s/README.md
```diff
@@ -1,3 +1,112 @@
...
+### Make bundle content available but do not install it
+
+There is a natural separation between sourcing of the content and application of that content via two separate RukPak APIs, `Bundle` and `BundleInstance`. 
+A user can specify a particular `Bundle` to be available in the cluster for inspection before any application of the resources. 
+Given a `Bundle` resource named `my-bundle`, the k8s provisioner will pull down and unpack the bundle to a variety of storage backends, such as ConfigMaps or PersistentVolumes.
+
+Currently the ConfigMap backend is supported by the provisioner, and the content can be inspected by viewing the metadata ConfigMap, which holds
+information on all ConfigMaps related to this bundle. 
+
+`kubectl -n rukpak-system get cm --selector core.rukpak.io/owner-name=<bundle-name>` gets all ConfigMaps related to the `Bundle` on-cluster.
```
timflannagan, line 54: Can we make a note that the `rukpak-system` is the default namespace for deploying provisioner-related system resources? I've been playing around with making this namespace configurable in #72.

provisioner/k8s/README.md
```diff
@@ -1,3 +1,112 @@
...
+
+Surfacing the content of a bundle in a more user-friendly way, via a plugin or additional API, is on the RukPak roadmap.
```
timflannagan, line 56: We have support for custom columns when querying for Bundles through the CLI. Do you think it's worth showing the output of those Bundles when we walk through the examples you added?

provisioner/k8s/README.md
timflannagan: Also: any value in outputting a sample Bundle metadata ConfigMap?

provisioner/k8s/README.md
```diff
@@ -1,3 +1,112 @@
...
+## Summary
+
+The k8s provisioner is a core rukpak provisioner that knows how to interact with `k8s` format bundles. 
+The `k8s` bundles, or plain bundles, are simply container images containing a set of static kubernetes YAML manifests in a given directory.  
+For example, below is an example Dockerfile that builds a plain `k8s` bundle from a directory containing static manifests.  
+```dockerfile
+FROM scratch
+LABEL manifests="/manifests"
+COPY ./manifests /manifests
+```
```
timflannagan, line 12: I think this example is a bit more clearer if we showed off the local directory structure (e.g. using `tree ./manifests`) so the user gets an idea on what we're copying in the container. Any thoughts? This can be done in a follow-up if we think that it's clearer.

README.md
```diff
@@ -22,80 +22,47 @@ Much like the contents of a container image need to be pulled and unpacked in or
...
 spec:
-  class: rukpack.io/k8s
-  refs:
-  - file://content
-  volumeMounts:
-  - mountPath: /content
-    configMap:
-      name: local 
-      namespace: plumbus
+  image: quay.io/tyslaton/combo-plain-bundle:v0.0.1
```
tylerslaton, line 35: Mental note to come back here once we have a more official plain bundle format.

provisioner/k8s/README.md
```diff
@@ -1,3 +1,112 @@
...
+
+Surfacing the content of a bundle in a more user-friendly way, via a plugin or additional API, is on the RukPak roadmap. 
+
+### Pivot between bundle versions 
+
+The `BundleInstance` API is meant to indicate the version of the bundle that should be active within the cluster. Given two unpacked bundles in the cluster,
+`my-bundle-v0.0.1` and `my-bundle-v0.0.2`, the `spec.bundleName` field of the related `BundleInstance` can be updated to pivot to a different version. For example by pivoting from an older version,
+to the newer version, by updating the `spec.bundleName` field the k8s provisioner will create the new bundle content on-cluster and remove the old content.  
+
+## Running locally
```
timflannagan, line 64: It would nice to see an e2e workflow of creating the Bundle image from scratch from a local directory of static YAML manifests, pushing that built image to a container registry, and then creating the requisite Bundle and BundleInstance resources (but this can be done as a follow-up as well).

provisioner/k8s/README.md
timflannagan, line 112: This might be an opportunity to show off the Bundle custom column output, and instructing the user to verify the `PHASE` column is reporting an Unpacked phase, before we create the BundleInstance resource. Thoughts?

README.md
```diff
@@ -22,80 +22,47 @@ Much like the contents of a container image need to be pulled and unpacked in or
...
-A `ProvisionerClass` specifies a specific configuration of provisioning (i.e. settings to interpret Instance and Bundle APIs). 
-A `Provisioner` will only operate on Instance and Bundle objects that reference a `ProvisionerClass` that contain the provisioner’s unique id.
-
-If this seems familiar, it is the same pattern that is used by StorageClass / PeristentVolume / PersistentVolumeClaim in Kubernetes. This design is meant to be extendable 
-to support `Bundles` and `Instances` backed by different content sources. 
+### Provisioner
+A Provisioner is a controller that understands `BundleInstance` and `Bundle` APIs and can take action. Each `Provisioner` is assigned a unique ID, and is responsible for reconciling
+a `Bundle` and `BundleInstance` with a `spec.provisionerClassName` that matches that particular ID. For example, in this repository the [plain k8s](provisioner/k8s/README.md) provisioner is implemented. 
+The `k8s` provisioner is able to unpack a given `k8s` bundle onto a cluster and then instantiate it, making the
+content of the bundle available in the cluster.
```
tylerslaton, line 63: super nit: adding new lines to README's makes it weird to look at on smaller screens.

```suggestion
A Provisioner is a controller that understands `BundleInstance` and `Bundle` APIs and can take action. Each `Provisioner` is assigned a unique ID, and is responsible for reconciling a `Bundle` and `BundleInstance` with a `spec.provisionerClassName` that matches that particular ID. For example, in this repository the [plain k8s](provisioner/k8s/README.md) provisioner is implemented. The `k8s` provisioner is able to unpack a given `k8s` bundle onto a cluster and then instantiate it, making the content of the bundle available in the cluster. 
```

provisioner/k8s/README.md
```diff
@@ -1,3 +1,112 @@
-# k8s
+# k8s Provisioner
```
tylerslaton, line 1: non-blocking question: I've been thinking about this for a bit now, but does it still make sense to call this the `k8s` provisioner? Personally it feels more like the `default` or `plain` nomenclature makes more sense.

provisioner/k8s/README.md
```diff
@@ -1,3 +1,112 @@
...
 
-A concrete Provisioner implementation for "plain" k8s bundles.
+## Summary
+
+The k8s provisioner is a core rukpak provisioner that knows how to interact with `k8s` format bundles. 
+The `k8s` bundles, or plain bundles, are simply container images containing a set of static kubernetes YAML manifests in a given directory.  
+For example, below is an example Dockerfile that builds a plain `k8s` bundle from a directory containing static manifests.
```
tylerslaton, line 7: generalized nit: going back to that newline thing, feel free to ignore. It isn't blocking.

```suggestion
The k8s provisioner is a core rukpak provisioner that knows how to interact with `k8s` format bundles. The `k8s` bundles, or plain bundles, are simply container images containing a set of static kubernetes YAML manifests in a given directory.  For example, below is an example Dockerfile that builds a plain `k8s` bundle from a directory containing static manifests.
```

provisioner/k8s/README.md
```diff
@@ -1,3 +1,112 @@
...
+For example, below is an example Dockerfile that builds a plain `k8s` bundle from a directory containing static manifests.  
+```dockerfile
+FROM scratch
+LABEL manifests="/manifests"
+COPY ./manifests /manifests
+```
+
+The `k8s` provisioner is able to unpack a given `k8s` bundle onto a cluster and then instantiate it, making the
+content of the bundle available in the cluster. It does so by reconciling the `Bundle` and `BundleInstance` types that 
+have the `spec.provisionerClassName` field set to `core.rukpak.io/plain`. This field must be set to the correct provisioner name
```
tylerslaton, line 16: +1 to the idea of renaming this to the `plain` provisioner since that is its className.

provisioner/k8s/README.md
```diff
@@ -1,3 +1,112 @@
...
+
+### Install and apply a specific version of a bundle 
+
+The k8s provisioner can install and make available a specific `k8s` bundle available in the cluster. 
+
+Simply create a `Bundle` resource pointing to a specific version of your bundle, and a `BundleInstance` which references that bundle.
+The k8s provisioner will unpack the provided Bundle onto the cluster, and eventually make the content available on the cluster. 
+
+```yaml
+apiVersion: core.rukpak.io/v1alpha1
```
tylerslaton, line 29: Did we test this out with the plain Combo bundle? Maybe it makes sense to replace it into here like with the root README.

provisioner/k8s/README.md
tylerslaton: +1

provisioner/k8s/README.md
tylerslaton: Ah nevermind, we use it in a local example below :)

provisioner/k8s/README.md
```diff
@@ -1,3 +1,112 @@
...
+kind: Bundle
+metadata:
+  name: combo-0.0.1
+spec:
+  image: quay.io/tyslaton/combo-plain-bundle:v0.0.1
+  provisionerClassName: core.rukpak.io/plain
+```
+
+Before creating the combo `BundleInstance` on-cluster, ensure that there is a namespace available that
```
tylerslaton, line 89: Was there an issue with running the bundle where it didn't have the namespace available? That bundle has a manifest for one so it should come with it.

README.md
exdx: yes correct, its not technically an API, but a concept worth describing here. Will update.

provisioner/k8s/README.md
exdx: whoops, that got through from the old doc

README.md
exdx: maybe it makes the most sense to remove that for now and then add it back once the functionality becomes available -- do we have a ticket tracking that feature?

provisioner/k8s/README.md
exdx: sure, that sounds good 👍

README.md
timflannagan: Yeah that sounds reasonable to me - I don't think we have a tracking ticket anywhere.

provisioner/k8s/README.md
joelanford: Also, this and similar lines may need to change based on my comment here: https://github.com/operator-framework/rukpak/pull/70/files#r816268935

provisioner/k8s/README.md
exdx: yes, it could be helpful to show the custom columns from `kubectl get bundles`, will add to the doc.

provisioner/k8s/README.md
joelanford: > make a note
I assume that our manifests will use the `rukpak-system` namespace? If so, maybe it isn't necessary in the context of the readme to get into the detail about the namespace being configurable?

provisioner/k8s/README.md
exdx: That makes sense, I think we can include that. What we really need is a plain bundle spec to link to, some documentation that discusses this bundle format. The k8s provisioner and the plain bundle spec are related but different, in the doc we are conflating the two because there is no plain bundle format doc yet

README.md
exdx: Yes, ideally these example images would live in the `operator-framework` repository

provisioner/k8s/README.md
exdx: Yes, sounds good

provisioner/k8s/README.md
exdx: Yes, now that we have the repository structure cleaned up and testing locally is easier, the local example can be expanded.

provisioner/k8s/README.md
exdx: Right now there is some inconsistency between k8s and plain in the docs. I personally don't like "plain" or "k8s" that much. We need to iron out our naming convention for this provisioner and the plain bundle spec.

provisioner/k8s/README.md
timflannagan: Yep completely agree.

provisioner/k8s/README.md
tylerslaton: We could use the process that I did on Combo here, if it is valuable.

joelanford: Random side note (and not an immediate concern) that I'm reminded of (and I think this is touched on in the strawman): It would be nice if the bundle API allowed for a configurable source (e.g. image vs configmap vs volume vs s3, etc.).

It seems like we might want a `source` union type that's currently just populated with an underlying `image` source.
```yaml
apiVersion: rukpak.io/v1alpha1
kind: Bundle
metadata:
  name: mycompany-bundle-v1.2.3
spec:
  source:
    image:
      ref: "quay.io/mycompany/mycompany-bundle:v1.2.3"
```

provisioner/k8s/README.md
tylerslaton: Agreed, I don't feel that this is blocking. We can address it later but I think it is worth getting a clear name for it.

provisioner/k8s/README.md
exdx: oh nice, ok I can remove that step. I think it's leftover from the registry+v1 bundle steps

provisioner/k8s/README.md
exdx: I guess so, just for consistency 👍

provisioner/k8s/README.md
joelanford: +1 ... However, technically it doesn't have to be unpacked before creating the BundleInstance (things will keep retrying), so I'd only ask that whatever wording we come up with doesn't make it seem like the bundle _must_ be unpacked before the BundleInstance is created.

README.md
joelanford: Whatever we chose, at least be consistent in the file. I wonder if there's a markdown linter that we can use for something like that?

fwiw, my preference is newlines, but with a strict column cutoff (e.g. 100 chars), but i fully realize this is like a tabs vs spaces flamewar, and I'm not interested in starting one here :rofl:

provisioner/k8s/README.md
joelanford: I think we need a better name than k8s for sure. Both registryv1 and this provisioner (and probably almost all others) are deploying to k8s. The name should more clearly reflect what the source content is.

README.md
timflannagan: I just created https://github.com/operator-framework/rukpak/issues/73 btw before we lose this thread in the sea of comments.

provisioner/k8s/README.md
timflannagan: Do we need a ticket that tracks that want/need for a more descriptive plain bundle provisioner?

provisioner/k8s/README.md
timflannagan: Yeah that's a good point. I still think it's worth noting the `rukpak-system` namespace is the default namespace before we start documenting kubectl commands. None of my comments were blocking though.

provisioner/k8s/README.md
timflannagan: @joelanford I created the #92 so maybe we can copy the want/need for a union type in that issue?

provisioner/k8s/README.md
timflannagan: tracking this in #87

timflannagan (commented):
Can we split up the README, and introducing the plain+v0 provisioner documentation into separate PRs? It should be easier to push this kind of work through when the overall set of changes are more tightly scoped imo.

provisioner/k8s/README.md
exdx: I kept the mention of the default namespace but did not mention overriding it since our manifests are hardcoded. 

Also decided not to output the metadata configmap as that feels pretty low-level and not part of a top-level README necessarily.

provisioner/k8s/README.md
exdx: I've added a tree example to this section, but will make another pass once we merge the plain+v0 bundle spec documentation to the repository so we can link to there

README.md
exdx: Going to use this image for now as it's the only semi-official plain+v0 out there right now

README.md
exdx: I looked for a markdown linter that we can add to our existing lint commands but didn't find one. I ended up just using the markdown linter in my IDE. Hopefully that will address any concerns for now.

provisioner/plain-v0/README.md
```diff
@@ -1,3 +1,197 @@
...
+```
+
+The `plain-v0` provisioner is able to unpack a given `plain+v0` bundle onto a cluster and then instantiate it, making
+the content of the bundle available in the cluster. It does so by reconciling `Bundle` and `BundleInstance` types that
+have the `spec.provisionerClassName` field set to `core.rukpak.io/plain-v0`. This field must be set to the correct
+provisioner name in order for the `plain-v0` provisioner to see and interact with the bundle.
+
+Below is an example of the provisioner reconciliation flow:
+
+```mermaid
```
tylerslaton, line 39: This is fantastic, really good to see here.

timflannagan: TIL that github has native support for this now. That's pretty neat.

provisioner/plain-v0/README.md
```diff
@@ -1,3 +1,197 @@
...
+The `plain+v0` bundles, or plain bundles, are simply container images containing a set of static kubernetes YAML
+manifests in a given directory. For example, below is an example Dockerfile that builds a plain `plain+v0` bundle from a
+directory containing static manifests.
+
+```dockerfile
+FROM scratch
+LABEL core.rukpak.io.bundle.mediatype.v1=plain+v0
+LABEL core.rukpak.io.bundle.manifests.v1=/manifests
+LABEL core.rukpak.io.bundle.package.v1=my-package
+LABEL core.rukpak.io.bundle.version.v1=0.0.1
```
timflannagan, line 15: nit: can we lose this labels right now as we're not actively using them anywhere in the provisioner?

provisioner/plain-v0/README.md
```diff
@@ -1,3 +1,197 @@
...
+COPY ./manifests /manifests
```
timflannagan, line 16: can we add a note that we explicitly need the directory of manifests to live in the /manifests for the provisioner to work?

provisioner/plain-v0/README.md
```diff
@@ -1,3 +1,197 @@
...
+
+### Make bundle content available but do not install it
+
+There is a natural separation between sourcing of the content and application of that content via two separate RukPak
+APIs, `Bundle` and `BundleInstance`. A user can specify a particular `Bundle` to be available in the cluster for
+inspection before any application of the resources. Given a `Bundle` resource named `my-bundle`, the plain-v0
+provisioner will pull down and unpack the bundle to a variety of storage backends, such as ConfigMaps or
+PersistentVolumes.
+
+By default, `rukpak-system` is the default namespace for deploying plain-v0 provisioner-related system resources.
```
timflannagan, line 108: :+1:

provisioner/plain-v0/README.md
```diff
@@ -1,3 +1,197 @@
...
+resources referenced by the bundle are present on the cluster.
+
+## Running locally
+
+To experiment with the `plain-v0` provisioner locally, first setup a local cluster, or simply
+have [kind](https://kind.sigs.k8s.io/) installed locally.
+
+Once the cluster has been setup, take the following step from the root of the repository:
+
+* Run `make run` to build and deploy the provisioner onto the local cluster.
```
timflannagan, line 129: Ensure the repository has been cloned before running this?

provisioner/plain-v0/README.md
```diff
@@ -1,3 +1,197 @@
...
+```
+
+Check the Bundle status via `kubectl get bundle combo-0.0.1`. Eventually the Bundle should show up as Unpacked.
+
+```
+NAME          IMAGE                                        PHASE      AGE
+combo-0.0.1   quay.io/tyslaton/combo-plain-bundle:v0.0.1   Unpacked   17h
+```
+
+Create the combo `BundleInstance` referencing the combo `Bundle` available in the cluster.
```
timflannagan, line 153: follow-up: mention we can create a set of Bundles/BundleInstances at the same time (e.g. it works with gitops and doesn't have any resource ordering requirements like OLM).

timflannagan (approved):
Awesome work. This reads much clearer since the first iteration. I didn't see anything blocking, but I'd like to call out some of the rough edges in follow-ups (or down the line):

- BundleInstance references Bundles with existing resources
- BundleInstance references Bundles that contain CRDs + instances of those CRDs
- BundleInstance references Bundles that don't contain a namespace manifest
- BundleInstance references Bundles that contain resources with templating

Maybe this can be an FAQ?

provisioner/plain-v0/README.md
```diff
@@ -1,3 +1,197 @@
...
+content available on the cluster.
+
+```yaml
+apiVersion: core.rukpak.io/v1alpha1
+kind: Bundle
+metadata:
+  name: my-bundle
+spec:
+  image: my-bundle@sha256:xyz123
+  provisionerClassName: core.rukpak.io/plain-v0
```
tylerslaton, line 65: Something I've been confused by with this specific field - why is it "class name" wouldn't "provisioner" make more sense? This is more of a conversation fit for our slack channels but wanted to bring it up regardless.

provisioner/plain-v0/README.md
```diff
@@ -1,3 +1,197 @@
...
+NAME          IMAGE                                        PHASE     AGE
+my-bundle     my-bundle@sha256:xyz123                      Pending   3s
+```
+
+Then eventually, as the bundle content is unpacked onto the cluster via the defined storage mechanism, the bundle status
+will be updated to Unpacked, indicating that all its contents have been stored on-cluster.
+
+```
+NAME          IMAGE                                        PHASE      AGE
+my-bundle     my-bundle@sha256:xyz123                      Unpacked   13m
```
tylerslaton, line 88: Does it normally take this long to unpack?

provisioner/plain-v0/README.md
tylerslaton, line 108:
```suggestion
By default, `rukpak-system` is the namespace for deploying plain-v0 provisioner-related system resources.
```

provisioner/plain-v0/README.md
```diff
@@ -1,3 +1,197 @@
...
+```
+
+Checking for the deployment again, it will be back on the cluster. The provisioner ensures that all resources required
+for the BundleInstance to run are accounted for on-cluster.
+
+To cleanup from the installation, simply remove the BundleInstance from the cluster. This will remove all references
+resources including the deployment, RBAC, and the operator namespace. Then the Bundle can be removed as well, which
+removes any unpacked manifests from the cluster.
+
+```shell
```
tylerslaton, line 192: 🎉

tylerslaton (approved):
This looks great, only had a few small nits but they aren't blocking.

provisioner/plain-v0/README.md
```diff
@@ -1,3 +1,197 @@
...
+
+```
+manifests
+├── 00_namespace.yaml
+├── 01_cluster_role.yaml
+├── 02_role.yaml
+├── 03_serviceaccount.yaml
+├── 04_cluster_role_binding.yaml
+├── 05_role_binding.yaml
+└── 06_deployment.yaml
```
joelanford, line 29: This is _currently_ misleading since the apply ordering is not preserved. The plain-v0 provisioner currently delegates to helm's defined ordering.

provisioner/plain-v0/README.md
```diff
@@ -1,3 +1,197 @@
-# plain-v0
+# Plain-v0 Provisioner
 
-A concrete Provisioner implementation for "plain" k8s bundles.
+## Summary
+
+The `plain-v0` provisioner is a core rukpak provisioner that knows how to interact with `plain+v0` format bundles.
```
joelanford, line 5: It seems confusing to me that most things use `plain-v0`, but we're calling the format `plain+v0`. Should we make it `plain-v0` everywhere for consistency?

akihikokuroda, line 5: It may help to have a link here to https://github.com/operator-framework/rukpak/blob/main/README.md#provisionerclass for what the `rukpak provisioner` is.

akihikokuroda (commented):
The contents are clear and easy to understand.  I have a few comments.  Thanks!

provisioner/plain-v0/README.md
```diff
@@ -1,3 +1,197 @@
...
+
+Below is an example of the provisioner reconciliation flow:
+
+```mermaid
+graph TD
+    C[Provisioner]
+    C -->|Watches| D[Bundle]
+    C -->|Watches| E[BundleInstance]
+    D -->|References| F[Content]
+    E -->|Creates/Manages| F[Content]
```
joelanford, line 45: At a high level, this flow makes sense, but the fact that there are arrows into the same content from both bundles and bundle instances could be confusing.

Technically, what's happening is that anytime the BI is reconciled it checks to see if `Content` is applied. If it isn't, it effectively creates a snapshot of the `Content` and applies that.

provisioner/plain-v0/README.md
joelanford: The main reason is that there will likely be a `ProvisionerClass` API in the future. So this field will eventually be the name of another `ProvisionerClass` object, and _that_ object will contain the name of a specific provisioner along with the configuration that should be used with that provisioner.

That setup will allow the same provisioner to be used with different configurations by creating two separate `ProvisionerClass` objects.

provisioner/plain-v0/README.md
joelanford: No. This is a good point. It might make sense to do `kubectl get bundles -w` and grab the relevant lines from that output to better reflect the time it takes.

In my tests, its usually just a few seconds (but its dependent on whether the image needs to be pulled and how big the image is)

provisioner/plain-v0/README.md
```diff
@@ -1,3 +1,197 @@
...
+my-bundle-instance   my-bundle        my-bundle          InstallationSucceeded   11s
+```
+
+### Make bundle content available but do not install it
+
+There is a natural separation between sourcing of the content and application of that content via two separate RukPak
+APIs, `Bundle` and `BundleInstance`. A user can specify a particular `Bundle` to be available in the cluster for
+inspection before any application of the resources. Given a `Bundle` resource named `my-bundle`, the plain-v0
+provisioner will pull down and unpack the bundle to a variety of storage backends, such as ConfigMaps or
+PersistentVolumes.
```
joelanford, line 106:
```suggestion
provisioner will pull down and unpack the bundle to a set of ConfigMaps.
```

provisioner/plain-v0/README.md
exdx: Yeah it doesn't, it was just sitting around on my cluster for a while 😄 good catch

provisioner/plain-v0/README.md
```diff
@@ -1,3 +1,197 @@
...
+```
+
+Create the combo `BundleInstance` referencing the combo `Bundle` available in the cluster.
+
+```yaml
+apiVersion: core.rukpak.io/v1alpha1
+kind: BundleInstance
+metadata:
+  name: combo-system
+spec:
```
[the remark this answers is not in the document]
exdx: Good catch

provisioner/plain-v0/README.md
exdx: Yes, I removed the number prefix so as not to be misleading.

provisioner/plain-v0/README.md
exdx: From what I understood we were doing that intentionally to separate between the provisioner type and the bundle format. Since we have the existing `registry+v1` format, calling this format `plain+v0` seems consistent.

exdx: I agree we could link back to the README, but we will remove the ProvisionerClass from the top-level README at the moment since we have not implemented that API yet

provisioner/plain-v0/README.md
tylerslaton: Ah okay, I see that makes a lot more sense now. Thanks Joe - resolving.

provisioner/plain-v0/README.md
exdx: Yes, I played around with some additional components to the chart, but fellback to something pretty simple at a high level. I think we could build this chart out to be more robust in a follow-up, I'd like to show the Bundle connected to backend storage and the BI connected to running components on-cluster

PR #72, opened 2022-02-28
internal,provisioner: Support configuring the system namespace

timflannagan:
Introduce a flag to the k8s provisioner binary for configuring the
namespace that's used to house system resources. Add a helper function
that checks whether the provisioner is running in a Pod vs. locally, and
defaults to the Pod's serviceaccount namespace file if present.


tylerslaton (commented):
/lgtm

PR #77, opened 2022-03-01
internal,provisioner: Fix how Bundle status conditions are applied

timflannagan:
Update the EnsureCondition helper function, and avoid setting the
Bundle's status condition array before checking whether the existing and
new conditions are semantically equal.

### DIFF
--- a/internal/updater/updater.go
+++ b/internal/updater/updater.go
@@ -68,8 +68,11 @@ func (u *Updater) Apply(ctx context.Context, b *olmv1alpha1.Bundle) error {
 func EnsureCondition(condition metav1.Condition) UpdateStatusFunc {
 	return func(status *olmv1alpha1.BundleStatus) bool {
 		existing := meta.FindStatusCondition(status.Conditions, condition.Type)
-		meta.SetStatusCondition(&status.Conditions, condition)
-		return existing == nil || !conditionsSemanticallyEqual(*existing, condition)
+		if existing == nil || !conditionsSemanticallyEqual(*existing, condition) {
+			meta.SetStatusCondition(&status.Conditions, condition)
+			return true
+		}
+		return false
 	}
 }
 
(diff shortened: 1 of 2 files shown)


provisioner/k8s/controllers/bundle_controller.go
```diff
@@ -132,7 +132,7 @@ func (r *BundleReconciler) handlePendingPod(u *updater.Updater, pod *corev1.Pod)
 		if cStatus.State.Waiting != nil && cStatus.State.Waiting.Reason == "ErrImagePull" {
 			messages = append(messages, cStatus.State.Waiting.Message)
 		}
-		if cStatus.State.Waiting != nil && cStatus.State.Waiting.Reason == "ImagePullBackoff" {
+		if cStatus.State.Waiting != nil && cStatus.State.Waiting.Reason == "ImagePullBackOff" {
```
timflannagan, line 135: Note: I didn't feel like creating a separate commit for this, but I also uncovered this when debugging locally.

internal/updater/updater.go
```diff
@@ -68,8 +68,11 @@ func (u *Updater) Apply(ctx context.Context, b *olmv1alpha1.Bundle) error {
 func EnsureCondition(condition metav1.Condition) UpdateStatusFunc {
 	return func(status *olmv1alpha1.BundleStatus) bool {
 		existing := meta.FindStatusCondition(status.Conditions, condition.Type)
-		meta.SetStatusCondition(&status.Conditions, condition)
-		return existing == nil || !conditionsSemanticallyEqual(*existing, condition)
```
timflannagan, line 72: Note: this logic is only problematic when we have an existing condition that matches the condition.Type (e.g. `Unpacked`), so we prematurely set the status conditions before checking whether the existing and new conditions are semantically equal.

joelanford (approved):
/lgtm

Good catches!

PR #80, opened 2022-03-01
Add basic k8s provisioner Bundle e2e tests

timflannagan:
Introduce a barebones Bundle e2e test to ensure we're covering the basic happy path for plain manifest Bundles.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -36,8 +36,18 @@ jobs:
         kind version
         kind create cluster
         kind export kubeconfig
-    - name: Run e2e tests
-      # TODO: Make install has been removed until the Kuberpak manifests are brought in
+    - name: Install the rukpak k8s provisioner
       run: |
-        # make install
+        make generate
+        kubectl create namespace rukpak-system
+        kubectl apply -f manifests
+
+        # Note: run the provisioner in the background so we don't block
+        # running ginkgo tests in the foreground.
+        # TODO: remove this hack when we have k8s provisioner manifests
+        # and more fleshed out CI pipeline that builds/pushes the unpack
+        # and provisioner image(s) so we can reference them at runtime.
+        nohup go run provisioner/k8s/main.go &
+    - name: Run e2e tests
+      run: |
         make test-e2e
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,7 @@ test-unit: ## Run the unit tests
 	$(Q)go test -count=1 -short $(UNIT_TEST_DIRS)
 
 test-e2e: ginkgo ## Run the e2e tests
-	$(GINKGO) run test/e2e
+	$(GINKGO) -v -trace -progress test/e2e
 
 verify: tidy generate ## Verify the current code generation and lint
 	git diff --exit-code
--- a/test/e2e/e2e_suite_test.go
+++ b/test/e2e/e2e_suite_test.go
@@ -2,12 +2,34 @@ package e2e_test
 
 import (
 	"testing"
+	"time"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	"github.com/operator-framework/rukpak/api/v1alpha1"
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	c client.Client
 )
 
 func TestE2E(t *testing.T) {
 	RegisterFailHandler(Fail)
+	SetDefaultEventuallyTimeout(1 * time.Minute)
+	SetDefaultEventuallyPollingInterval(1 * time.Second)
 	RunSpecs(t, "E2E Suite")
 }
+
+var _ = BeforeSuite(func() {
+	config := ctrl.GetConfigOrDie()
+
+	scheme := runtime.NewScheme()
+	err := v1alpha1.AddToScheme(scheme)
+	Expect(err).To(BeNil())
+
+	c, err = client.New(config, client.Options{Scheme: scheme})
+	Expect(err).To(BeNil())
+})
--- a/test/e2e/k8s_provisioner_test.go
+++ b/test/e2e/k8s_provisioner_test.go
@@ -1,0 +1,74 @@
+package e2e_test
+
+import (
+	"context"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"github.com/operator-framework/rukpak/api/v1alpha1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ = Describe("k8s provisioner", func() {
+	var (
+		bundle *v1alpha1.Bundle
+	)
+	BeforeEach(func() {
+		By("creating the testing Bundle resource")
+		bundle = &v1alpha1.Bundle{
+			ObjectMeta: v1.ObjectMeta{
+				Name: "olm-crds",
+			},
+			Spec: v1alpha1.BundleSpec{
+				ProvisionerClassName: "core.rukpak.io/plain",
+				Image:                "quay.io/tflannag/olm-plain-bundle:olm-crds-v0.20.0",
+			},
+		}
+		err := c.Create(context.Background(), bundle)
+		Expect(err).To(BeNil())
+	})
+	AfterEach(func() {
+		By("deleting the testing Bundle resource")
+		Expect(c.Delete(context.Background(), bundle)).To(BeNil())
+	})
+
+	When("a valid Bundle referencing a remote container image is created", func() {
+		It("should eventually report a successful state", func() {
+			By("eventually reporting an Unpacked phase", func() {
+				Eventually(func() bool {
+					if err := c.Get(context.Background(), client.ObjectKeyFromObject(bundle), bundle); err != nil {
+						return false
+					}
+					return bundle.Status.Phase == v1alpha1.PhaseUnpacked
+				}).Should(BeTrue())
+			})
+
+			By("eventually writing a non-empty image digest to the status", func() {
+				Eventually(func() bool {
+					if err := c.Get(context.Background(), client.ObjectKeyFromObject(bundle), bundle); err != nil {
+						return false
+					}
+					// Note(tflannag): This may make the test fairly brittle over time. Let's re-evaluate
+					// whether we want to test this kind of comparison if there's the potential for moving
+					// tags for Bundle container images.
+					expectedDigest := "quay.io/tflannag/olm-plain-bundle@sha256:2a72e4a7bc6c7598d1ecdb2f082c865a91dda35ab8e4a8e8bc128cf49a2a619b"
+					return bundle.Status.Digest == expectedDigest
+				}).Should(BeTrue())
+			})
+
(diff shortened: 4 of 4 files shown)


timflannagan:
I wasn't able to get `act` working with docker-based jobs so throwing this up now so I can get CI working while I continue to iterate.

Makefile
```diff
@@ -55,7 +55,7 @@ install: install-k8s ## Install all rukpak core CRDs and provisioners
 
 run-local: install-apis ## Install CRDs and run provisioner locally
 	kubectl create namespace rukpak-system
-	$(Q)go run provisioner/k8s/main.go
+	nohup $(Q)go run provisioner/k8s/main.go &
```
exdx, line 58: looked up `nohup`, very cool command!

test/e2e/e2e_suite_test.go
```diff
@@ -1,13 +1,101 @@
...
+				return bundle.Status.Phase == v1alpha1.PhaseUnpacked
+			}).Should(BeTrue())
+		})
+
+		It("should eventually contain a non-empty image digest", func() {
+			Eventually(func() bool {
+				if err := c.Get(context.Background(), client.ObjectKeyFromObject(bundle), bundle); err != nil {
+					return false
+				}
+				return bundle.Status.Digest != ""
```
exdx: nit: since the image under test is stable, should we test that the digest equals what we would expect versus simply non-empty? or is that overcoupling the image to the test?

test/e2e/e2e_suite_test.go
```diff
@@ -1,13 +1,105 @@
...
+
+			By("eventually the status contains a non-empty list of unpacked objects", func() {
+				Eventually(func() bool {
+					if err := c.Get(context.Background(), client.ObjectKeyFromObject(bundle), bundle); err != nil {
+						return false
+					}
+					if bundle.Status.Info == nil {
+						return false
+					}
+					return len(bundle.Status.Info.Objects) != 0
```
exdx, line 100: also here, since we know what the objects should be, should we compare them directly?

Makefile
timflannagan: Yep I think I'm going to move that nohup call to the actual workflow file as I'd imagine run-local should run the provisioner in the foreground by default. This is just a hack until we get a better idea of the provisioner manifests + a more concrete CI pipeline.

test/e2e/e2e_suite_test.go
timflannagan: Yep - I was trying to get the CI pipeline working first, but this sounds reasonable to me.

timflannagan: Actually this might make the test fairly brittle over time if there's a chance for a moving tag. I'm going to still add this, but I'll make a note that we may want to move away from this comparison over time.

test/e2e/e2e_suite_test.go
timflannagan: I was worried about test verbosity, and it feels like we're introducing more test brittleness, so I ended up just updating the conditional to ensure there's eight objects that will be present in the objects array. Does that sound reasonable to you?

exdx: SGTM

.github/workflows/test.yaml
```diff
@@ -36,8 +36,18 @@ jobs:
         kind version
         kind create cluster
         kind export kubeconfig
+    - name: Install the rukpak k8s provisioner
+      run: |
+        make generate
+        kubectl create namespace rukpak-system
```
tylerslaton, line 42: Should this get installed by running `k apply -f manifests` and we just create a `namespace.yaml` file in there?

test/e2e/e2e_suite_test.go
```diff
@@ -1,13 +1,109 @@
...
+
+	scheme := runtime.NewScheme()
+	err := v1alpha1.AddToScheme(scheme)
+	Expect(err).To(BeNil())
+
+	c, err = client.New(config, client.Options{Scheme: scheme})
+	Expect(err).To(BeNil())
+})
+
+var _ = Describe("k8s provisioner", func() {
```
joelanford, line 39: nit: should we put the k8s provisioner e2e in a separate file, specific to this provisioner?

test/e2e/e2e_suite_test.go
```diff
@@ -1,13 +1,109 @@
...
+	)
+	BeforeEach(func() {
+		By("creating the testing Bundle resource")
+		bundle = &v1alpha1.Bundle{
+			ObjectMeta: v1.ObjectMeta{
+				Name: "olm-crds",
+			},
+			Spec: v1alpha1.BundleSpec{
+				ProvisionerClassName: "core.rukpak.io/plain",
+				Image:                "quay.io/tflannag/olm-plain-bundle:olm-crds-v0.20.0",
```
joelanford, line 51: For follow-up: we should build a bundle image locally as part of the e2e setup, load it into kind, and then reference that image here.

test/e2e/e2e_suite_test.go
```diff
@@ -1,13 +1,109 @@
...
+			},
+		}
+		err := c.Create(context.Background(), bundle)
+		Expect(err).To(BeNil())
+	})
+	AfterEach(func() {
+		By("deleting the testing Bundle resource")
+		Eventually(func() error {
+			return c.Delete(context.Background(), bundle)
+		}).Should(Succeed())
```
joelanford, line 61: Curious why we need Eventually here. Do we expect `Delete` calls to fail temporarily?

test/e2e/e2e_suite_test.go
```diff
@@ -1,13 +1,109 @@
...
+			By("eventually reporting an Unpacked phase", func() {
+				Eventually(func() bool {
+					if err := c.Get(context.Background(), client.ObjectKeyFromObject(bundle), bundle); err != nil {
+						return false
+					}
+					return bundle.Status.Phase != v1alpha1.PhaseUnpacked
+				}).Should(BeTrue())
+			})
+
+			By("eventually the status contains a non-empty image digest", func() {
```
tylerslaton, line 83: super-nit: this flows a bit better
```suggestion
			By("eventually writing a non-empty image digest to the status", func() {
```

test/e2e/e2e_suite_test.go
```diff
@@ -1,13 +1,109 @@
...
+
+	When("a valid Bundle referencing a remote container image is created", func() {
+		It("should eventually report a successful state", func() {
+			By("eventually reporting a pending state")
+			Eventually(func() bool {
+				if err := c.Get(context.Background(), client.ObjectKeyFromObject(bundle), bundle); err != nil {
+					return false
+				}
+				return bundle.Status.Phase == v1alpha1.PhasePending
+			}).Should(BeTrue())
```
joelanford, line 72: This seems racy. We might not ever see the `Pending` phase if the unpack happens quickly enough.

test/e2e/e2e_suite_test.go
```diff
@@ -1,13 +1,109 @@
...
+					}
+					// Note(tflannag): This may make the test fairly brittle over time. Let's re-evaluate
+					// whether we want to test this kind of comparison if there's the potential for moving
+					// tags for Bundle container images.
+					expectedDigest := "quay.io/tflannag/olm-plain-bundle@sha256:2a72e4a7bc6c7598d1ecdb2f082c865a91dda35ab8e4a8e8bc128cf49a2a619b"
+					return bundle.Status.Digest == expectedDigest
+				}).Should(BeTrue())
+			})
+
+			By("eventually the status contains a non-empty list of unpacked objects", func() {
```
tylerslaton, line 96: similar to the super nit above ^

```suggestion
			By("eventually writing a non-empty list of unpacked objects to the status", func() {
```

tylerslaton (commented):
Looks good, small question and some nits.

.github/workflows/test.yaml
```diff
@@ -36,8 +36,18 @@ jobs:
...
+        kubectl apply -f manifests
+
+        # Note: run the provisioner in the background so we don't block
+        # running ginkgo tests in the foreground.
+        # TODO: remove this hack when we have k8s provisioner manifests
+        # and more fleshed out CI pipeline that builds/pushes the unpack
+        # and provisioner image(s) so we can reference them at runtime.
+        nohup go run provisioner/k8s/main.go &
```
joelanford, line 50: Nit/follow-up only: It would be great to get this into a make target such that:
1. e2e's are easily runnable locally
2. e2e setup and environment is fully encapsulated by the Makefile so that local runs are identical to CI runs (e.g. exact versions of kind, the cluster version, and envtest)

Ideal it would be:
- Locally I can run `make e2e`
- This workflow is just `checkout`, `setup-go`, `make e2e`

.github/workflows/test.yaml
timflannagan: I haven't give this a ton of thought yet, but if we introduce a root namespace manifest does that imply all provisioners all share the same system namespace? These steps mainly emulate the run-local Makefile target, so I was making it consistent with how it applied the provisioner.

joelanford: Tangential/Food for thought: do we want to have a distinction between `run-local` and running in CI? Seems like a more productive setup would entail CI jobs that use the same workflows as developers use locally.

tylerslaton: That brings into question how an end-user gets this onto their cluster and we're probably not there quite yet. Ultimately though we should iron out that UX to make it as simple as possible. Ideally a user wouldn't need to run "k create namespace rukpak-system" though it isn't a huge deal.

test/e2e/e2e_suite_test.go
joelanford: To avoid raciness, we likely need a more controlled test where we initialize some state in etcd, start the caches, but not the controllers, and then call `Reconcile()` manually to step through the transitions one `Reconcile()` call at a time.

That kind of setup is only necessary to catch intermediate states. It's fine to use this style to assert on the "final" expected state.

joelanford: Example of how I did that for the helm-operator-plugin's reconciler is here: https://github.com/operator-framework/helm-operator-plugins/blob/0d534ef8e09aed9e4bab7f084d17ab38731b353e/pkg/reconciler/reconciler_test.go#L386

.github/workflows/test.yaml
timflannagan: I think we still need to move all the OLM/registry kind provisioner stuff to an o-f/test-infra and use that as a shared library so we don't have to constantly re-invent the wheel here.

test/e2e/e2e_suite_test.go
timflannagan: I ended up removing this entirely as I think we're mainly interested in a valid Bundle reaching an unpacked state (and potentially ensuring it's not flapping to other phases once it reaches that state) right now, and we can build on top of this when we get some capacity.

timflannagan:
@tylerslaton @joelanford I can try and carve out some time to handle any follow-ups but is there anything else blocking? Just want to make sure we have some super basic coverage before I potentially context switch to other work tomorrow.

test/e2e/e2e_suite_test.go
timflannagan: Yep, it could be nice to also house the testing manifests, the Dockerfile, and also build/load dynamically at test runtime.

tylerslaton (approved):
Looks good to me, sending an approval with the assumption follow-up will be addressed later.

.github/workflows/test.yaml
timflannagan: @joelanford I took at stab at doing this in https://github.com/operator-framework/rukpak/pull/85.

PR #82, opened 2022-03-01
add k8s provisioner manifests; add local deployment target to Makefile

exdx:

This PR adds the manifests required to deploy the k8s/plain bundle provisioner locally. It also adds a new `run` target to the Makefile which builds the container image referenced in the deployment manifest, loads it onto the kind cluster, and then applies the required manifests to the cluster. The result is a running k8s provisioner in the cluster. 

One assumption made was that the k8s-provisioner should run in the `rukpak-system` namespace instead of its own private namespace. I think this is fair since the k8s provisioner is the only installable part of rukpak, apart from the CRDs, and it should be in the system namespace. It's fair to assume it will come default with all rukpak installations. 

Follow up: introduce local Dockerfile for much faster local build times.


timflannagan (commented):
Can we separate out the Makefile changes into their own commit?

Makefile
```diff
@@ -68,6 +82,11 @@ bin/k8s:
 bin/unpack:
 	CGO_ENABLED=0 go build -o $@ ./cmd/unpack/...
 
+build-local-container: ## Builds provisioner container image locally
+	docker build -f Dockerfile -t $(IMAGE) .
+
+load-image: ## Load-image loads the currently constructed image onto the cluster
+	 kind load docker-image $(IMAGE)
```
timflannagan, line 89: It would be nice if we outlined these steps in dev documentation at some point.

Makefile
```diff
@@ -57,6 +62,15 @@ run-local: install-apis ## Install CRDs and run provisioner locally
 	kubectl create namespace rukpak-system
 	$(Q)go run provisioner/k8s/main.go
 
+deploy: install-apis ## Deploy the operator to the current cluster
+	kubectl apply -f provisioner/k8s/manifests
+
+run: build-local-container load-image deploy
```
timflannagan, line 68: I think this is reasonable while we iterate but we may need to segment this to be provisioner-specific, or rethink how we install rukpak provisioners long term.

Makefile
tylerslaton: Combo has an outline for this we could just template from.

tylerslaton (approved):
Nice! It's all coming together.

Makefile
```diff
@@ -7,6 +7,11 @@ help: ## Show this help screen
 	@echo ''
 	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z0-9_-]+:.*?##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
 
+# Container build options
+IMAGE_REPO=quay.io/operator-framework/k8s-provisioner
```
timflannagan, line 11: Note: we need to create this quay repository still.

Makefile
```diff
@@ -57,11 +62,17 @@ run-local: install-apis ## Install CRDs and run provisioner locally
...
 	$(Q)go run provisioner/k8s/main.go
 
+deploy: install-apis ## Deploy the operator to the current cluster
+	kubectl apply -f provisioner/k8s/manifests
+
+run: build-local-container load-image deploy ## Build image and run operator in-cluster
+
 ## --------------------------------------
-## Building
+## Build and Load
```
timflannagan, line 71: We're missing the `.PHONY: build bin/k8s bin/unpack` that was lost during the rebase.

timflannagan: Whoops - had a stale diff. Disregard.

Makefile
exdx: That sounds good

> Can we separate out the Makefile changes into their own commit?

Done

Makefile
exdx: I have went ahead and created it.

> what happens when we have a second image?
That's a good point, I think our Makefile/repository structure will have to change

PR #83, opened 2022-03-02
.github/workflow: Split up the test workflow into individual jobs

timflannagan:
Update the test.yaml workflow and split the unit and e2e-kind jobs into
their own workflows. This enables the ability to retest them
independently when we encounter a workflow failure as GHA doesn't
support that use case right now.


tylerslaton (approved):
Yeah I like this much better.

PR #85, opened 2022-03-02
Add an e2e-kind target for deploying rukpak to kind clusters

timflannagan:
Follow-up to #80 which added basic Bundle e2e tests.

This adds a more streamlined e2e experience by ensuring both CI and local environments use the same workflow through a single Makefile target. Users that want an e2e cluster stood up for them can run `make e2e-kind`, which will create a kind cluster, build the provisioner container image, load that container image into kind, and deploy the provisioner stack to the kind cluster. In the case you have an existing cluster running, or you don't want to use kind, you can run `make test-e2e` to run the e2e suite locally.

Remove the build.yaml workflow now that the e2e.yaml workflow builds both the binaries and container images under-the-hood using the `make e2e-kind` target.


provisioner/k8s/manifests/01_cluster_role.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
joelanford, line 1: Is all the rbac generated with `controller-gen` from[ `+kubebuilder` markers in the controllers](https://github.com/operator-framework/rukpak/blob/79f1e29557df669bb12dc5bef9d222f3c14c7c55/provisioner/k8s/controllers/bundle_controller.go#L66-L70)? If not, did we make a conscious decision to not use that?

joelanford: Also, doesn't the k8s provisioner need `*`/`*`/`*` permissions in order to be able to deploy arbitrary manifests?

Makefile
```diff
@@ -57,17 +65,37 @@ run-local: install-apis ## Install CRDs and run provisioner locally
...
+build-local-container: ## Builds provisioner container image locally
+	docker build -f Dockerfile -t $(IMAGE) .
+
+kind-load: ## Load-image loads the currently constructed image onto the cluster
+	 ${KIND} load docker-image $(IMAGE)
+
+kind-cluster: ## Standup a kind cluster for e2e testing usage
+	${KIND} get clusters | grep ${KIND_CLUSTER_NAME} || ${KIND} create cluster --name ${KIND_CLUSTER_NAME}
+
+e2e-kind: build-local-container kind-cluster kind-load deploy test-e2e ## Run e2e tests against a kind cluster
```
joelanford, line 98: WDYT about being more opinionated and setting this up such that `test-e2e` _always_ uses kind.

IMO, hermtically sealing as much as possible is and leaving fewer choices to developers means more guarantees of consistency between local and CI test runs.

Makefile
```diff
@@ -57,17 +65,37 @@ run-local: install-apis ## Install CRDs and run provisioner locally
...
 	CGO_ENABLED=0 go build -o $@ ./cmd/unpack/...
 
+build-local-container: ## Builds provisioner container image locally
+	docker build -f Dockerfile -t $(IMAGE) .
+
+kind-load: ## Load-image loads the currently constructed image onto the cluster
+	 ${KIND} load docker-image $(IMAGE)
+
+kind-cluster: ## Standup a kind cluster for e2e testing usage
+	${KIND} get clusters | grep ${KIND_CLUSTER_NAME} || ${KIND} create cluster --name ${KIND_CLUSTER_NAME}
```
joelanford, line 96: Once upon a time (and maybe this is still the case), I had the operator-sdk e2e setup to automatically teardown the e2e cluster as part of the run to ensure the cluster didn't retain state between runs.

I'm fuzzy on the details, but I think I had it setup to tear it down after successful completion so that if the e2e failed, the cluster would still be there for inspection.

provisioner/k8s/manifests/01_cluster_role.yaml
exdx: This is what I was thinking about when writing the RBAC. I wanted to avoid `"*"` on everything at the very beginning, since the actual set of permissions required to deploy an operator via a plain k8s bundle is limited to the realistic scope of the manifests in the bundle: things in the core API group, RBAC, CRDs, and maybe some additional basic types. 

If it's possible, we should give it the smallest set of permissions at the beginning, and then expand. If it's possible to avoid https://github.com/operator-framework/operator-lifecycle-manager/issues/1685 up front we should.

exdx: > Is all the rbac generated with `controller-gen` from[ `+kubebuilder` markers in the controllers](https://github.com/operator-framework/rukpak/blob/79f1e29557df669bb12dc5bef9d222f3c14c7c55/provisioner/k8s/controllers/bundle_controller.go#L66-L70)? If not, did we make a conscious decision to not use that?

We should use those if possible, right now the RBAC is done manually -- it looks like the markers weren't updated with the new `core.rukpak.io` label though, which is interesting

Makefile
```diff
@@ -86,8 +89,13 @@ bin/unpack:
 build-local-container: ## Builds provisioner container image locally
 	docker build -f Dockerfile -t $(IMAGE) .
 
-load-image: ## Load-image loads the currently constructed image onto the cluster
-	 kind load docker-image $(IMAGE)
+kind-load: ## Load-image loads the currently constructed image onto the cluster
```
exdx, line 92: looks like if we have a custom kind cluster name then this command needs to accept the name as a flag to load successfully, ie` kind load docker-image $(IMAGE) --name $(KIND_CLUSTER_NAME)` otherwise we see https://github.com/operator-framework/rukpak/runs/5386209881?check_suite_focus=true

Makefile
```diff
@@ -57,17 +65,37 @@ run-local: install-apis ## Install CRDs and run provisioner locally
...
+GO_BUILD := $(Q)go build
+build: bin/k8s bin/unpack
 
 bin/k8s:
 	CGO_ENABLED=0 go build -o $@ ./provisioner/k8s
 
 bin/unpack:
 	CGO_ENABLED=0 go build -o $@ ./cmd/unpack/...
 
+build-local-container: ## Builds provisioner container image locally
```
exdx, line 89: is there any interest in having a local.Dockerfile which just takes the built binaries from the local machine and puts them in a container instead, like combo has? The build time for the container is pretty slow locally, wondering if it makes sense to avoid that, or if we only want to test one specific way as @joelanford is suggesting in https://github.com/operator-framework/rukpak/pull/85/files#r817721452.

Makefile
timflannagan: Yeah I don't know where I land with this most of the time. I think the idea of hermetically sealing clusters makes a lot of sense, but the potential churn of create a cluster, waiting for that cluster to be ready, installing the rukpak stack, running e2e tests, and then deleting the cluster isn't the greatest experience when you're developing e2e tests and just want quick feedback from the cluster.

I think this is still worth pursuing, but maybe it's fine in the short term to have separate Makefile targets that enable both workflows while we iterate? We could track hermetically sealing cluster runs (or even make the OLM kind provisioner a library we import here) as an issue.

Makefile
```diff
@@ -12,6 +12,9 @@ IMAGE_REPO=quay.io/operator-framework/k8s-provisioner
 IMAGE_TAG=latest
 IMAGE=$(IMAGE_REPO):$(IMAGE_TAG)
 
+KIND_CLUSTER_NAME := olm-e2e
```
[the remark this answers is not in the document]
timflannagan: Oh yep - good catch.

Makefile
timflannagan: That would be my preference but I thought we ran into some opposition when we were talking about that last week? Maybe it's fine for now and we can track the want/need for a dev.Dockerfile?

Makefile
joelanford: Yeah, my thought would be something like:
```
# Set up the environment with a new kind cluster
# Run each of the actual e2e tests
# Tear down the kind cluster
# Restore the original environment
make e2e  

# Just run the e2e against the current environment
make e2e-provisioner-k8s 
```

I guess that's really close to what you've got, and maybe all I'm suggesting is this, such that a "default"/"obvious" target is the hermetically sealed one, but sub-targets exist if you want to manage setup/teardown yourself:
```suggestion
e2e: build-local-container kind-cluster kind-load deploy test-e2e ## Run e2e tests against a kind cluster
```

Makefile
joelanford: Can we have the _real_ build:
1. Enforce a specific version of Go
2. Ensure the locally built binary is up-to-date, including all the necessary build flags that we would want in our production image?
3. Copy that binary into the image.

Makefile
timflannagan: I took in your suggested change in the last force push. Maybe we can experiment with that `kind-cluster` target and switch the conditional to check whether there's an _existing_ cluster with that special e2e kind cluster name, and if so, delete that cluster and create a new one? 

Existing:

```make
kind-cluster: ## Standup a kind cluster for e2e testing usage
	${KIND} get clusters | grep ${KIND_CLUSTER_NAME} || ${KIND} create cluster --name ${KIND_CLUSTER_NAME}
```

And then we could update that conditional to something like the following:

```make
kind-cluster: ## Standup a kind cluster for e2e testing usage
ifeq (1, $(shell kind get clusters | grep ${KIND_CLUSTER_NAME} | wc -l))
	${KIND} delete cluster --name ${KIND_CLUSTER_NAME}
endif
        ${KIND} create cluster --name ${KIND_CLUSTER_NAME}
```

provisioner/k8s/main.go
```diff
@@ -60,7 +60,7 @@ func main() {
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.StringVar(&systemNamespace, "system-namespace", "rukpak-system", "Configures the namespace that gets used to deploy system resources.")
-	flag.StringVar(&unpackImage, "unpack-image", "quay.io/tflannag/unpack:latest", "Configures the container image that gets used to unpack Bundle contents.")
+	flag.StringVar(&unpackImage, "unpack-image", "quay.io/operator-framework/k8s-provisioner:latest", "Configures the container image that gets used to unpack Bundle contents.")
```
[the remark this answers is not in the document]
timflannagan: Talked about this in slack - we're leaning towards removing the run-local target, as it's not guaranteed to work using the default unpack-image container image as that target doesn't exist right now. Users in this context would now be required to explicitly override this flag's value, or build it locally and load it into a kind cluster.

exdx (approved):

Makefile
timflannagan: I tackled partially implementing this in https://github.com/operator-framework/rukpak/pull/90. Should leave a more consistent cluster cluster state between local and CI runs.

PR #88, opened 2022-03-02
e2e: Ensure the underlying Bundle unpack Pod is re-created when deleted

timflannagan:
The k8s provisioner currently maintains a Bundle unpack Pod, and parses the Pod's logs in order to read unpacked Bundle contents so it can store it in a series of ConfigMaps. We should ensure we a simple e2e test to ensure this behavior so we're not regressing silently.

### DIFF
--- a/test/e2e/k8s_provisioner_test.go
+++ b/test/e2e/k8s_provisioner_test.go
@@ -11,8 +11,15 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	// TODO: make this is a CLI flag?
+	defaultSystemNamespace = "rukpak-system"
 )
 
 func Logf(f string, v ...interface{}) {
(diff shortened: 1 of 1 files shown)


timflannagan:
Wait until #85 lands.

/hold

test/e2e/k8s_provisioner_test.go
```diff
@@ -70,5 +79,47 @@ var _ = Describe("k8s provisioner", func() {
 				}).Should(BeTrue())
 			})
 		})
+
+		It("should re-create underlying system resources", func() {
```
timflannagan, line 83: Note: we can extend this logic to also check that the object and metadata ConfigMaps for this Bundle object get re-created. This can be done as a follow-up?

test/e2e/k8s_provisioner_test.go
```diff
@@ -70,5 +79,47 @@ var _ = Describe("k8s provisioner", func() {
...
+					LabelSelector: labels.NewSelector().Add(*requirement),
+					Namespace:     "rukpak-system",
+				}); err != nil {
+					return false
+				}
+				if len(pods.Items) != 1 {
+					return false
+				}
+				podNN = client.ObjectKeyFromObject(&pods.Items[0])
+				return true
```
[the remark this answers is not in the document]
timflannagan: Yeah honestly this was my first iteration. I'm not entirely sure why I settled on using a podNN vs. a pod variable but it was also 3am in the morning, so it's possible my brain wasn't working at 100% capacity. It seems like this would allow us to eliminate a single call, so it seems worthwhile to change.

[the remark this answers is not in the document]
timflannagan: Updated - note: I had to wrap the delete call in an eventually handler as to avoid 1/3 runs being flaky. This seems like a reasonable workaround to me?

test/e2e/k8s_provisioner_test.go
timflannagan: Added a TODO comment so we don't lose track. It would be nice to differentiate between FIXME's and TODO's, where the latter has a github action that automatically creates follow-up tickets, and the former is a considered a blocker for merging the PR and mainly used in the WIP/draft PR contest. cc @joelanford in case this logic makes sense to you?

test/e2e/k8s_provisioner_test.go
```diff
@@ -70,5 +84,46 @@ var _ = Describe("k8s provisioner", func() {
...
+				}
+				pod = &pods.Items[0]
+				return true
+			}).Should(BeTrue())
+
+			By("storing the pod's original UID")
+			originalUID := pod.GetUID()
+
+			By("deleting the underlying pod and waiting for it to be re-created")
+			Eventually(func() error {
```
joelanford, line 118: Nit: do we need `Eventually` here?

joelanford: Ah, just saw your other comment. That's... interesting. Any clue why its flaky without the eventually?

timflannagan: I'm not entirely sure - this is what I had logged in a markdown file when I was debugging yesterday:

```
    Expected
        <*errors.StatusError | 0xc0003c0aa0>: {
            ErrStatus: {
                TypeMeta: {Kind: "", APIVersion: ""},
                ListMeta: {
                    SelfLink: "",
                    ResourceVersion: "",
                    Continue: "",
                    RemainingItemCount: nil,
                },
                Status: "Failure",
                Message: "pods \"plain-unpack-bundle-olm-crds-valid\" not found",
                Reason: "NotFound",
                Details: {
                    Name: "plain-unpack-bundle-olm-crds-valid",
                    Group: "",
                    Kind: "pods",
                    UID: "",
                    Causes: nil,
                    RetryAfterSeconds: 0,
                },
                Code: 404,
            },
        }
    to be nil
```

My initial thought was an object cache issue but we create the c-r client.Client using client.New so we're likely directly reading/writing to the apiserver throughout the e2e suite. Maybe the `CreateOrRecreate(...)` function is deleting the Pod when it detects we need to make changes to the underlying bundle unpack pod?

I realize I could also just wrap the Delete call with `client.IgnoreNotFound(c.Delete(ctx, pod))` to burn down the flakes, which seems fine to me, but I'm not sure if there's the possibility we sweep legitimate controller issues under the rug over time.

exdx (approved):

test/e2e/k8s_provisioner_test.go
joelanford: Somewhat mystifying that it exists for the `List` but not the `Get`. If it was a caching issue, `Get` would always have it if a previous `List` did (assuming it didn't get deleted in-between).

`CreateOrRecreate()` is an interesting thought.

> there's the possibility we sweep legitimate controller issues under the rug over time.
I think this is deserving of a little more attention for this reason.

timflannagan: Added a TODO comment for re-evaluating this in #105. Any existing TODO comments should be created as issues in #106.

joelanford: I think I figured it out: https://github.com/operator-framework/rukpak/pull/107

PR #96, opened 2022-03-02
e2e: Surface invalid Bundle container images back to the Bundle status

timflannagan:
Add a simple Bundle e2e test that contains an invalid spec.Image container image that doesn't exist, and ensure the Bundle reports back that underlying Pod's status back to the set of Bundle status conditions. This should help avoid regressing on the fixes that went into #77, where we didn't have an e2e structure in place at the time of that PRs' creation.


timflannagan:
This should be helpful in having to avoid manually uncovering and verifying changes like we did in #77, and weed out any regressions in the behavior over time.

PR #100, opened 2022-03-03
rename k8s provisioner to plain-v0

### LINKED ISSUE #87: Rename the k8s provisioner to something more appropriate
Goal: rename the k8s provisioner to a more descriptive name.

Currently this provisioner operates on the plain Kubernetes manifest bundle format (which is still WIP) but the naming of this provisioner is still ambiguous without further context. We should try and iron out a better name before cutting an alpha release.

exdx:
Closes #87 

Renames the k8s provisioner to plain-v0 based on the new name for the plain bundle format. Updates binary names, provisioner ID, and all references to the k8s provisioner. A quay.io/operator-framework/plain-v0-provisioner repository was created and the previous k8s repository was deleted. Incoming documentation will reflect the new provisioner name. 

Note: the provisioner name is plain-v0, named after the plain bundle format which is plain+v0.


Dockerfile
```diff
@@ -18,7 +18,7 @@ RUN make build
 FROM gcr.io/distroless/static:debug
 
 WORKDIR /
-COPY --from=builder /workspace/bin/k8s .
+COPY --from=builder /workspace/bin/plainv0 .
```
joelanford, line 21: I know this is probably annoying, but wdyt about `plain-v0` instead, just to make the name/version distinction more obvious?

exdx: That sounds reasonable. We can still use the name as an import path and everywhere else, so I think it's a good change. Will update.

provisioner/plain-v0/main.go
```diff
@@ -36,7 +36,7 @@ import (
 	olmv1alpha1 "github.com/operator-framework/rukpak/api/v1alpha1"
 	"github.com/operator-framework/rukpak/internal/storage"
 	"github.com/operator-framework/rukpak/internal/util"
-	"github.com/operator-framework/rukpak/provisioner/k8s/controllers"
+	"github.com/operator-framework/rukpak/provisioner/plain-v0/controllers"
```
joelanford, line 39: Very much a nit, and probably debatable anyway, so not for this PR:

We could consider making this `github.com/operator-framework/rukpak/provisioner/plain/v0` too.

joelanford (approved):
/approve

provisioner/plain-v0/main.go
njhale: are we planning on having a separate controller for every single version? I imagine the go module versions will be separate from the bundle type version and there will be some compatibility published.

PR #101, opened 2022-03-03
fix: kind-cluster target was failing to remove existing cluster 

exdx:
Small PR that fixes the logic in the Makefile `make kind-cluster` target that deletes existing clusters before creating a new one. The existing cluster was not being removed because the whitespace returned from `wc -l` was causing the equality check to fail. Piping the arguments to `xarg` removes any whitespace and causes the equality to return the right result.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -88,7 +88,7 @@ kind-load: ## Load-image loads the currently constructed image onto the cluster
 	${KIND} load docker-image $(IMAGE) --name $(KIND_CLUSTER_NAME)
 
 kind-cluster: ## Standup a kind cluster for e2e testing usage
-ifeq (1, $(shell kind get clusters | grep ${KIND_CLUSTER_NAME} | wc -l))
+ifeq (1, $(shell ${KIND} get clusters | grep ${KIND_CLUSTER_NAME} | wc -l | xargs))
 	@echo "Deleting the existing ${KIND_CLUSTER_NAME} test cluster"
 	${KIND} delete cluster --name ${KIND_CLUSTER_NAME}
 endif


Makefile
```diff
@@ -88,7 +88,7 @@ kind-load: ## Load-image loads the currently constructed image onto the cluster
 	${KIND} load docker-image $(IMAGE) --name $(KIND_CLUSTER_NAME)
 
 kind-cluster: ## Standup a kind cluster for e2e testing usage
-ifeq (1, $(shell kind get clusters | grep ${KIND_CLUSTER_NAME} | wc -l))
+ifeq (1, $(shell kind get clusters | grep ${KIND_CLUSTER_NAME} | wc -l | xargs))
```
timflannagan, line 91: What's the actual problem here? That `wc -l` is producing a newline on some shells? We could also just use `tr` to trim any whitespace from the grep buffer.

Side note: I realize I should've swapped `shell kind get clusters` with `shell ${KIND} get clusters`.

timflannagan:
(had a second to bang out some PR reviews - I'm going back to disappearing on PTO now :laughing:)

Makefile
exdx: The issue I saw locally was that `wc -l` was returning with extra whitespace at the front in my zsh shell:
```
rukpak(update-provisioner-name): kind get clusters | grep rukpak-e2e | wc -l
       1
```
which caused the comparison to not be accurate. There are a lot of ways to trim whitespace, I just went with the one I'm most familiar with. 

I can update to use the target to use the $KIND variable

Makefile
```diff
@@ -88,7 +88,7 @@ kind-load: ## Load-image loads the currently constructed image onto the cluster
 	${KIND} load docker-image $(IMAGE) --name $(KIND_CLUSTER_NAME)
 
 kind-cluster: ## Standup a kind cluster for e2e testing usage
-ifeq (1, $(shell kind get clusters | grep ${KIND_CLUSTER_NAME} | wc -l))
+ifeq (1, $(shell ${KIND} get clusters | grep ${KIND_CLUSTER_NAME} | wc -l | xargs))
```
timflannagan, line 91: A quick github search relayed this as a workaround too:

```makefile
ifeq (1, $(shell kind get clusters | grep ${KIND_CLUSTER_NAME} | wc -l | tr -d '[:space:]'))
```

timflannagan:
I tried this out locally and it works on my environment too. I ran into this issue when running the e2e target, but I'd imagine this was a problem before these changes:

```bash
$ make e2e
$ docker build -f Dockerfile -t quay.io/operator-framework/k8s-provisioner:latest .
...
Step 11/18 : RUN make build
 ---> Running in f4f13037b112
/bin/sh: 1: kind: not found
CGO_ENABLED=0 go build -o bin/k8s ./provisioner/k8s
CGO_ENABLED=0 go build -o bin/unpack ./cmd/unpack/...
```

timflannagan (approved):

exdx:
I see `/bin/sh: 1: kind: not found` floating around in the `make build-container` logs as well, not sure where it's coming from but we can fix that up as well.

PR #102, opened 2022-03-03
build executables locally and package them in an image

### LINKED ISSUE #98: Use local dev Dockerfile for faster development locally
Right now the local development process requires building the Dockerfile at the root of the repository and then loading it onto the kind cluster. Due to the large number of dependencies in building an operator building this image takes a large amount of time. The end result of the built image is two statically built binaries (k8s and unpack) with the `gcr.io/distroless/static:debug` image as a base image. 

It would be ideal if in local development environments a local Dockerfile can be used instead, which would simply copy the  binaries built locally from `bin` onto the base image. See https://github.com/operator-framework/combo/blob/main/Dockerfile.local for an example of a local Dockerfile. The relevant make target would build the binaries locally and then add them to the development Dockerfile. This would speed up local development significantly.

akihikokuroda:
Signed-off-by: akihikokuroda <akihikokuroda2020@gmail.com>

Closes #98

Add the docker file for the container build and update Makefile to prebuild the executables before container build.

### DIFF
--- a/Dockerfile.local
+++ b/Dockerfile.local
@@ -1,0 +1,9 @@
+FROM gcr.io/distroless/static:debug
+
+WORKDIR /
+COPY bin/plain-v0 .
+COPY bin/unpack .
+EXPOSE 8080
+
+ENTRYPOINT ["/plain-v0"]
+CMD ["run"]
--- a/Makefile
+++ b/Makefile
@@ -81,8 +81,11 @@ bin/plain-v0:
 bin/unpack:
 	CGO_ENABLED=0 go build -o $@ ./cmd/unpack/...
 
-build-local-container: ## Builds provisioner container image locally
+build-container: ## Builds provisioner container image locally
 	docker build -f Dockerfile -t $(IMAGE) .
+
+build-local-container: build ## Builds the provisioner container image using locally built binaries
+	docker build -f Dockerfile.local -t $(IMAGE) .
 
 kind-load: ## Load-image loads the currently constructed image onto the cluster
 	${KIND} load docker-image $(IMAGE) --name $(KIND_CLUSTER_NAME)


Makefile
```diff
@@ -84,6 +84,9 @@ bin/unpack:
 build-local-container: ## Builds provisioner container image locally
 	docker build -f Dockerfile -t $(IMAGE) .
 
+build-container: build
```
timflannagan, line 87: Should we swap the build-container and the build-local-container target names, i.e. the target that's responsible for building the Dockerfile.local dockerfile is named build-local-container?

timflannagan, line 87: It would be nice to hook this target into the kind-load target to help enable better dev workflows:

- build the local dockerfile
- load that local image into kind
- respin the provisioner deployment pod manually?

Maybe this is where the (now removed in #79) `run-local` target could come back into play again, and that target builds the local image and loads it into kind?

akihikokuroda: It's nice.  I'll look at the dev workflow.

Dockerfile.local
```diff
@@ -0,0 +1,9 @@
+FROM gcr.io/distroless/static:debug
+
+WORKDIR /
+COPY bin/k8s .
```
exdx, line 4: nit: we've decided to rename the provisioner to `registry-v1` in #100 so depending which lands first this may have to be rebased to point to the new binary.

Makefile
```diff
@@ -81,9 +81,12 @@ bin/k8s:
 bin/unpack:
 	CGO_ENABLED=0 go build -o $@ ./cmd/unpack/...
 
-build-local-container: ## Builds provisioner container image locally
+build-container: ## Builds provisioner container image locally
 	docker build -f Dockerfile -t $(IMAGE) .
 
+build-local-container: build
```
exdx, line 87: could we add a Make comment here to explain the target? Something simple like
`## Builds the provisioner container image using locally built binaries` sounds fine.

Dockerfile.local
exdx, line 4: one additional thought I've had around this -- do we want to ensure somehow that the binaries being copied into the image are not stale? maybe `build-container` should run `make build` first?

exdx:
Thank you for taking this one @akihikokuroda! It will greatly improve our local development experience.

Dockerfile.local
```diff
@@ -0,0 +1,9 @@
...
+COPY bin/unpack .
+EXPOSE 8080
+
+ENTRYPOINT ["/k8s"]
+CMD ["run"]
```
joelanford, line 9: Do we need a separate Dockerfile? It seems like that could lead to local development behaviors diverging from CI/release image behaviors.

What if we change the existing Dockerfile to copy in a local binary and update the make targets that build it to:
1. Enforce a specific go version
2. Ensure the underlying binary is rebuilt before running `docker build`

Dockerfile.local
akihikokuroda: I see.  I'll change to plain-v0.  Thanks!

akihikokuroda: Yes, `build-container` (I changed to `build-local-container`) target is depend on `build` target

Dockerfile.local
akihikokuroda: The existing Dockerfile is a multi-phase docker file.  It setup the binary build environment, build binaries and creates the container image.  It doesn't depend on the environment of the system.  It can build the consistent image.  The Dockerfile.local has just container built part.  The executable builds are depend of the local system.  It must be setup correctly but it builds the binaries faster than the building in the docker build process.

Makefile
akihikokuroda: I'll add it.  Thanks.

Dockerfile.local
joelanford: Yeah, I'm asserting that the enforcement of a specific Go version is really the only system environment thing that matters. 

I'm making that assertion based on the fact that the popular GoReleaser project builds images by building binaries locally and then copying them into an image: https://goreleaser.com/customization/docker/

joelanford: Interested in @timflannagan and @exdx thoughts on this idea.

timflannagan: Honestly I think we just punt on consolidating any CI vs. production Dockerfile situation right now, and enable the ability to build binaries locally, copy those binaries into a dockerfile, and load that image to a kind cluster to help enable dev workflows but maybe I'm biased? The only consideration that comes into mind is whether this kind of setup plays well with loading images to kind node(s) if docker/podman/etc. aren't able to detect new digest images, and if that's the case, then this kind of setup is absolutely watered down being able to quickly iterate without pushing these build dev container images to a remote container registry.

joelanford: > enable the ability to build binaries locally, copy those binaries into a dockerfile

Big +1 from me. I'm saying that's all we need. Let's use that same setup in CI.

> docker/podman/etc. aren't able to detect new digest images

This flow should work. Though I think we would need to `uninstall` and then `install` to cycle the deployment.

akihikokuroda: Both ways have pros and cons.  I think that providing make targets and docker files for both way is good at this stage.

exdx: I'm also ok with having two separate Dockerfiles at this stage but if @joelanford feels strongly we should consolidate then we could update this PR to remove the existing Dockerfile as well. 

My impression is that this PR was meant to be purely additive though and just provide a local Dockerfile. We could reassess our deployment strategy in a follow-up.

exdx: > Yeah, I'm asserting that the enforcement of a specific Go version is really the only system environment thing that matters.

I'm wondering how we could enforce that requirement - just have a script that runs `go version` before building and checks that the output is what we expect?

joelanford: My only thought is that one make target + one Dockerfile is easier to maintain and reason about than two, but I don't have a strong enough feeling to block progress.

>  runs go version before building and checks that the output is what we expect

Yep, pretty much.

exdx (approved):

Dockerfile.local
timflannagan: I just took another look at this lgtm. It sounds like there's reservations around potential skew between CI/local Dockerfiles, and we may want to consolidate this over time (which Joe has already pointed put here). Can we open an issue to track this concern so we don't lose this context over time, and push this through? @akihikokuroda would you mind creating this issue, and then we can merge this?

akihikokuroda:
I opened the issue: https://github.com/operator-framework/rukpak/issues/112

timflannagan (approved):

Issue #99, opened 2022-03-02
Support logical deletion of objects when deleting BundleInstances


exdx:
The order in which objects composing operators or other applications on-cluster are removed matters for the successful removal of the application from the cluster. The default behavior is to mark all objects for deletion and remove them when the parent resource, in this case the BundleInstance, is deleted. However a more fine grained removal policy may be desirable, which may be achievable entirely by finalizers or potentially by other resources. 

See https://github.com/operator-framework/operator-lifecycle-manager/issues/2235 for an example of an issue that can arise when deleting objects all at once: if the roles are removed before the deployment the operator loses the ability to cleanup after itself. 

The goal is to support an ordered removal of objects related to the installation such that the cluster can be in the same state as it was prior to the installation when the operator is uninstalled.

exdx:
Moving to 0.3 milestone -- this is something we definitely want to support sooner rather than later, to ensure a clean teardown of operators and other applications.

exdx:
Moving to a design ticket to architect this feature out.

exdx:
Move to backlog for now.

PR #103, opened 2022-03-04
e2e: Introduce basic BundleInstance tests

timflannagan:
Introduces barebones BundleInstance e2e tests.


timflannagan:
(just dumping what I've had sitting locally since earlier this week so we're not duplicating work while I'm on PTO)

test/e2e/plain-v0_provisioner_test.go
```diff
@@ -192,3 +194,188 @@ var _ = Describe("plain-v0 provisioner", func() {
...
+	When("a BundleInstance targets an invalid Bundle", func() {
+		var (
+			bundle *v1alpha1.Bundle
+			bi     *v1alpha1.BundleInstance
+		)
+		BeforeEach(func() {
+			By("creating the testing Bundle resource")
+			bundle = &v1alpha1.Bundle{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "olm-apis",
```
exdx, line 300: nit: should we be using `GenerateName()` in these tests to be consistent and avoid flakes per #107?

test/e2e/plain-v0_provisioner_test.go
```diff
@@ -192,3 +194,188 @@ var _ = Describe("plain-v0 provisioner", func() {
...
+					return false
+				}
+				if bi.Status.InstalledBundleName != "" {
+					return false
+				}
+				existing := meta.FindStatusCondition(bi.Status.Conditions, "Installed")
+				if existing == nil {
+					return false
+				}
+				expected := metav1.Condition{
```
exdx, line 280: In this test case, when the bundle was installed, BI was created, and then the bundle was removed, we show this status condition. But the BI is on cluster right, so the objects in the bundle are still on-cluster as well. The underlying storage of those objects was removed. This condition may be confusing because "`Installed: False`" suggests that the application is not installed, but it still is?

test/e2e/plain-v0_provisioner_test.go
```diff
@@ -192,3 +194,188 @@ var _ = Describe("plain-v0 provisioner", func() {
...
+				}
+				if bi.Status.InstalledBundleName != "" {
+					return false
+				}
+				existing := meta.FindStatusCondition(bi.Status.Conditions, "Installed")
+				if existing == nil {
+					return false
+				}
+				// TODO(tflannag): This kind of comparison feels brittle?
+				expectedMessage := `
```
exdx, line 347: this is the error we get when attempting to install a CR for a CRD that is not on the cluster? Maybe we should have custom error handling for this type of error as it seems pretty common

test/e2e/plain-v0_provisioner_test.go
timflannagan: Yep - needed to rebase this after those changes landed. Good catch.

test/e2e/plain-v0_provisioner_test.go
timflannagan: There's a bunch of potential cleanup needed in the BundleInstance controller around hardcoding condition types/reasons/phases/etc. so I didn't want to float this PR with making those kind of changes all at once. I've largely agree that this kind of installation failure should warrant it's own error type though. Maybe I should just add a TODO for now and we can revisit next week?

test/e2e/plain-v0_provisioner_test.go
```diff
@@ -203,7 +205,7 @@ var _ = Describe("plain-v0 provisioner", func() {
 			Eventually(func() bool {
 				pod := &corev1.Pod{}
 				if err := c.Get(ctx, types.NamespacedName{
-					Name:      fmt.Sprintf("plain-unpack-bundle-%s", bundle.GetName()),
+					Name:      util.PodName("plain", bundle.GetName()),
```
timflannagan, line 208: Note: it looks like this was potentially missed in the plain/k8s -> plain-v0 renaming PR as we hardcode the provisioners name in a couple of places. Right now the plain/v0 provisioner creates unpack Pods that are prefixed with `plain-unpack-bundle-*` vs. `plain-v0-unpack-bundle-*`.

test/e2e/plain-v0_provisioner_test.go
```diff
@@ -240,3 +242,183 @@ var _ = Describe("plain-v0 provisioner", func() {
...
+				}
+				existing := meta.FindStatusCondition(bi.Status.Conditions, "Installed")
+				if existing == nil {
+					return false
+				}
+				expected := metav1.Condition{
+					Type:    "Installed",
+					Status:  metav1.ConditionStatus(corev1.ConditionTrue),
+					Reason:  "InstallationSucceeded",
+					Message: "",
```
timflannagan, line 310: Note: there's likely a follow-up needed next week around cleaning up the BundleInstance controller and moving any hardcoded reasons/types/phases/etc. to the bundleinstance_types.go file, and using constants instead so it reads better, and it's easier to maintain over time. We should also populate a message here vs. supplying an empty message?

test/e2e/plain-v0_provisioner_test.go
timflannagan: That's a good call out - I was fairly confident that you were right, and that we still left the resources on-cluster as the owner references for those resources _should_ be the BundleInstance and not the unpacked Bundle object, but I double checked this assumption locally anyways:

```bash
$ k get bundleinstances olm-crds -o yaml
kind: BundleInstance
metadata:
  name: olm-crds
  resourceVersion: "115145"
  ...
spec:
  bundleName: olm-v0.20.0-crds
  provisionerClassName: core.rukpak.io/plain-v0
status:
  conditions:
  - lastTransitionTime: "2022-03-04T17:17:20Z"
    message: Bundle.core.rukpak.io "olm-v0.20.0-crds" not found
    reason: BundleLookupFailed
    status: "False"
    type: Installed
$ k get crds
NAME                                          CREATED AT             LABELS
bundleinstances.core.rukpak.io                2022-03-03T22:38:42Z   <none>
bundles.core.rukpak.io                        2022-03-03T22:38:42Z   <none>
catalogsources.operators.coreos.com           2022-03-04T17:16:52Z   app.kubernetes.io/managed-by=Helm,core.rukpak.io/owner-kind=BundleInstance,core.rukpak.io/owner-name=olm-crds
clusterserviceversions.operators.coreos.com   2022-03-04T17:16:52Z   app.kubernetes.io/managed-by=Helm,core.rukpak.io/owner-kind=BundleInstance,core.rukpak.io/owner-name=olm-crds
installplans.operators.coreos.com             2022-03-04T17:16:52Z   app.kubernetes.io/managed-by=Helm,core.rukpak.io/owner-kind=BundleInstance,core.rukpak.io/owner-name=olm-crds
olmconfigs.operators.coreos.com               2022-03-04T17:16:52Z   app.kubernetes.io/managed-by=Helm,core.rukpak.io/owner-kind=BundleInstance,core.rukpak.io/owner-name=olm-crds
operatorconditions.operators.coreos.com       2022-03-04T17:16:52Z   app.kubernetes.io/managed-by=Helm,core.rukpak.io/owner-kind=BundleInstance,core.rukpak.io/owner-name=olm-crds
operatorgroups.operators.coreos.com           2022-03-04T17:16:52Z   app.kubernetes.io/managed-by=Helm,core.rukpak.io/owner-kind=BundleInstance,core.rukpak.io/owner-name=olm-crds
operators.operators.coreos.com                2022-03-04T17:16:52Z   app.kubernetes.io/managed-by=Helm,core.rukpak.io/owner-kind=BundleInstance,core.rukpak.io/owner-name=olm-crds
subscriptions.operators.coreos.com            2022-03-04T17:16:52Z   app.kubernetes.io/managed-by=Helm,core.rukpak.io/owner-kind=BundleInstance,core.rukpak.io/owner-name=olm-crds
```

Maybe this is something we can iron out in #99, or tackle during the refactor of the BundleInstance controller? Should I just add a TODO comment for tracking this behavior?

timflannagan:
I'm going to disappear for the rest of the day so I added a TODO comment around re-evaluating the UX around the `Installed: false` BI status condition that @exdx outlined, that we create when an unpacked Bundle was stamped out, but then subsequently removed, and the result is those unpacked Bundle contents are still present on-cluster but the status condition we write to the BI can be confusing to end users.

test/e2e/plain-v0_provisioner_test.go
exdx: I think #99 is slightly different, it's more of a feature to be honest. I think a TODO comment is fine and we can look at it during the refactor. Essentially, we have to figure out how to describe the state of the BI in the test.

Does the BI attempt to redeploy resources that pertain to a particular installation if an object is removed? Looking at the code, it looks like the BI controller sets up dynamic watches for all objects associated with the install, so that would result in a requeue of the BI in case of a deletion event, but I'm not sure if we actually recreate objects. If so, then deleting the bundle would potentially put the BI in a bad state if it can't recreate objects as expected. If on the other hand the BI creates the objects from the bundle once and then does not manage them then removing the Bundle isn't as impactful. The first case seems like more kube-like behavior however. If that's the case then I think we would need to have a custom status for this condition like `BundleRemoved` indicating that the installation was successful but the underlying Bundle is no longer on the cluster. As long as we do the most intuitive thing it shouldn't confuse users.

test/e2e/plain-v0_provisioner_test.go
exdx: Yes, good call. I will open a small PR to address that issue.

test/e2e/plain-v0_provisioner_test.go
joelanford: We need a concept of the BI's current "release". That release becomes the thing that the BI controller uses its dynamic watches to ensure that the underlying release objects aren't modified or deleted.

A release needs to be pivoted. During a pivot, I'd expect that we'd maintain the state of both the old and new releases. If the new release fails to be applied or fails to achieve a successful state within a certain period of time, we need the old release to remain active.

From a BI status standpoint, what I think this boils down to is that we need to differentiate between the state of what's active and the state of what is currently being attempted.

exdx (approved):

PR #105, opened 2022-03-04
e2e: Ensure the Bundle's metadata ConfigMap gets re-created

timflannagan:

### DIFF
--- a/test/e2e/plain-v0_provisioner_test.go
+++ b/test/e2e/plain-v0_provisioner_test.go
@@ -1,4 +1,4 @@
-package e2e_test
+package e2e
 
 import (
 	"context"
@@ -7,7 +7,8 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
-	"github.com/operator-framework/rukpak/api/v1alpha1"
+	rukpakv1alpha1 "github.com/operator-framework/rukpak/api/v1alpha1"
+	"github.com/operator-framework/rukpak/internal/util"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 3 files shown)


timflannagan:
(just dumping what I've had sitting locally since earlier this week so we're not duplicating work while I'm on PTO)

test/e2e/plain-v0_provisioner_test.go
```diff
@@ -22,6 +22,10 @@ const (
 	defaultSystemNamespace = "rukpak-system"
 )
 
+var (
+	ctx = context.Background()
+)
```
joelanford, line 27: Nit/question: This _probably_ doesn't matter, but I wonder if this could ever cause problems as a global, which gets used across all tests.  WDYT about initializing `ctx` in a `BeforeEach` in the root `Describe`?

```go 
var _ = Describe("root", func() {
 	var ctx context.Context
	BeforeEach(func() {
		ctx = context.Background()
	})
})
```

test/e2e/plain-v0_provisioner_test.go
```diff
@@ -86,22 +90,59 @@ var _ = Describe("plain-v0 provisioner", func() {
...
 		})
 
-		// TODO(tflannag): add test specs for ensuring the object and metadata ConfigMaps get re-created when deleted
+		It("should re-create the bundle configmaps", func() {
+			var (
+				cm *corev1.ConfigMap
+			)
+
+			By("getting the metadata configmap for the bundle")
+			configMapTypeRequirement, err := labels.NewRequirement("core.rukpak.io/configmap-type", selection.Equals, []string{"metadata"})
```
joelanford, line 99: For follow-up/TODO: add a constant for these labels?

test/e2e/plain-v0_provisioner_test.go
```diff
@@ -192,3 +233,15 @@ var _ = Describe("plain-v0 provisioner", func() {
 		})
 	})
 })
+
+func newBundleLabelSelector(bundle *v1alpha1.Bundle) labels.Selector {
```
joelanford, line 237: Follow-up/TODO: Maybe an internal util we can reuse?

timflannagan: Sounds reasonable - moved this to the internal package.

test/e2e/plain-v0_provisioner_test.go
timflannagan: I just ended up making a TODO comment and pushing that up. I think this is a general issue across all packages, so hopefully we can have establish some consistency throughout this repository vs. continuing to hardcode labels over time.

test/e2e/plain-v0_provisioner_test.go
timflannagan: It's probably worth eliminating any global state that's not needed so I updated to instantiate this variable in each top-level BeforeEach.

test/e2e/plain-v0_provisioner_test.go
timflannagan: Added an action that creates issues from TODO comments in #106.

exdx (approved):

PR #107, opened 2022-03-04
e2e: use generateName for bundles to avoid test concurrency issues

joelanford:


timflannagan:
I could've sworn I already tried this, and pivoted away from using this an implementation that uses generateName, but I'm blanking on the context now. Maybe I was specifying `ginkgo run -p` and running into parallelization issues?

joelanford:
After this change, I ran the e2e 3-4 times and got consistent passes. It's possible I got lucky and/or introduced another unrelated issue I'm not thinking about.

timflannagan:
Alright I ran these changes on top of on some other e2e changes ~10 times in the background while doing crypto taxes, and I can't think of any reasons why this approach would produce more flakes outside of ginkgo parallelization so I think this is good to go.

timflannagan (approved):

PR #108, opened 2022-03-04
add version information to rukpak built binaries

exdx:
This PR cleans up the Makefile to make it more consistent, and adds a version flag to the binaries that are produced in rukpak: the unpack binary and the plain-v0-provisioner binary. This version information will be helpful as development accelerates and encoding version information in the binary will be useful when debugging. 

```
$ ./bin/plain-v0 --version
Git commit: ba0e6da94d140e7e8b2780bc4f9a541d7f2ccdc8
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,6 +13,11 @@ COPY cmd cmd
 COPY api api
 COPY internal internal
 COPY provisioner provisioner
+# copy git-related information for binary version information
+COPY .git/HEAD .git/HEAD
+COPY .git/refs/heads/. .git/refs/heads
+RUN mkdir -p .git/objects
+
 RUN make build
 
 FROM gcr.io/distroless/static:debug
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,31 @@
+###########################
+# Configuration Variables #
+###########################
+ORG := github.com/operator-framework
+PKG := $(ORG)/rukpak
+IMAGE_REPO=quay.io/operator-framework/plain-v0-provisioner
+IMAGE_TAG=latest
+IMAGE=$(IMAGE_REPO):$(IMAGE_TAG)
+KIND_CLUSTER_NAME ?= kind
+KIND := kind
+VERSION_PATH := $(PKG)/internal/version
+GIT_COMMIT ?= $(shell git rev-parse HEAD)
+PKGS = $(shell go list ./...)
 
+# kernel-style V=1 build verbosity
+ifeq ("$(origin V)", "command line")
+  BUILD_VERBOSE = $(V)
+endif
+
+ifeq ($(BUILD_VERBOSE),1)
+  Q =
+else
+  Q = @
+endif
+
+###############
+# Help Target #
+###############
 .PHONY: help
 help: ## Show this help screen
 	@echo 'Usage: make <OPTIONS> ... <TARGETS>'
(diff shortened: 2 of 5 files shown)


exdx:
Looks like this is still running into an error when running `make build` inside the container: ` fatal: HEAD is neither a commit nor blob`. I think it's important to fix since we want the version to come through on the deployed container image as well. 
Right now the container builds but the version output is empty:
```
$ docker run -it --rm quay.io/operator-framework/plain-v0-provisioner:latest --version
Rukpak version:
Git commit:
```

exdx:
Looks like adding the entire git tree resolves the issue. 

/hold cancel

Makefile
```diff
@@ -1,4 +1,32 @@
...
+ORG := github.com/operator-framework
+PKG := $(ORG)/rukpak
+VERSION_PATH := $(PKG)/pkg/version
+IMAGE_REPO=quay.io/operator-framework/plain-v0-provisioner
+IMAGE_TAG=latest
+IMAGE=$(IMAGE_REPO):$(IMAGE_TAG)
+KIND_CLUSTER_NAME := rukpak-e2e
+KIND := kind
+GIT_COMMIT := $(shell git rev-parse HEAD)
+RUKPAK_VERSION :=  $(shell git describe --match 'v[0-9]*' --tags --always)
```
timflannagan, line 13: I think we can consolidate this into a single variable right now:

```make
VERSION ?= $(shell git describe --match 'v[0-9]*' --tags --always)
```

pkg/version/version.go
```diff
@@ -0,0 +1,13 @@
+package version
```
timflannagan, line 1: Move this to internal while we continue fleshing out the repository structure?

provisioner/plain-v0/main.go
```diff
@@ -57,19 +59,26 @@ func main() {
...
 	opts := zap.Options{
 		Development: true,
 	}
 	opts.BindFlags(flag.CommandLine)
 	flag.Parse()
 
+	if rukpakVersion {
+		fmt.Print(version.String())
+		os.Exit(0)
+	}
```
timflannagan, line 80: This is fine for now, but I think it's worthwhile also logging the version of the binary at controller startup so we have that information available for debugging outside of when users supply the --version flag to this binary.

timflannagan: And that realistically would just look like the following:

```go
setupLog.Info("starting up the provisioner", zap.String("version", version))
```

exdx: Great idea, thanks

joelanford:
I think Go 1.18 makes this much simpler: https://tip.golang.org/doc/go1.18#go-version

Is it worth waiting for that?

tylerslaton (approved):
Looks good to me, it'll be good to get this in. I think Joe's point about Go 1.18 fixing this and making it easier but I feel this is a good stop gap until we update to that. Let's fix up the CI issue and things look good.

timflannagan:
Going to play around with this locally before approving.

timflannagan:
Ran into an error locally:

```
make build-container
docker build -f Dockerfile -t quay.io/operator-framework/plain-v0-provisioner:latest .
Sending build context to Docker daemon    404MB
Step 1/20 : FROM golang:1.17-buster AS builder
 ---> fce131d0083a
Step 2/20 : WORKDIR /workspace
 ---> Using cache
 ---> a501ae717111
Step 3/20 : COPY go.mod go.mod
 ---> Using cache
 ---> 84221af3ddae
Step 4/20 : COPY go.sum go.sum
 ---> Using cache
 ---> 73155e76c4cf
Step 5/20 : RUN go mod download
 ---> Using cache
 ---> 0bc4d486549c
Step 6/20 : COPY Makefile Makefile
 ---> 6527f4cf18ad
Step 7/20 : COPY cmd cmd
 ---> 808ab6ce3c1a
Step 8/20 : COPY api api
 ---> ab3f6067d45a
Step 9/20 : COPY internal internal
 ---> 8bb7e882daf0
Step 10/20 : COPY provisioner provisioner
 ---> c60a98a84d8c
Step 11/20 : COPY pkg pkg
COPY failed: file not found in build context or excluded by .dockerignore: stat pkg: file does not exist
make: *** [Makefile:116: build-container] Error 1
```

Not sure why CI didn't catch this?

timflannagan:
Figured it out: https://github.com/operator-framework/rukpak/pull/121.

Dockerfile
```diff
@@ -13,6 +13,9 @@ COPY cmd cmd
 COPY api api
 COPY internal internal
 COPY provisioner provisioner
+# copy git-related information for binary versioning
+COPY .git .git
```
timflannagan, line 17: Hmm - do we need to copy the entire .git to be able to access version information? Can we supply this Dockerfile with a build-arg variable instead, and that variable can derive the requisite git information before the build stage?

timflannagan: Thinking about this some more: I'm mainly interested in copying over the entire .git directory due to increasing build times if that directory is bloated (and/or contains sensitive information). If we don't like the approach of using a build argument to drive version information to the go binaries, can we exclude everything in the .git directory that we _don't_ need to parse the version information in the .dockerignore?

exdx: Yeah, it makes sense. I tried copying over just the minimum like in OLM's Dockerfile:
```
COPY .git/HEAD .git/HEAD
COPY .git/refs/heads/. .git/refs/heads
RUN mkdir -p .git/objects
```

but then I see this error in the container build logs: `fatal: HEAD is neither a commit nor blob` -- which suggests that the git tree is messed up somehow and I'm not sure how to fix it.

exdx: We could change the version field to just capture the git commit, like in OLM and then the above would work. 
```
VERSION ?= $(shell git rev-parse HEAD)
```
we lose some metadata about the tag and number of commits since the tag though.

tylerslaton: Side note - this made me realize that Combo's version in its local container is not working so I'm making an issue against that repo.

timflannagan (approved):

PR #110, opened 2022-03-04
Add storage happy-path unit test

tylerslaton:
# Summary
Adding initial stab at the unit testing approach in RukPak. Right now the easiest way I found to test the Storage package was via `envtest`. This is because it is actually making and checking resources under the hood so a mocked client would break things partly.

# Open questions
1. Should we mock out Load and Storage partly so we can test them individually or are they better tested together (as seen here) since they're so coupled?
2. Are we okay with using `envtest`? There has been some discussion on it.


Makefile
```diff
@@ -32,9 +32,11 @@ generate: controller-gen ## Generate code and manifests
 
 test: test-unit test-e2e ## Run the tests
 
+ENVTEST_VERSION = $(shell go list -m k8s.io/client-go | cut -d" " -f2 | sed 's/^v0\.\([[:digit:]]\{1,\}\)\.[[:digit:]]\{1,\}$$/1.\1.x/')
 UNIT_TEST_DIRS=$(shell go list ./... | grep -v /test/)
 test-unit: ## Run the unit tests
-	$(Q)go test -count=1 -short $(UNIT_TEST_DIRS)
+	go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
+	eval $$(setup-envtest use -p env $(ENVTEST_VERSION)) && go test -count=1 -short $(UNIT_TEST_DIRS)
```
timflannagan, line 39: Add envtest to the hack/tools/tools.go and use the pattern for installing this binary defined below?

tylerslaton:
Yeah, that's fa

Makefile
tylerslaton: Yeah, that's fair.

exdx (approved):
I'm pretty happy with this approach. If we can use envtest we can have more unit tests and less e2e tests which will overall be better.

Makefile
```diff
@@ -83,14 +84,18 @@ TOOLS_BIN_DIR := $(TOOLS_DIR)/bin
...
+setup-envtest: $(SETUP_ENVTEST) ## Build a local copy of envtest
 
 $(CONTROLLER_GEN): $(TOOLS_DIR)/go.mod # Build controller-gen from tools folder.
 	cd $(TOOLS_DIR); go build -tags=tools -o $(BIN_DIR)/controller-gen sigs.k8s.io/controller-tools/cmd/controller-gen
 $(GINKGO): $(TOOLS_DIR)/go.mod # Build ginkgo from tools folder.
 	cd $(TOOLS_DIR); go build -tags=tools -o $(BIN_DIR)/ginkgo github.com/onsi/ginkgo/ginkgo
 $(GOLANGCI_LINT): $(TOOLS_DIR)/go.mod # Build golangci-lint from tools folder.
 	cd $(TOOLS_DIR); go build -tags=tools -o $(BIN_DIR)/golangci-lint github.com/golangci/golangci-lint/cmd/golangci-lint
+$(SETUP_ENVTEST): $(TOOLS_DIR)/go.mod
+	cd $(TOOLS_DIR); go build -tags=tools -o $(BIN_DIR)/setup-envtest sigs.k8s.io/controller-runtime/tools/setup-envtest
```
timflannagan, line 101: nit: end of file newline

internal/storage/storage_test.go
```diff
@@ -0,0 +1,82 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func TestStoreAndLoad(t *testing.T) {
+	// Setup envtest client
+	kubeclient, err := unit.SetupClient()
+	if err != nil {
+		t.Fatal("failed to create kube client:", err)
+	}
```
timflannagan, line 23: nit: require.NoError?

internal/storage/storage_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+	for _, tt := range []struct {
+		name  string
+		cms   ConfigMaps
+		owner client.Object
+		owned []client.Object
+		err   error
+	}{
+		{
+			name: "can store a value correctly",
```
timflannagan, line 33: flesh out this other store/load edges in a follow up?

internal/storage/storage_test.go
```diff
@@ -0,0 +1,82 @@
...
+					TypeMeta:   metav1.TypeMeta{APIVersion: "apps/v1", Kind: "Deployment"},
+					ObjectMeta: metav1.ObjectMeta{Name: "test-owned"},
+				},
+			},
+			err: nil,
+		},
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			// Create a client.Object to use as an owner for storing via ConfigMaps
+			err = kubeclient.Create(context.TODO(), tt.owner)
```
timflannagan, line 49: nit: context.Background?

PR #114, opened 2022-03-04
fix: update unpack pod name to start with plain-v0

exdx:
Small leftover from #100

### DIFF
--- a/provisioner/plain-v0/controllers/bundle_controller.go
+++ b/provisioner/plain-v0/controllers/bundle_controller.go
@@ -48,7 +48,8 @@ import (
 )
 
 const (
-	bundleUnpackContainerName = "bundle"
+	bundleUnpackContainerName  = "bundle"
+	plainBundleProvisionerName = "plain-v0"
 )
 
 // BundleReconciler reconciles a Bundle object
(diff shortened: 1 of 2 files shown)


joelanford:
Not a blocker, but something to think about: there is a maximum length to that pod name and the config map names. We currently use a prefix and the bundle name, so the longer the prefix, the shorter the allowable bundle name.

We could drop bundle name and use the bundle uid or a hash of something but that hurts debugability and discoverability.

Is it possible to limit bundle name length in the CRD validation schema?

timflannagan:
We could also truncate any built-up Pod/ConfigMap generated metadata.Name to enforce that 63 character limit, but that method also falls victim of decreased debuggability, and that truncation process being used under-the-hood may be surprising behavior to users. You could also validate that the bundle.Name is within a certain allowed buffer of characters, so we have room to fit that into the overall generated metadata.Name but that's also a potentially problematic solution.

exdx:
That's a really interesting point -- I know we saw this issue with CSV names crop up in registry+v1 bundles. Truncation has it's pros and cons. I haven't seen CRD schema validation enforce field length rules, but we can look into it. As more of these custom validation issues on-cluster come into play, it may be worth thinking about implementing an admission webhook.

exdx:
I propose that we get this through as-is and tackle the naming length problem separately as a follow-up, since it affects all pods, configmaps, and other resources that rukpak generates.

timflannagan (approved):
LGTM now that we created a tracking ticket, and this PR gets rebased.

PR #116, opened 2022-03-07
fix: specify linux build flag for make build-local-container

exdx:
Building the local Dockerfile fails on non-linux machines since no GOOS or GOARCH variables are specificied to make build, which causes execution to fail. For example, building the binaries locally on macOS, then putting them in a `gcr.io/distroless/static:debug` linux based image, we see `standard_init_linux.go:228: exec user process caused: exec format` error on startup. This PR sets the default build GOOS flag to linux when building the container locally.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,6 +84,7 @@ bin/unpack:
 build-container: ## Builds provisioner container image locally
 	docker build -f Dockerfile -t $(IMAGE) .
 
+build-local-container: export GOOS=linux
 build-local-container: build ## Builds the provisioner container image using locally built binaries
 	docker build -f Dockerfile.local -t $(IMAGE) .
 


Makefile
```diff
@@ -73,13 +73,15 @@ run: build-local-container kind-load deploy ## Build image and run operator in-c
...
 
 bin/plain-v0:
-	CGO_ENABLED=0 go build -o $@ ./provisioner/plain-v0
+	CGO_ENABLED=0 GOOS=$(GOOS) GOARCH=$(GOARCH) go build -o $@ ./provisioner/plain-v0
 
 bin/unpack:
-	CGO_ENABLED=0 go build -o $@ ./cmd/unpack/...
+	CGO_ENABLED=0 GOOS=$(GOOS) GOARCH=$(GOARCH) go build -o $@ ./cmd/unpack/...
 
 build-container: ## Builds provisioner container image locally
```
joelanford, line 86: I have an Apple Silicon Mac. I think what we want here is:
`make bin/*` - by default does not set GOOS or GOARCH. That way the local build works for the local os/arch.
`make build-container` - by default, builds a linux binary, but uses GOARCH from the local environment. The assumption would be that the developer has a docker daemon running for their host OS's architecture.

I was actually thinking _just_ this:
```suggestion
build-container: GOOS=linux
build-container: ## Builds provisioner container image locally
```

exdx: Sounds good, updated

Makefile
```diff
@@ -73,17 +73,19 @@ run: build-local-container kind-load deploy ## Build image and run operator in-c
 
 # Binary builds
 GO_BUILD := $(Q)go build
+GO_OS=$(if $(GOOS),GOOS=$(GOOS))
```
exdx, line 76: I had to use this extra variable to get the behavior that was described below. Simply setting `GOOS=linux` as an environmental variable to the build-local-container target did not result in a linux binary being built for whatever reason.

joelanford: Hmm I swear I made that change locally and had it working. Let me try something.

Makefile
```diff
@@ -73,17 +73,19 @@ run: build-local-container kind-load deploy ## Build image and run operator in-c
...
+	CGO_ENABLED=0 $(GO_OS) go build -o $@ ./provisioner/plain-v0
 
 bin/unpack:
-	CGO_ENABLED=0 go build -o $@ ./cmd/unpack/...
+	CGO_ENABLED=0 $(GO_OS) go build -o $@ ./cmd/unpack/...
 
 build-container: ## Builds provisioner container image locally
 	docker build -f Dockerfile -t $(IMAGE) .
 
+build-local-container: GOOS=linux
```
joelanford, line 88: Ah ha! I missed an `export` in my original suggestion.

I _think_ this is the only change necessary from what exists on `main`.

```suggestion
build-local-container: export GOOS=linux
```

exdx: Nice, ok that works!

timflannagan (approved):
LGTM - but let's try and a second approval for someone with a mac?

joelanford:
/lgtm

(tested on a mac m1 with a colima arm64 docker socket)

joelanford:
One thing that bugs me _a little_ still is that `bin/*` is dual purpose, so depending on whether I ran `make bin/plain-v0` or `make build-local-container`, I'll end up with either a darwin or linux binary.

Maybe for the build container we should also name the binary with a linux suffix?

exdx:
> Non-blocking: One thing that bugs me _a little_ still is that `bin/*` is dual purpose, so depending on whether I ran `make bin/plain-v0` or `make build-local-container`, I'll end up with either a darwin or linux binary.
> 
> Maybe for the build container we should also name the binary with a linux suffix?

Yes I see what you're saying and I agree -- I also see what you meant by complexity maintaining two Dockerfiles as well now. I can open a small PR for this.

PR #122, opened 2022-03-07
ci(deploy): add new workflow for deploying image with main tag

### LINKED ISSUE #91: Deploy new tags to the quay.io/operator-framework/plain-v0-provisioner repository when changes merge to master
Goal: ensure we have a default official image that we can use when deploying the current rukpak stack.

AC:

- Add a build.yaml workflow that has the `push` trigger enabled for the main branch
- Create a secret in this repository that contains the credentials for that k8s-provisioner repository
- Use the docker login action to login into that quay repository using the credentials secret
- Use the docker-build-push action and build/push to the `main` or `latest` tag

tylerslaton:
Adding a new action that pushes a new `main` tagged image to our quay repository on a push to `main`. There is a follow-up PR coming that will also do this on `release` which will create a semantically-version tagged image.

*Note* this will continue to fail until we add the required secrets to this repository. I didn't have the permission to do this.

Closes #91

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -1,0 +1,27 @@
+name: deploy
+
+on:
+  push:
+    branches:
+    - main
+
+jobs: 
+  main:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Login to quay.io
+        uses: docker/login-action@v1
+        with:
+          registry: quay.io
+          username: ${{ secrets.QUAY_USERNAME }}
+          password: ${{ secrets.QUAY_PASSWORD }}
+
+      - name: Build and push Docker image
+        uses: docker/build-push-action@ad44023a93711e3deb337508980b4b5e9bcdc5dc
+        with:
+          context: .
+          push: true
+          tags: quay.io/operator-framework/plain-v0-provisioner:main
+
--- a/.github/workflows/sanity.yaml
+++ b/.github/workflows/sanity.yaml
@@ -8,7 +8,7 @@ on:
     - main
 
 jobs:
-  sanity:
+  verify:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2


.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,17 @@
+name: deploy
+
+on:
+  push:
+    branches:
+    - main
+
+jobs:
+  main:
+    if: ${{ github.event_name == 'push' }}
```
timflannagan, line 10: Note: this is technically not needed as we only specify the `push` trigger for the `on` value.

tylerslaton: Oop, this is leftover from earlier experimenting. Removing.

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,17 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Login to Quay
+        run: docker login -u="${{ secrets.QUAY_USER }}" -p="${{ secrets.QUAY_SECRET }}" quay.io
```
timflannagan, line 15: nit: can we use the first-class `docker/login-action` action to handle this for us?

tylerslaton: Totally, let me add that.

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,16 @@
...
+
+jobs:
+  main:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Login to Quay
+        run: docker login -u="${{ secrets.QUAY_USER }}" -p="${{ secrets.QUAY_SECRET }}" quay.io
+      - name: Push image to Quay
+        run: make IMAGE_TAG=main push
```
timflannagan, line 16: nit: can we use the `docker/build-push-action` action to handle the build/push process for us?

tylerslaton: Only push-back I'll have on that is we'll need to define the image tag we're using in multiple places (I think) unless I'm missing something.

timflannagan: We can't simply update the Makefile/plain+v0 provisioner manifests to use the main tag instead of the latest tag?

tylerslaton: I was meaning that the image tag that is defined in the Makefile would need to be defined in the action here, right? So at that point we have it written in the Makefile and this action which could lead to drift. Maybe that isn't an issue?

timflannagan: Ah, I understand what you're saying now. I'm not sure I'm too worried about potential skew between what's defined in the Makefile and CI workflows.

tylerslaton: Sounds good! I updated the PR to use the official actions and also support pushing on `releases`.

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,67 @@
+name: deploy
+
+on:
+  push:
+    branches:
+    - main
+  release:
+    types: [published]
```
joelanford, line 8: IMO we should check for tag pushes, not release publishes.

E.g. https://github.com/operator-framework/operator-registry/blob/9999f79682d0073bf9ed5edf2cca1476abf2fc9a/.github/workflows/goreleaser.yaml#L6-L7

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,67 @@
...
+
+      - name: Build and push Docker image
+        uses: docker/build-push-action@ad44023a93711e3deb337508980b4b5e9bcdc5dc
+        with:
+          context: .
+          push: true
+          tags: main
+          labels: ${{ steps.meta.outputs.labels }}
+
+  release:
```
joelanford, line 42: In the past, I've never had to duplicate the jobs in their entirety. We're doing the exact same steps in both, just with different tags right?

Seems like it should be possible to set the desired push tag by using [type=ref](https://github.com/docker/metadata-action#typeref), which does exactly what we want for both branches and tags.

.github/workflows/deploy.yaml
timflannagan: Yep - just hopping on a quick call with Tyler and we talked about using the tag event trigger over this release trigger. I think we'll just specify the main branch trigger, and hold off on adding support for new tags.

.github/workflows/deploy.yaml
timflannagan: Yep - another thing Tyler/I were talking about offline. He's looking into consolidating this.

.github/workflows/deploy.yaml
tylerslaton: I think what I'll end up doing here is getting this PR through for a process to deploy for main pushes. Then have a follow-up that will release on tag pushes. Does that make sense to you all?

joelanford (approved):
LGTM

PR #123, opened 2022-03-07
Makefile: only run 'kind' when necessary

joelanford:
The `ifeq` block in the Makefile controlling access to the `kind` delete call is executed when the Makefile is parsed. This causes the Makefile to take longer to evaluate and it is confusing to see something like "kind: command not found" in the output of an unrelated task.

Before:
```
$ make generate
/bin/sh: kind: command not found
...
```

After:
```
$ make generate
...
```

This PR just forcefully calls `kind delete cluster`, which happily succeeds if the named cluster is non-existent.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,10 +91,7 @@ kind-load: ## Load-image loads the currently constructed image onto the cluster
 	${KIND} load docker-image $(IMAGE) --name $(KIND_CLUSTER_NAME)
 
 kind-cluster: ## Standup a kind cluster for e2e testing usage
-ifeq (1, $(shell ${KIND} get clusters | grep ${KIND_CLUSTER_NAME} | wc -l | xargs))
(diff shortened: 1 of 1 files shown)


timflannagan (approved):
LGTM:

```bash
tflannag@localhost rukpak [fix/version]  kind get clusters
rukpak-e2e
tflannag@localhost rukpak [fix/version]  kind delete cluster --name rukpak-e2ez 
Deleting cluster "rukpak-e2ez" ...
tflannag@localhost rukpak [fix/version]  echo $?
0
tflannag@localhost rukpak [fix/version]  kind delete cluster --name rukpak-e2e
Deleting cluster "rukpak-e2e" ...
tflannag@localhost rukpak [fix/version]  echo $?
0
```

exdx (approved):
Nice!

PR #125, opened 2022-03-07
fix: add back * RBAC rules to successfully install arbitrary manifests

exdx:
Adds * RBAC rules to the provisioner to install combo successfully -- this is part of the documentation work, since in the plain-v0 provisioner docs we use combo as an example bundle to install.

### DIFF
--- a/provisioner/plain-v0/manifests/01_cluster_role.yaml
+++ b/provisioner/plain-v0/manifests/01_cluster_role.yaml
@@ -3,15 +3,6 @@ kind: ClusterRole
 metadata:
   name: plain-v0-provisioner-admin
 rules:
-- apiGroups: ["core.rukpak.io"]
+- apiGroups: ["*"]
   resources: ["*"]
-  verbs: ["*"]
-- apiGroups: ["apiextensions.k8s.io"]
-  resources: ["customresourcedefinitions"]
-  verbs: ["*"]
-- apiGroups: [""]
-  resources: ["*"]
-  verbs: ["*"]
-- apiGroups: ["rbac.authorization.k8s.io"]
-  resources: ["clusterroles", "roles"]
   verbs: ["*"]
--- a/provisioner/plain-v0/manifests/02_role.yaml
+++ b/provisioner/plain-v0/manifests/02_role.yaml
@@ -4,9 +4,6 @@ metadata:
   name: plain-v0-provisioner-admin
   namespace: rukpak-system
 rules:
-- apiGroups: [""]
+- apiGroups: ["*"]
   resources: ["*"]
-  verbs: ["*"]
-- apiGroups: ["rbac.authorization.k8s.io"]
-  resources: ["roles"]
   verbs: ["*"]


timflannagan (commented):
Just _literally_ ran into this locally (and I'm somewhat surprised we don't have e2e coverage for this yet) and basically elevated both the ClusterRole permissions:

```diff
-- apiGroups: ["core.rukpak.io"]
-  resources: ["*"]
-  verbs: ["*"]
-- apiGroups: ["apiextensions.k8s.io"]
-  resources: ["customresourcedefinitions"]
-  verbs: ["*"]
-- apiGroups: [""]
-  resources: ["*"]
-  verbs: ["*"]
-- apiGroups: ["rbac.authorization.k8s.io"]
-  resources: ["clusterroles", "roles"]
-  verbs: ["*"]
+- apiGroups:
+  - "*"
+  resources:
+  - "*"
+  verbs:
+  - "*"
```

Can we try creating the OLM API bundle that's used throughout the e2e suite to see if those changes work as well?

exdx:
> Can we try creating the OLM API bundle that's used throughout the e2e suite to see if those changes work as well?

I'll check those out. I'm looking to avoid * on everything for now, if possible. Otherwise we run into https://github.com/operator-framework/operator-lifecycle-manager/issues/1685 very early on.

timflannagan:
Okay I just manually editted the ClusterRole on a local e2e cluster, creating the following resources 

<details>

```yaml
kind: Bundle
metadata:
  name: olm-v0.20.0-api
spec:
  image: quay.io/tflannag/olm-plain-bundle:olm-api-v0.20.0
  provisionerClassName: core.rukpak.io/plain-v0

---

kind: BundleInstance
metadata:
  name: olm-api
spec:
  provisionerClassName: core.rukpak.io/plain-v0
  bundleName: olm-v0.20.0-api

---

kind: Bundle
metadata:
  name: olm-v0.20.0-crds
spec:
  image: quay.io/tflannag/olm-plain-bundle:olm-crds-v0.20.0
  provisionerClassName: core.rukpak.io/plain-v0

---

kind: BundleInstance
metadata:
  name: olm-crds
spec:
  provisionerClassName: core.rukpak.io/plain-v0
  bundleName: olm-v0.20.0-crds
```

</details>

And got the following error logged in the olm-api BI:

```yaml
status:
  conditions:
  - lastTransitionTime: "2022-03-07T23:34:43Z"
    message: 'rendered manifests contain a resource that already exists. Unable to
      continue with install: could not get information about the resource CatalogSource
      "operatorhubio-catalog" in namespace "olm": catalogsources.operators.coreos.com
      "operatorhubio-catalog" is forbidden: User "system:serviceaccount:rukpak-system:plain-v0-provisioner-admin"
      cannot get resource "catalogsources" in API group "operators.coreos.com" in
      the namespace "olm"'
    reason: InstallFailed
    status: "False"
    type: Installed
```

joelanford:
That error is thanks to the plain-v0 provisioner's use of helm. So we get out-of-the-box sanity checking for existing resources that we don't own. Pretty nice!

I do think we need to give this provisioner */*/* though. Isn't the point that it can install arbitrary kube manifests?

timflannagan:
Yeah it looks like we had that originally in #18 and changed the permissions in #82. I think we'd need to give this provisioner essentially god-like permissions in order to handle the arbitrary resources use-case.

timflannagan:
> So we get out-of-the-box sanity checking for existing resources that we don't own. Pretty nice!

Yep - agreed. Small caveat is that we also need to adhere to helm's adoption of existing resources now, and we'll need to mark existing resources with the annotations it's expecting, so deploying BundleInstances that reference Bundles that contain existing resources doesn't result in an installation failure.

exdx:
If you are both OK going with `*` then I suppose we need to do it. I worry about the implications down the line, but I suppose if we want to support any arbitrary manifest, then we have to have these permissions.

timflannagan (approved):

PR #126, opened 2022-03-08
WIP: Add a BI e2e test that has a dependency on another BI

timflannagan:
Introduce a BI e2e test that's responsible for creating a Bundle/BI for the OLM API(s), ensure that BI results in a failed installation state due to the CRDs being referenced in that API Bundle not existing on cluster, then create the CRD Bundle/BI and wait for the API BI to report a successful installation state once those resources are present.

### DIFF
--- a/test/e2e/plain-v0_provisioner_test.go
+++ b/test/e2e/plain-v0_provisioner_test.go
@@ -345,6 +345,139 @@ var _ = Describe("plain-v0 provisioner bundleinstance", func() {
 		})
 	})
 
+	// TODO: try to create a BundleInstance that references a Bundle that contains
+	// existing resources.
+	When("a BundleInstance depends on another BundleInstance", func() {
+		var (
+			bundle *rukpakv1alpha1.Bundle
+			bi     *rukpakv1alpha1.BundleInstance
+			ctx    context.Context
+		)
+		BeforeEach(func() {
+			ctx = context.Background()
+
+			By("creating the testing OLM APIs Bundle resource")
+			bundle = &rukpakv1alpha1.Bundle{
+				ObjectMeta: metav1.ObjectMeta{
+					GenerateName: "olm-apis",
+				},
+				Spec: rukpakv1alpha1.BundleSpec{
+					ProvisionerClassName: plainProvisionerID,
+					Image:                "quay.io/tflannag/olm-plain-bundle:olm-api-v0.20.0",
+				},
+			}
+			err := c.Create(ctx, bundle)
+			Expect(err).To(BeNil())
+
+			bi = &rukpakv1alpha1.BundleInstance{
+				ObjectMeta: metav1.ObjectMeta{
+					GenerateName: "olm-apis",
+				},
+				Spec: rukpakv1alpha1.BundleInstanceSpec{
+					ProvisionerClassName: plainProvisionerID,
+					BundleName:           bundle.GetName(),
+				},
+			}
+			err = c.Create(ctx, bi)
+			Expect(err).To(BeNil())
+		})
+		AfterEach(func() {
+			By("deleting the testing Bundle resource")
+			Eventually(func() error {
+				return client.IgnoreNotFound(c.Delete(ctx, bundle))
+			}).Should(Succeed())
+
+			By("deleting the testing BundleInstance resource")
+			Eventually(func() error {
+				return client.IgnoreNotFound(c.Delete(ctx, bi))
+			}).Should(Succeed())
+		})
+
+		It("should project a failed installation state", func() {
+			By("projecting a failed installation state")
+			Eventually(func() bool {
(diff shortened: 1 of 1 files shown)


test/e2e/plain-v0_provisioner_test.go
```diff
@@ -345,6 +345,139 @@ var _ = Describe("plain-v0 provisioner bundleinstance", func() {
...
+				ObjectMeta: metav1.ObjectMeta{
+					GenerateName: "olm-crds",
+				},
+				Spec: rukpakv1alpha1.BundleInstanceSpec{
+					ProvisionerClassName: plainProvisionerID,
+					BundleName:           bundleCRDs.GetName(),
+				},
+			}
+			err = c.Create(ctx, biCRDS)
+			Expect(err).To(BeNil())
```
timflannagan, line 456: Note: I was having a hard time structuring this e2e spec and wanted to throw this up now to get some opinions. Right now, the CRD Bundle/BI testing resources get created in a test spec (vs. in a BeforeEach/AfterEach setup) and using the defer keyword in ginkgo seems like a poor idea given potential consistency issues.

timflannagan:
I wanted to throw this up to get some feedback on the overall e2e structure. I was mainly arrived as this structure to avoid having ginkgo create/re-create bundle/bundleinstance resources on a per-spec case, and so I avoided creating tightly scoped "It"(s) in favor of splitting up control flow through "By" statements. We're getting to a point where this package is starting to get somewhat verbose, and I more-or-less copy-pasted the "a BundleInstance targets an invalid Bundle" test spec, and adding logic to create the dependency CRD bundle/bundleinstances, and verify that the original API BI gets to a happy state eventually.

timflannagan:
cc @joelanford / @exdx

test/e2e/plain-v0_provisioner_test.go
```diff
@@ -345,6 +345,139 @@ var _ = Describe("plain-v0 provisioner bundleinstance", func() {
...
+
+			By("deleting the testing BundleInstance resource")
+			Eventually(func() error {
+				return client.IgnoreNotFound(c.Delete(ctx, bi))
+			}).Should(Succeed())
+		})
+
+		It("should project a failed installation state", func() {
+			By("projecting a failed installation state")
+			Eventually(func() bool {
```
joelanford, line 398: Is it possible to return `(bool, error)` from these things so that unexpected errors can short circuit the wait?

test/e2e/plain-v0_provisioner_test.go
joelanford: The pattern I've generally seen is that `BeforeEach` and `AfterEach` are purely for setup/teardown for the parent's `Describe`'d scenario.

The the sibling `It`s are the various things you want to test in that scenario. There also the possibility of nesting more Describe/Before|AfterEach/It scopes on child layers to build a sort of test matrix.

For this particular `Describe`, I might have done something like this:
```go 
Describe("when a BI has a dependency", func() {
	BeforeEach() // setup the BI that has a dependency
	AfterEach() // teardown the BI that has a dependency
	When("the dependency is unmet", func() {
		It("should propagate a failure to the BI")
	})
	When("the dependency is met", func() {
		BeforeEach() // setup the dependency
		AfterEach() // teardown the dependency
		It("should successfully install the BI")
	})
	When("the dependency is initially unmet", func() {
		It("should eventually succeed installing the BI after the dependent is installed", func() {
			By("waiting for the failure")
			By("installing the dependency")
			By("waiting for success")
		})
	})
})
```

These may seem repetitive and overkill, but I've found:
- its easier to add new tests to existing scenarios later
- you can catch hard to find bugs by forcing the controller to permute more state transitions
    - nil to failure
    - nil to direct success
    - nil to failure to success

It's not super obvious now, but when we add more conditions and other status fields, I think the value will become more clear.

test/e2e/plain-v0_provisioner_test.go
```diff
@@ -345,6 +345,139 @@ var _ = Describe("plain-v0 provisioner bundleinstance", func() {
...
+			ctx = context.Background()
+
+			By("creating the testing OLM APIs Bundle resource")
+			bundle = &rukpakv1alpha1.Bundle{
+				ObjectMeta: metav1.ObjectMeta{
+					GenerateName: "olm-apis",
+				},
+				Spec: rukpakv1alpha1.BundleSpec{
+					ProvisionerClassName: plainProvisionerID,
+					Image:                "quay.io/tflannag/olm-plain-bundle:olm-api-v0.20.0",
```
joelanford, line 366: Update now that we have locally built and loaded images?

timflannagan: Definitely - I'm hoping to pick this back up sometime this week.

timflannagan: I ended up creating a new PR as I didn't want to have to deal with a deleted file during rebasing this PR.

PR #127, opened 2022-03-08
add optional GOOS tag to built binaries to specify linux build for build-local-container

exdx:
This PR changes the `make build` target to name the binary with a GOOS suffix to indicate that the binary is for linux systems when `make build-local-container` is ran. This change does not affect the binary name when building locally via `make build` and GOOS is not set.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,15 +76,16 @@ GO_BUILD := $(Q)go build
 build: bin/plain-v0 bin/unpack
 
 bin/plain-v0:
-	CGO_ENABLED=0 go build -o $@ ./provisioner/plain-v0
+	CGO_ENABLED=0 go build -o $@$(BIN_SUFFIX) ./provisioner/plain-v0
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -73,13 +73,14 @@ run: build-local-container kind-load deploy ## Build image and run operator in-c
 
 # Binary builds
 GO_BUILD := $(Q)go build
+GOOS_TAG=$(if $(GOOS),-$(GOOS))
```
joelanford, line 76: Some folks might intentionally set GOOS in their local environment, which would result in this suffix always being added.

I wonder if we could do something like:
```
build-local-container: export BIN_SUFFIX=-$(GOOS)
```

And then in the go build lines:
```
CGO_ENABLED=0 go build -o $@$(BIN_SUFFIX) ./provisioner/plain-v0
```

timflannagan (approved):

PR #133, opened 2022-03-08
test(e2e): add test case for two bi's with duplicate resources

tylerslaton:
## Summary
Adding a e2e test case for creating two BundleInstances that reference the same Bundle. In this particular case we are simulating and expecting a InstallFailed status with a Message of that references the duplicated resources. 

## Open Questions
1. Is there a situation where we want to have multiple BundleInstances reference the same Bundle? If so, how do we get around this test in the future?
2. Is this a valid way of checking this behavior?

### DIFF
--- a/test/e2e/plain-v0_provisioner_test.go
+++ b/test/e2e/plain-v0_provisioner_test.go
@@ -423,8 +423,118 @@ var _ = Describe("plain-v0 provisioner bundleinstance", func() {
 			}).Should(BeTrue())
 		})
 	})
+
+	When("a BundleInstance targets a valid bundle but the bundle contains resources that already exist", func() {
+		var (
+			bundle      *rukpakv1alpha1.Bundle
+			biOriginal  *rukpakv1alpha1.BundleInstance
+			biDuplicate *rukpakv1alpha1.BundleInstance
+			ctx         context.Context
+		)
+		BeforeEach(func() {
+			ctx = context.Background()
+
+			By("creating the testing Bundle resource")
+			bundle = &rukpakv1alpha1.Bundle{
+				ObjectMeta: metav1.ObjectMeta{
+					GenerateName: "olm-crds-valid",
+				},
+				Spec: rukpakv1alpha1.BundleSpec{
+					ProvisionerClassName: plainProvisionerID,
+					Image:                "quay.io/tflannag/olm-plain-bundle:olm-crds-v0.20.0",
+				},
+			}
+			err := c.Create(ctx, bundle)
+			Expect(err).To(BeNil())
+
+			biOriginal = &rukpakv1alpha1.BundleInstance{
+				ObjectMeta: metav1.ObjectMeta{
+					GenerateName: "olm-apis-original",
+				},
+				Spec: rukpakv1alpha1.BundleInstanceSpec{
+					ProvisionerClassName: plainProvisionerID,
+					BundleName:           bundle.GetName(),
+				},
+			}
+
+			err = c.Create(ctx, biOriginal)
+			Expect(err).To(BeNil(), "failed to create original bundle instance")
+
+			biDuplicate = &rukpakv1alpha1.BundleInstance{
+				ObjectMeta: metav1.ObjectMeta{
+					GenerateName: "olm-apis-duplicate",
+				},
+				Spec: rukpakv1alpha1.BundleInstanceSpec{
+					ProvisionerClassName: plainProvisionerID,
+					BundleName:           bundle.GetName(),
+				},
+			}
+
+			err = c.Create(ctx, biDuplicate)
+			Expect(err).To(BeNil(), "failed to create duplicate bundle instance")
+		})
+
+		AfterEach(func() {
+			By("deleting the testing Bundle resource")
+			Eventually(func() error {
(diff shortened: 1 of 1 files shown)


test/e2e/plain-v0_provisioner_test.go
```diff
@@ -423,8 +423,118 @@ var _ = Describe("plain-v0 provisioner bundleinstance", func() {
...
+				ObjectMeta: metav1.ObjectMeta{
+					GenerateName: "olm-apis-duplicate",
+				},
+				Spec: rukpakv1alpha1.BundleInstanceSpec{
+					ProvisionerClassName: plainProvisionerID,
+					BundleName:           bundle.GetName(),
+				},
+			}
+
+			err = c.Create(ctx, biDuplicate)
```
exdx, line 473: I'm not sure but I think this may be a bit racy. Creating one object after the other sequentially like this doesn't necessarily mean that they will be created on the API server in the same order. I guess for a small kind cluster with no load we can expect it to be the case, but for example on Openshift it may not be always true that the second bundle is always seen after the first bundle if you create them at the same time. Maybe we could add an Eventually() or even sleep between creating the first and second BI, just to ensure the test is not prone to these types of timing issues?

exdx (commented):
Nice test. I don't think there is a scenario where we would want two BI's referencing the same Bundle in the new cluster-scoped operator world that Rukpak is based on. In OLM today, operators can install themselves in a specific namespace and watch a set of other namespaces. This behavior supports multitenancy, where multiple copies of the same operator can exist on the same cluster. The idea is to move away from this model to one where every operator (BI) is by default cluster-scoped and no more than one version should ever be running on a cluster at a given time. 

We could prevent this by keeping in memory a map of BIs -> BundleNames, and reference that before installing a new BI. This map would have to stay up to date in the case of deletion events though, and may slowly drift from the state of the cluster. Another option is an admission webhook that intercepts the BI create request, and queries the state of the cluster to see if there is already a BI referencing that bundle. Third option would be to simply let the new BI fail (as we do now) and provide a really clear error message as to why it happened

test/e2e/plain-v0_provisioner_test.go
tylerslaton: What if we had an Eventually stop gap between them, like you said, that attempts to grab each BI. It would guarantee that it is created and thus prevent proceeding before both are existing and ready.

test/e2e/plain-v0_provisioner_test.go
```diff
@@ -423,8 +423,118 @@ var _ = Describe("plain-v0 provisioner bundleinstance", func() {
...
+	})
 })
 
 func conditionsSemanticallyEqual(a, b metav1.Condition) bool {
 	return a.Type == b.Type && a.Status == b.Status && a.Reason == b.Reason && a.Message == b.Message
 }
+
+func conditionsLooselyEqual(a, b metav1.Condition) bool {
+	return a.Type == b.Type && a.Status == b.Status && a.Reason == b.Reason && strings.Contains(b.Message, a.Message)
+}
```
timflannagan, line 540: Yup - this is exactly why I think 1:1 error message (or in this case status condition message) checking makes tests brittle over time. I think this is fine for now, but we really should move towards a strings.Contains approach imo.

test/e2e/plain-v0_provisioner_test.go
```diff
@@ -423,8 +423,118 @@ var _ = Describe("plain-v0 provisioner bundleinstance", func() {
...
+				looselyExpectedMessage := "rendered manifests contain a resource that already exists"
+				expected := metav1.Condition{
+					Type:    "Installed",
+					Status:  metav1.ConditionStatus(corev1.ConditionFalse),
+					Reason:  "InstallFailed",
+					Message: looselyExpectedMessage,
+				}
+
+				return conditionsLooselyEqual(expected, *existing)
+			}).Should(BeTrue())
```
timflannagan, line 529: not blocking: it would be nice to generalize this now that we have this pattern in a couple of different places.

test/e2e/plain-v0_provisioner_test.go
timflannagan: I think this is fine for now imo but Dan's point seems valid - we may need a couple of different helper functions to drive Bundle/BI resources creation at this point, and revisit refactoring the e2e suite structure I had built up.

timflannagan (approved):

test/e2e/plain-v0_provisioner_test.go
tylerslaton: I'll make a todo comment for it 👍🏻

test/e2e/plain-v0_provisioner_test.go
tylerslaton: So you're thinking instead of having exact equality, all of these should be `.Contains`?

test/e2e/plain-v0_provisioner_test.go
tylerslaton: I agree, I have a solution coming out soon for this that should clean that up.

test/e2e/plain-v0_provisioner_test.go
tylerslaton: https://github.com/operator-framework/rukpak/issues/138

PR #136, opened 2022-03-08
doc: cleanup root level README from stale documentation

exdx:
Revamps the top-level README and removes references to the ProvisionerClass API which has not yet been implemented.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,101 +1,94 @@
+[![License](http://img.shields.io/:license-apache-blue.svg)](http://www.apache.org/licenses/LICENSE-2.0.html)
+[![Slack Channel](https://img.shields.io/badge/chat-4A154B?logo=slack&logoColor=white "Slack Channel")](https://kubernetes.slack.com/archives/C0181L6JYQ2)
+
 # RukPak
 
-RukPak runs in a Kubernetes cluster and defines an API for installing cloud native bundle content. 
+RukPak runs in a Kubernetes cluster and defines an API for installing cloud native bundle content.
 
-## Introduction 
+## Introduction
 
-RukPak is a pluggable solution for the packaging and distribution of cloud-native content and supports advanced strategies for installation, updates, and policy.
-The project provides a content ecosystem for installing a variety of artifacts, such as Git repositories, Helm charts, OLM bundles, and more onto a Kubernetes cluster. 
-These artifacts can then be managed, scaled, and upgraded in a safe way to enable powerful cluster extensions. 
+RukPak is a pluggable solution for the packaging and distribution of cloud-native content and supports advanced
+strategies for installation, updates, and policy. The project provides a content ecosystem for installing a variety of
+artifacts, such as Git repositories, Helm charts, OLM bundles, and more onto a Kubernetes cluster. These artifacts can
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,101 +1,95 @@
...
-content is being installed on-cluster and how to create a running instance of the content. 
+At its core, RukPak is a small set of APIs, packaged as Kubernetes CustomResourceDefinitions, and controllers that watch
+for those APIs. These APIs express what content is being installed on-cluster and how to create a running instance of
+the content.
 
-## Components 
+## Components
 
-RukPak is composed of a few primary APIs: `Bundle`, `Instance`, and `ProvisionerClass`
+RukPak is composed of a few primary APIs: `Bundle`, `BundleInstance`, and `Provisioner`. These APIs work together to
```
timflannagan, line 18: Link to the header sections for each of these APIs?

timflannagan, line 18: Link to the header sections for each of these APIs?

timflannagan: I still think that `Provisioner` shouldn't be documented as an "API" but rather a concept?

timflannagan: I still think that `Provisioner` shouldn't be documented as an "API" but rather a concept?

README.md
```diff
@@ -1,101 +1,95 @@
...
-metadata:
-  name: <good name>
-provisioner: <unique id>
-# parameters has no schema, is provisioner-specific
-parameters: {}
-```
+### License
 
-## Contributing 
+Apache License 2.0
```
timflannagan, line 94: Add a github badge for this as well?

timflannagan:
I think is largely fine - I'd like to see us call out concrete examples of provisioners a bit better, but I don't have any huge suggestions right now. Another thing would be linking out to the roadmap (which we don't _really_ have outside of tracked milestones) and provisioner-specific documentation.

README.md
exdx: Yes, I updated the language to not use the term API.

README.md
```diff
@@ -1,101 +1,94 @@
...
+### Provisioner
+
+A Provisioner is a controller that understands `BundleInstance` and `Bundle` APIs and can take action.
+Each `Provisioner` is assigned a unique ID, and is responsible for reconciling a `Bundle` and `BundleInstance` with
+a `spec.provisionerClassName` that matches that particular ID.
 
-RukPak is a community-based open source project, and all are welcome to get involved. 
\ No newline at end of file
+For example, in this repository the [plain-v0](provisioner/plain-v0/README.md) provisioner is implemented.
+The `plain-v0` provisioner is able to unpack a given `plain+v0` bundle onto a cluster and then instantiate it, making
```
tylerslaton, line 93: non-blocking: `plain-v0` followed by `plain+v0` may be confusing, but we've been having that conversation for a bit.

tylerslaton (approved):
Looks good to me, just one thing to think about moving forward.

README.md
exdx: Yes, I agree 😄 , this is the current state of our naming conventions though. When the `plain+v0` bundle spec becomes available, and this sentence can link to that document, hopefully it will be less confusing.

PR #137, opened 2022-03-09
Update plain-v0 README

awgreene:
Introduces a few changes to the plain-v0 README:
- Include shell commands for applying the bundle and bundle instance.
- Change bash and shell codeblocks to console formatting.
- Capitalize Kubernetes.
- Update cleanup section to remove all bundles created during the demo.

Signed-off-by: Alexander Greene <greene.al1991@gmail.com>


tylerslaton (approved):
Nice! I like giving the reader more places to validate they're on the right path.

timflannagan (approved):
TIL that console is the right highlight identifier and not bash.

PR #140, opened 2022-03-09
Integrate goreleaser to produce release images

timflannagan:
Integrate goreleaser into this repository for building and pushing release container images (and multi-arch manifests).


.goreleaser.yml
```diff
@@ -0,0 +1,49 @@
...
+    - amd64
+    ldflags:
+    - -X {{ .Env.PKG }}.GitCommit={{ .ShortCommit }}
+  - id: plain-v0
+    main: ./provisioner/plain-v0
+    binary: plain-v0
+    goos:
+    - linux
+    goarch:
+    - amd64
```
timflannagan, line 27: note: linux-amd64 is the only supported container image right now. I just wanted to get consensus on this before I expand the list of archs (follow-up?).

timflannagan:
I was playing around with this in my fork. You can see that https://quay.io/repository/tflannag/plain-v0-provisioner?tab=tags&tag=latest got populated with a new manifest + individual platform tag.

.goreleaser.yml
tylerslaton: I'd say a follow-up would make sense as we're already adding a lot here.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,43 @@
+name: release
+
+on:
+  push:
+    branches:
+      - '**'
+    tags:
+      - 'v*'
+  pull_request:
```
tylerslaton, line 9: Do we want this to run on pull requests - wouldn't we only release on a tag event? Also, I notice that we are specifying on `push` events but the only thing those trigger is a `checkout`, `install` and `login`.

tylerslaton (commented):
Looks good to me. Just want to iron out some details around what triggers the workflow.

.github/workflows/release.yaml
timflannagan: We're running goreleaser on every PR and merge to any upstream branch right now, but updating the goreleaser arguments and logging into quay when a tag event has been detected. Does that make sense? We could also release rukpak when changes land in the main branch too.

.goreleaser.yml
timflannagan: I think I'll look into combing the deploy + release workflow in a follow-up PRs as well, so both of them build and publish containers through goreleaser.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,43 @@
...
+      if: startsWith(github.ref, 'refs/tags')
+      with:
+        registry: quay.io
+        username: ${{ secrets.QUAY_USERNAME }}
+        password: ${{ secrets.QUAY_PASSWORD }}
+
+    - name: Run goreleaser
+      run: |
+        if [[ $GITHUB_REF == refs/tags/* ]]; then
+          export GORELEASER_ARGS="--rm-dist"
```
exdx, line 39: It looks like we pass the `--rm-dist` argument in the Makefile as well -- should we be passing it twice?

exdx (approved):
one small nit

.github/workflows/release.yaml
timflannagan: So we're actually overriding the default GORELEASER_ARGS variable, and _removing_ the --snapshot option. It's not the most readable (or obvious) implementation, so I can understand any potential confusion.

.goreleaser.yml
tylerslaton: Yeah that sounds like a great idea.

.github/workflows/release.yaml
tylerslaton: So we're running goreleaser on all of these events, but it only actually releases on a tag event?

tylerslaton (approved):
Looks good to me, good to see the release automation come through.

.github/workflows/release.yaml
timflannagan: Yup that's the correct read.

PR #141, opened 2022-03-09
docs(plain-v0): add note about behavior for BI's with same Bundle referenced

tylerslaton:
Adding a small note about the behavior of the plain-v0 provisioner when multiple BundleInstances reference the same Bundle.

### DIFF
--- a/provisioner/plain-v0/README.md
+++ b/provisioner/plain-v0/README.md
@@ -96,6 +96,8 @@ NAME                 DESIRED BUNDLE   INSTALLED BUNDLE   INSTALL STATE          
 my-bundle-instance   my-bundle        my-bundle          InstallationSucceeded   11s
 ```
 
+> Note: Creation of more than one BundleInstance from the same Bundle will likely result in an error.
+
 ### Make bundle content available but do not install it
 
 There is a natural separation between sourcing of the content and application of that content via two separate RukPak


provisioner/plain-v0/README.md
```diff
@@ -96,6 +96,8 @@ NAME                 DESIRED BUNDLE   INSTALLED BUNDLE   INSTALL STATE
 my-bundle-instance   my-bundle        my-bundle          InstallationSucceeded   11s
 ```
 
+> Note: Creation of BundleInstances from the same Bundle will result in an error if any namespaced resources (CRDs, Namespaces, etc) are defined in the Bundle that both BundleInstances reference.
```
timflannagan, line 99: A couple of things:

- Can we move this note to be above the command output?
- CRDs are cluster-scoped resources
- Should we just document that 2+ BundleInstances referencing the same Bundle will result in an error (or it's unsupported workflow)?

tylerslaton: 🤦🏻 I meant to say cluster-scoped here, not name-spaced scoped - my bad. When you say document do you mean run through an example of this like we do for the other use cases?

timflannagan: Nope - s/document/make a note. So basically, instead of specifying what scenarios two BIs targeting the same Bundle is problematic (e.g. it contains cluster-scoped resources), we just make a note that this workflow is unsupported, and you'll likely get an error. Thoughts?

tylerslaton: Ah okay, so moving from partial support to entirely not supporting it. I agree with that and will make the update.

exdx (approved):

PR #144, opened 2022-03-10
reorg: move provisioner to internal; rename plain-v0 to plain

exdx:
This PR renames the plain provisioner from plain-v0 to plain. The idea is that there can be a single controller that has reconciliation logic for many potential versions of bundles, where that reconciliation logic is in separate packages, by bundle version. The top-level controller is then just injected with the latest compatibility matrix of what versions the plain provisioner wants to support. There is therefore no need to append the bundle schema version to the name of the provisioner. The provisioner is now explicitly referenced as `plain` everywhere, including in the spec.provisionerClassName references of Bundles/BundleInstances. 

This PR also moves the entire /provisioner under internal/. This provides more flexibility with establishing a package format going forward.


internal/provisioner/README.md
```diff
@@ -2,5 +2,5 @@
 
 The following subdirectories contain concrete Provisioner implementations.
 
-- [plain-v0](plain-v0/README.md) - provisions "plain+v0" k8s bundles
+- [plain](plain/README.md) - provisions "plain+v0" k8s bundles
```
tylerslaton, line 5:
```suggestion
- [plain](plain/README.md) - provisions "plain" k8s bundles
```

internal/provisioner/plain-v0/main.go
```diff
@@ -19,6 +19,7 @@ package main
 import (
 	"flag"
 	"fmt"
+	controllers2 "github.com/operator-framework/rukpak/internal/provisioner/plain-v0/controllers"
```
exdx: this was just my IDE being weird and was changed back in the subsequent commit.

internal/provisioner/plain/README.md
```diff
@@ -1,8 +1,8 @@
-# Plain-v0 Provisioner
+# Plain Provisioner
 
 ## Summary
 
-The `plain-v0` provisioner is a core rukpak provisioner that knows how to interact with `plain+v0` format bundles.
+The `plain` provisioner is a core rukpak provisioner that knows how to interact with `plain+v0` format bundles.
```
tylerslaton, line 5: Maybe it is worth pointing out very explicitly that `plain` is the Provisioner and `plain+v0` is just the version of the format it uses.

tylerslaton (approved):
It'll be good to get this debt and clarification out prior to the first release, thanks for owning this work @exdx.

internal/provisioner/plain/README.md
exdx: It probably is extremely confusing for newer users now that I'm thinking about it...as someone looking at this for a while it seems normal to me.

exdx: Especially right at the beginning of the README, might be off-putting. I will update.

timflannagan (commented):
It looks like we need to also replace any plain+v0 references?

exdx:
> It looks like we need to also replace any plain+v0 references?

I don't think so -- I think we want to keep referring the bundle type as name+schema version like we do with `registry+v1` bundles. I think that naming convention is good for now, WDYT

timflannagan:
Okay so basically the provisioner -> non-version; bundle format -> versioned. That's fine with me.

timflannagan (approved):

PR #145, opened 2022-03-10
test(e2e): add eventually handler around creation multiple bundle-instances

### LINKED ISSUE #138: Address potentially flakey tests
## Summary:
The test case for duplicated bundle instance resources could could be flakey during the creation of one bundle instance after another. This is because we cannot always guarantee the order of their creation. As a result, it likely makes sense to add some form of `Eventually` to guarantee all resources are ready before proceeding into the specs.

## Context
https://github.com/operator-framework/rukpak/pull/133#pullrequestreview-903544821

tylerslaton:
Closes #138

### DIFF
--- a/test/e2e/plain-v0_provisioner_test.go
+++ b/test/e2e/plain-v0_provisioner_test.go
@@ -459,6 +459,9 @@ var _ = Describe("plain-v0 provisioner bundleinstance", func() {
 
 			err = c.Create(ctx, biOriginal)
 			Expect(err).To(BeNil(), "failed to create original bundle instance")
+			Eventually(func() error { // Guarantee that the original bundle instance is create prior to the duplicate
+				return c.Get(ctx, client.ObjectKeyFromObject(biOriginal), &rukpakv1alpha1.BundleInstance{})
+			}).Should(Succeed(), "failed to validate original bundle instance was created in time")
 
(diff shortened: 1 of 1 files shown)


test/e2e/plain-v0_provisioner_test.go
```diff
@@ -472,6 +475,9 @@ var _ = Describe("plain-v0 provisioner bundleinstance", func() {
 
 			err = c.Create(ctx, biDuplicate)
 			Expect(err).To(BeNil(), "failed to create duplicate bundle instance")
+			Eventually(func() error {
+				return c.Get(ctx, client.ObjectKeyFromObject(biDuplicate), &rukpakv1alpha1.BundleInstance{})
+			}).Should(Succeed(), "failed to validate duplicate bundle instance was created in time")
```
exdx, line 480: nit: do we need to worry about `biDuplicate` being created on time via this Eventually? It's really only that `biOriginal` is established on-cluster before `biDuplicate` that seems important to the context of the test.

exdx (approved):

test/e2e/plain-v0_provisioner_test.go
tylerslaton: Yeah that's a good call out. I think that you're right and we only care about the original being present.

Issue #23, opened 2021-10-22
Investigate how to expose package metadata for the plain manifest bundle format


timflannagan:
The static k8s bundle format is currently defined as a series of static YAML manifests that are wrapped in a scratch container image that get unpacked and stored at runtime. Right now, that implementation has no way of explicitly exposing metadata, like the current package version, the default channel, or any dependencies that come with this package. Exposing that same kind of metadata at the Bundle-level (vs. the package level) may be challenging for gitops workflows, and defining an optional metadata file that gets packaged with the static k8s bundle format could be a decent way to expose those knobs.

joelanford:
We definitely don't want to carry forward the package and channel-level metadata fields (channels, default channel).

I think we will need metadata though (schema: "plain+v1", package name, bundle version).

The registry+v1 bundle also has scorecard and metrics metadata, which we'll need to consider.

exdx:
This seems like a dupe of #178 which was closed since we have https://hackmd.io/Tk6uXGZ2SKqreteOUAF-NA now and a follow-up implementation issue in #242. Should we close this one out as well?

timflannagan:
Yep this looks like a duplicate - closing this out.

PR #149, opened 2022-03-15
doc: add plain+v0 bundle spec

### LINKED ISSUE #75: Document the plain manifest bundle format
Goal: document the plain manifest bundle format and the required Dockerfile format.

Motivation: #71 introduces user documentation for interacting with the plain bundle provisioner, and we're missing the ability to link to documentation that acts as a source-of-truth for this bundle format.

exdx:
Closes #75 .

Additional metadata (as discussed in #23) is being treated as optional in the current plain+v0 spec, since there are no provisioner specific requirements around metadata at this time. This spec omits metadata until we agree on a specified metadata format and the benefits of requiring such a format.

### DIFF
--- a/docs/plain-bundle-spec.md
+++ b/docs/plain-bundle-spec.md
@@ -1,0 +1,160 @@
+# Plain Bundle Spec
+
+## Overview
+
+This document is meant to define the plain bundle format as a reference for those publishing plain bundles for use with
+RukPak. A bundle is a collection of Kubernetes resources that are packaged together for the purposes of installing onto
+a Kubernetes cluster. A bundle can be unpacked onto a running cluster, where controllers can then create the underlying
+content embedded in the bundle. The bundle can be used as the underlying `spec.source` for
+a [Bundle](https://github.com/operator-framework/rukpak#bundle) resource.
+
+A plain bundle is simply a collection of static, arbitrary, Kubernetes YAML manifests in a given directory. A plain
+bundle can consist of a container image, a directory in a git repository, or any other content source that
+the [plain bundle provisioner](https://github.com/operator-framework/rukpak/blob/main/internal/provisioner/plain/README.md)
+supports.
+
+The currently implemented plain bundle format is the `plain+v0` format. The name of the bundle format, `plain+v0`
+combines the type of bundle (plain) with the current schema version (v0).
+The [plain bundle provisioner](https://github.com/operator-framework/rukpak/blob/main/internal/provisioner/plain/README.md)
+is able to source `plain+v0` bundles and install them onto a Kubernetes cluster.
+
+> Note: the plain+v0 bundle format is at schema version v0, which means it's an experimental format that is subject
+> to change.
+
+Supported source types for a plain bundle currently include the following:
+
+* A container image
+* A directory in a git repository
+
+Additional source types, such as a local volume or a generic URI-based resource, are on the roadmap. These source types
+all present the same content, a directory containing static Kubernetes YAML manifests, in a different ways.
+
+## Common Terminology
+
+* `bundle` is a collection of files that define content to be deployed to a cluster
+* `bundle image` is a container image that contains a bundle within its filesystem
+* `bundle git repo` is a git repository that contains a bundle within a directory
+
+## Example
+
+For example, below is a minimal example of a Dockerfile that builds a `plain+v0` bundle image from a directory
+containing static Kubernetes manifests.
+
+```dockerfile
+FROM scratch
+COPY /manifests /manifests
+```
+
+where the given `manifests` directory contains the Kubernetes resources required to deploy an application, for example:
+
+```
+manifests
+├── namespace.yaml
+├── cluster_role.yaml
+├── role.yaml
+├── serviceaccount.yaml
+├── cluster_role_binding.yaml
+├── role_binding.yaml
+└── deployment.yaml
+```
+
+For a bundle git repo, any directory that contains only static Kubernetes manifests checked into a git repository
+accessible via a remote URL can be considered a plain bundle and sourced by the plain provisioner.
+
+> Note: there must be at least one resource in the manifests directory in order for the bundle to be a valid
+> plain+v0 bundle.
+
+## Technical Details
+
+* The static manifests must be located in the root-level /manifests directory in a bundle image for the bundle to be a
+  valid `plain+v0` bundle that the provisioner can unpack. A plain bundle image without a /manifests directory is
+  invalid and will not be successfully unpacked onto the cluster.
+* For a bundle git repo, this limitation does not exist, and the manifests can be in any directory in the repository.
+  The manifest directory is assumed to be ./manifests in a bundle git repo but can be provided at runtime.
+* The manifests directory should be flat: all manifests should be at the top-level with no subdirectories.
+* The plain bundle image can be built from any base image, but `scratch` is recommended as it keeps the resulting bundle
+  image a minimal size.
+* Including any content in the root `manifests` directory of a plain bundle that is not static manifests will result in
+  a failure when creating content on-cluster from that bundle via
+  a [BundleInstance](https://github.com/operator-framework/rukpak#bundleinstance). Essentially, any file that would not
+  successfully `kubectl apply` will result in an error, but multi-object YAML files, or JSON files, are fine. There will
+  be validation tooling provided that can determine whether a given artifact is a valid bundle.
+
+## Quickstart
+
+As an example, we can package the [combo operator](https://github.com/operator-framework/combo) into a `plain+v0` bundle
+image by taking the following steps:
+
+1. First let's pull down the combo repository.
+
+```bash
+$ git clone https://github.com/operator-framework/combo
+```
+
+2. Let's take a look at the combo manifests directory to make sure it is a valid bundle.
+
+```bash
+$ tree combo/manifests
+combo/manifests
+├── 00_combo.io_templates_crd.yaml
+├── 01_combo.io_combinations_crd.yaml
+├── 02_namespace.yaml
+├── 03_service_account.yaml
+├── 04_cluster_role.yaml
+├── 05_cluster_role_binding.yaml
+└── 06_deployment.yaml
+```
+
+This manifests directory is a flattened directory that contains arbitrary Kubernetes manifests and therefore can be
+sourced and unpacked by the plain provisioner. Let's package it up as a plain bundle image.
+
+3. Create a new Dockerfile at the root of the RukPak repository named Dockerfile.example
+
+```bash
+$ touch Dockerfile.example
+```
+
+4. Edit the Dockerfile to include the following:
+
+```bash
+$ cat <<EOF > Dockerfile.example
+FROM scratch
+COPY combo/manifests /manifests
+EOF
+```
+
+5. Build the image using a container tool like docker or podman. Use an image tag that references a repository that you
+   have push access to. For example,
+
+```bash
+$ docker build -f Dockerfile.example -t quay.io/operator-framework/plain-provisioner:example .
+```
+
+6. Push the image to the remote registry
+
+```bash
+$ docker push quay.io/operator-framework/plain-provisioner:example
+```
+
+7. Make sure rukpak is installed locally on a running cluster.
+
+```bash
+$ make run
+```
+
+8. Now that the plain bundle image has been built, it can be referenced in a Bundle and applied to the cluster.
+
+```bash
+$ kubectl apply -f -<<EOF
+apiVersion: core.rukpak.io/v1alpha1
+kind: Bundle
+metadata:
+  name: my-bundle
+spec:
+  source:
+    type: image
+    image:
+      ref: quay.io/operator-framework/plain-provisioner:example
+  provisionerClassName: core.rukpak.io/plain
+EOF
+```
--- a/internal/provisioner/plain/README.md
+++ b/internal/provisioner/plain/README.md
@@ -4,34 +4,13 @@
 
 The `plain` provisioner is a core rukpak provisioner that knows how to interact with bundles of a particular format.
 These `plain+v0` bundles, or plain bundles, are simply container images containing a set of static Kubernetes YAML
-manifests in a given directory. For example, below is an example Dockerfile that builds a plain `plain+v0` bundle from a
-directory containing static manifests.
+manifests in a given directory. For more information on the `plain+v0` format, see 
+the [plain+v0 bundle spec](/docs/plain-bundle-spec.md).
 
-```dockerfile
-FROM scratch
-COPY /manifests /manifests
-```
-
-where the given `manifests` directory contains the Kubernetes resources required to deploy an application, for example:
-
-```
-manifests
-├── namespace.yaml
-├── cluster_role.yaml
-├── role.yaml
-├── serviceaccount.yaml
-├── cluster_role_binding.yaml
-├── role_binding.yaml
-└── deployment.yaml
-```
-
-> Note: The static manifests must be located in the root-level /manifests directory for the bundle image to be a
-valid `plain+v0` bundle that the provisioner can unpack.
-
-The `plain` provisioner is able to unpack a given `plain+v0` bundle onto a cluster and then instantiate it, making
-the content of the bundle available in the cluster. It does so by reconciling `Bundle` and `BundleInstance` types that
-have the `spec.provisionerClassName` field set to `core.rukpak.io/plain`. This field must be set to the correct
-provisioner name in order for the `plain` provisioner to see and interact with the bundle.
+The `plain` provisioner is able to unpack a given `plain+v0` bundle onto a cluster and then instantiate it, making the
+content of the bundle available in the cluster. It does so by reconciling `Bundle` and `BundleInstance` types that have
+the `spec.provisionerClassName` field set to `core.rukpak.io/plain`. This field must be set to the correct provisioner
+name in order for the `plain` provisioner to see and interact with the bundle.
 
 Below is an example of the provisioner reconciliation flow:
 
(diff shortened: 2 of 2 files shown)


docs/plain-bundle-spec.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Overview
+
+This document is meant to define the plain bundle format as a reference for those publishing plain bundles for use with
+RukPak. A bundle is a collection of Kubernetes resources that are packaged into a container image. These bundles can be
+unpacked onto a running cluster, where controllers can then create the underlying content embedded in the bundles.
+
+Plain bundles are simply container images containing a set of static, arbitrary, Kubernetes YAML manifests in a given
+directory. The currently implemented plain bundle schema is the `plain+v0` format. The name of the bundle
+format, `plain+v0` combines the type of bundle (plain) with the current schema version (v0).
```
akihikokuroda, line 11: Where is the current schema version is defined (or specified)?

docs/plain-bundle-spec.md
```diff
@@ -0,0 +1,95 @@
...
+manifests
+├── namespace.yaml
+├── cluster_role.yaml
+├── role.yaml
+├── serviceaccount.yaml
+├── cluster_role_binding.yaml
+├── role_binding.yaml
+└── deployment.yaml
+```
+
```
akihikokuroda, line 37: Can the manifests directory have sub-directories that also have yaml files?

docs/plain-bundle-spec.md
exdx: This is a good question. I'm not sure that we have a formal definition anywhere of how we version the schema. The proposed schema version `v0` is meant to indicate that the schema version is experimental and still not fully set in stone. We can clarify this more in the doc.

docs/plain-bundle-spec.md
```diff
@@ -0,0 +1,101 @@
...
+5. Now that the plain bundle image has been built, it can be referenced in a Bundle and applied to the cluster.
+
+```yaml
+apiVersion: core.rukpak.io/v1alpha1
+kind: Bundle
+metadata:
+  name: plain-provisioner-0.1.0
+spec:
+  image: quay.io/operator-framework/plain-provisioner:example
```
exdx, line 99: nit: this needs to be updated once #160 lands.

internal/provisioner/plain/README.md
```diff
@@ -4,34 +4,13 @@
...
-```
-
-> Note: The static manifests must be located in the root-level /manifests directory for the bundle image to be a
-valid `plain+v0` bundle that the provisioner can unpack.
-
-The `plain` provisioner is able to unpack a given `plain+v0` bundle onto a cluster and then instantiate it, making
-the content of the bundle available in the cluster. It does so by reconciling `Bundle` and `BundleInstance` types that
-have the `spec.provisionerClassName` field set to `core.rukpak.io/plain`. This field must be set to the correct
-provisioner name in order for the `plain` provisioner to see and interact with the bundle.
+The `plain` provisioner is able to unpack a given `plain+v0` bundle onto a cluster and then instantiate it, making the
```
exdx, line 10: The changes here and below in this file are all as a result of markdown linting.

exdx, line 10: The changes here and below in this file are all just linting modifications.

docs/plain-bundle-spec.md
```diff
@@ -0,0 +1,101 @@
...
+3. Build the image using a container tool like docker or podman. Use an image tag that references a repository that you
+   have push access to. For example,
+
+```console
+docker build -f Dockerfile.example -t quay.io/operator-framework/plain-provisioner:example .
+```
+
+4. Push the image to the remote registry
+
+```console
```
timflannagan, line 87: nit: bash

docs/plain-bundle-spec.md
```diff
@@ -0,0 +1,101 @@
...
+LABEL core.rukpak.io.bundle.name.v0=plain-provisioner
+LABEL core.rukpak.io.bundle.version.v0=0.1.0
+
+COPY /internal/provisioner/plain/manifests /manifests
+```
+
+3. Build the image using a container tool like docker or podman. Use an image tag that references a repository that you
+   have push access to. For example,
+
+```console
```
timflannagan, line 81: nit: bash

docs/plain-bundle-spec.md
```diff
@@ -0,0 +1,101 @@
...
+Image labels can be used to describe the type of bundle image as well as the bundle name and version. For example, below
+is a set of optional labels which can be applied to the bundle image:
+
+```dockerfile
+LABEL core.rukpak.io.bundle.mediatype.v0=plain+v0
+LABEL core.rukpak.io.bundle.name.v0=my-operator
+LABEL core.rukpak.io.bundle.version.v0=0.0.1
+```
+
+## Building a plain+v0 Bundle
```
timflannagan, line 61: Duplication compared to the plain provisioner README?

docs/plain-bundle-spec.md
```diff
@@ -0,0 +1,101 @@
...
+* Additional bundle metadata can be provided in two ways: either via image labels or by including a `/metadata`
+  directory inside the bundle. For a`plain+v0` bundle either is acceptable and entirely optional.
+
+Image labels can be used to describe the type of bundle image as well as the bundle name and version. For example, below
+is a set of optional labels which can be applied to the bundle image:
+
+```dockerfile
+LABEL core.rukpak.io.bundle.mediatype.v0=plain+v0
+LABEL core.rukpak.io.bundle.name.v0=my-operator
+LABEL core.rukpak.io.bundle.version.v0=0.0.1
```
timflannagan, line 58: I hate to be a broken record: can we remove these labels as they're essentially a no-op right now (and there's no concrete evidence, at least for now, that these labels will be needed)?

timflannagan: If we're interesting in attaching metadata right now to an individual Bundle, maybe use labels/annotations on the actual Bundle resource?

timflannagan (commented):
I think this is getting closer once we rebase this against the latest master changes now that the Bundle API contains a source type.

In general, I think I had the following outline in mind:

- What is this bundle format
- What are the support source types for this bundle format?
- How can I use this format and/or quickstart
- (optional) How can I validate this bundle format is valid?

WDYT?

docs/plain-bundle-spec.md
exdx: Ok, no problem, I'll take them out -- I think the reason I'm so interested in the labels is because they are used heavily in the `registry+v1` format and I think they may eventually be useful. But as you stated, there's no evidence that they are useful for the plain+v0 format right now and we can drop them for now.

timflannagan: Yep - this can always be an additive change if we ever find the need for exposing container-level labels like we do with the registry+v1 format.

exdx:
> 

That sounds good. Yes, the validation is something that we are currently missing and may want to implement soon. Even though the format is pretty straightforward, having a tool that can validate up-front may be useful.

docs/plain-bundle-spec.md
```diff
@@ -0,0 +1,138 @@
...
+RukPak. A bundle is a collection of Kubernetes resources that are packaged together for the purposes of installing onto
+a Kubernetes cluster. A bundle can be unpacked onto a running cluster, where controllers can then create the underlying
+content embedded in the bundle. The bundle can be used as the underlying `spec.source` for a [Bundle](/README.md#Bundle)
+resource.
+
+A plain bundle is simply a collection of static, arbitrary, Kubernetes YAML manifests in a given directory. A plain
+bundle can consist of a container image, a directory in a git repository, or any other content source that
+the [plain bundle provisioner](/internal/provisioner/plain/README.md) supports.
+
+The currently implemented plain bundle format is the `plain+v0` format. The name of the bundle format, `plain+v0`
```
joelanford, line 15: I've been thinking about our naming. I wonder if we should reuse the `metav1.Object` concept of using apiVersion and kind (e.g. `apiVersion: core.rukpak.io/v0, kind: Plain`).

It gives us everything we need:
- explicit namespacing with a domain name-based group name
- major versioning
- different kinds of bundles
- extensibility (anyone can create a bundle format and a provisioner in their own group)

It also gives us a more structured type to store all of this metadata, so that we don't have to argue over "is it plain.v0, plain+v0, or plain-v0?"

docs/plain-bundle-spec.md
```diff
@@ -0,0 +1,138 @@
...
+> Note: there must be at least one resource in the manifests directory in order for the bundle to be a valid
+> plain+v0 bundle.
+
+## Technical Details
+
+* The static manifests must be located in the root-level /manifests directory in a bundle image for the bundle to be a
+  valid `plain+v0` bundle that the provisioner can unpack. A plain bundle image without a /manifests directory is
+  invalid and will not be successfully unpacked onto the cluster. For a git-based plain bundle, this limitation does not
+  exist, and the manifests can be in any directory in the repository. The manifest directory is assumed to be
+  ./manifests in a git-based plain bundle but can be overwritten.
```
joelanford, line 68: Should we define some terms somewhere?
- `bundle` is a collection of files that define content to be deployed to a cluster.
- `bundle image` is a container image that contains a bundle (any reason the bundle has to be at the image root? we could augment the `image` source type to allow specifying a subdirectory in the image, defaulting to `/manifests`, just like we're doing for bundle repos)
- `bundle git repo` is a git repo that contains a bundle (where the bundle can be anywhere within the repo)

docs/plain-bundle-spec.md
```diff
@@ -0,0 +1,138 @@
...
+* The /manifests directory may contain additional sub-directories which contain static manifests.
```
joelanford, line 69: I'd suggest that we start with a flat directory for now. We can always allow subdirectories in the future without making a breaking change, but we can't go the other way around. Supporting subdirectories might complicate the ordering conversation we were having earlier.

docs/plain-bundle-spec.md
```diff
@@ -0,0 +1,138 @@
...
+* The plain image-based bundle can be built from any base image, but `scratch` is recommended as it keeps the resulting
+  bundle image a minimal size.
+* Additional bundle metadata can be provided in two ways: either via image labels or by including a `/metadata`
+  directory inside the bundle. For a `plain+v0` bundle either is acceptable and entirely optional.
```
joelanford, line 73: IMO, we should not allow metadata to be provided in different ways. That introduces unnecessary ambiguity and complexity:
- do I have to include both?
- if I include both, which takes precedence?
- is precedence always the same no matter the context (e.g. building a catalog vs unpacking the bundle on cluster)?

My vote: Only metadata on the filesystem. That ensures that all filesystem-based sources can provide metadata in a consistent way. The downside is that for images, you have to pull the image to get metadata, rather than just requesting labels. But IMO the tradeoff is easily worth it, at least given my personal experience and confusion with registry+v1 metadata.

For bundle metadata, I think we should define a single file (`bundle.yaml`) that must exist in the root of _ALL_ bundle directories for all bundles types (that would mean we'd need special handling for registry+v1 bundles). And the only required fields would be `apiVersion` and `kind`.  That way any provisioner can look at any bundle and immediately parse this one file to definitely identify what it is looking at.

docs/plain-bundle-spec.md
```diff
@@ -0,0 +1,138 @@
...
+* Including any content in the plain bundle that is not static YAML manifests will result in a failure when creating
+  content on-cluster from that bundle via a [BundleInstance](/README.md#BundleInstance).
```
joelanford, line 75: We should try to be as specific as we can here. For example:
- Are JSON manifests allowed?
- Are multiple objects allowed in a single file?
- Are specific file extensions expected?

This doc should include instructions (maybe TODO for now) for how someone can build and/or validate a plain+v0 bundle.

docs/plain-bundle-spec.md
```diff
@@ -0,0 +1,138 @@
...
+```bash
+$ touch Dockerfile.example
+```
+
+3. Edit the Dockerfile to include the following:
+
+```dockerfile
+FROM scratch
+COPY /internal/provisioner/plain/manifests /manifests
+```
```
joelanford, line 110: Use heredoc syntax so folks following along can copy paste into their terminal?

docs/plain-bundle-spec.md
```diff
@@ -0,0 +1,138 @@
...
+kind: Bundle
+metadata:
+  name: my-bundle
+spec:
+  source:
+    type: image
+    image:
+      ref: quay.io/operator-framework/plain-provisioner:example
+  provisionerClassName: core.rukpak.io/plain
+```
```
joelanford, line 138: Same here re: heredoc (and perhaps add the `kubectl apply` command?)

docs/plain-bundle-spec.md
timflannagan: Yeah I think adding a common terminology section might help make this easier to reason about over time?

docs/plain-bundle-spec.md
timflannagan: +1

docs/plain-bundle-spec.md
timflannagan: Do we even needs to document that we have support (or rather, will have support down the line) for exposing bundle metadata? Can this be added over time once we have a more concrete use case for bundle-level metadata?

joelanford: Yeah, if we completely remove this piece, that also satisfies me for now.

docs/plain-bundle-spec.md
exdx: I do like this idea, I think it's intuitive. Using `Kind` to specify different types of Bundles is nice UX too.

docs/plain-bundle-spec.md
```diff
@@ -0,0 +1,151 @@
+# Plain Bundle Spec
+
+## Overview
+
+This document is meant to define the plain bundle format as a reference for those publishing plain bundles for use with
+RukPak. A bundle is a collection of Kubernetes resources that are packaged together for the purposes of installing onto
+a Kubernetes cluster. A bundle can be unpacked onto a running cluster, where controllers can then create the underlying
+content embedded in the bundle. The bundle can be used as the underlying `spec.source` for a [Bundle](/README.md#Bundle)
```
timflannagan, line 8: nit: broken link.

docs/plain-bundle-spec.md
```diff
@@ -0,0 +1,151 @@
...
+resource.
+
+A plain bundle is simply a collection of static, arbitrary, Kubernetes YAML manifests in a given directory. A plain
+bundle can consist of a container image, a directory in a git repository, or any other content source that
+the [plain bundle provisioner](/internal/provisioner/plain/README.md) supports.
```
timflannagan, line 13: nit: broken link.

docs/plain-bundle-spec.md
```diff
@@ -0,0 +1,151 @@
...
+
+The currently implemented plain bundle format is the `plain+v0` format. The name of the bundle format, `plain+v0`
+combines the type of bundle (plain) with the current schema version (v0).
+The [plain bundle provisioner](/internal/provisioner/plain/README.md) is able to source `plain+v0` bundles and install
```
timflannagan, line 17: nit: broken link.

docs/plain-bundle-spec.md
joelanford: My suggestion for combo for now would be:

```
combo
└── bundle
    └── manifests
        └── manifest.yaml
```

in case we put anything else in the bundle spec that would be a sibling of `manifests`. If we collapse it down and remove the `bundle` directory from the tree, then there's potential for bundle metadata stuff to be intermingled with other operator source code.

That's kind of why I like the idea of `bundle.yaml` with that name; it _could_ live at that root level and not be some non-descript file sitting there.

docs/plain-bundle-spec.md
```diff
@@ -0,0 +1,151 @@
...
+
+* The static manifests must be located in the root-level /manifests directory in a bundle image for the bundle to be a
+  valid `plain+v0` bundle that the provisioner can unpack. A plain bundle image without a /manifests directory is
+  invalid and will not be successfully unpacked onto the cluster.
+* For a bundle git repo, this limitation does not exist, and the manifests can be in any directory in the repository.
+  The manifest directory is assumed to be ./manifests in a bundle git repo but can be provided at runtime.
+* The manifests directory should be flat: all manifests should be at the top-level with no subdirectories.
+* The plain bundle image can be built from any base image, but `scratch` is recommended as it keeps the resulting bundle
+  image a minimal size.
+* Including any content in the plain bundle that is not static manifests will result in a failure when creating content
```
joelanford, line 76:
```suggestion
* Including any content in the root `manifests` directory of a plain bundle that is not static manifests will result in a failure when creating content
```

docs/plain-bundle-spec.md
```diff
@@ -0,0 +1,151 @@
...
+5. Push the image to the remote registry
+
+```bash
+$ docker push quay.io/operator-framework/plain-provisioner:example
+```
+
+6. Make sure rukpak is installed locally on a running cluster.
+
+```bash
+$ make run
```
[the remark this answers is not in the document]
joelanford: Nvmd: I think I see what's happening. We're creating a plain manifest bundle for the provisioner, then we're installing the provisioner separately with `make run`, then we're creating a `Bundle` so that we can see the contents of the provisioner's manifests.

It's pretty confusing IMO to use the plain provisioner's manifests for the bundle image.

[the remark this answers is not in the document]
exdx: That's true, I partially chose it because it meant the user didn't have to go and pull down another repository or set of files in order to build a bundle.

We could potentially use combo but it's `manifests` directory is not flat and so it doesn't adhere to the spec. 

I made https://github.com/exdx/combo-bundle for testing purposes, we could point to that instead, WDYT?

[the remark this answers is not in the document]
timflannagan: Can we simply update combo and deflate everything to a single directory?

timflannagan (approved):
I think this is ready to go as an initial stab, and we can continue to iterate on the content/structure over time. Note: I _think_ there's still an issue with the plain-bundle-spec.md containing broken links still when I use the rendered markdown view, and individual click on those links, but I'll let you decide whether this is needed, and if so, can be done as a quick follow-up.

Let's get another reviewer though to approve. Maybe @joelanford if he has a chance today?

joelanford (approved):

PR #150, opened 2022-03-15
feat(plain): improve user-experience around BundleInstances that reference a deleted Bundle

### LINKED ISSUE #129: Improve the UX around a BI that reference a Bundle that have been removed
Right now when a user creates a set of Bundle/BI resources, the spec.BundleName Bundle resource has been deleted and the BI controller has already stamped out those resources, then the BI controller will report a failed Bundle lookup in it's status:

```bash
$ k apply -f samples/olm-crds.yaml
$ # wait until the BI is reporting a successful installation
$ k get bundleinstances
NAME       DESIRED BUNDLE     INSTALLED BUNDLE   INSTALL STATE      AGE
olm-crds   olm-v0.20.0-crds   olm-v0.20.0-crds   InstallSucceeded   22s
$ k delete bundle olm-v0.20.0-crds
$ k get bundleinstances -o yaml
status:
  conditions:
  - lastTransitionTime: "2022-03-08T16:12:53Z"
    message: Bundle.core.rukpak.io "olm-v0.20.0-crds" not found
    reason: BundleLookupFailed
    status: "False"
    type: Installed
```

The problem with this status condition is that the BI controller places the BI owner references on the underlying Bundle resources, so the result is these resources still existing on cluster:

```bash
$ k get crds -l core.rukpak.io/owner-kind=BundleInstance,core.rukpak.io/owner-name=olm-crds
NAME                                          CREATED AT
catalogsources.operators.coreos.com           2022-03-08T16:12:31Z
clusterserviceversions.operators.coreos.com   2022-03-08T16:12:31Z
installplans.operators.coreos.com             2022-03-08T16:12:31Z
olmconfigs.operators.coreos.com               2022-03-08T16:12:31Z
operatorconditions.operators.coreos.com       2022-03-08T16:12:31Z
operatorgroups.operators.coreos.com           2022-03-08T16:12:31Z
operators.operators.coreos.com                2022-03-08T16:12:31Z
subscriptions.operators.coreos.com            2022-03-08T16:12:31Z
```

tylerslaton:
# Summary
Adding changes that improve the UX around BundleInstances that reference a delete Bundle. Originally, a BundleInstance with an underlying Bundle that was deleted would emit a "BundleLookupFailed" reason for the "Install" condition. Now there is a "HasValidBundle" condition that is emitted on this event and the original "InstallSucceeded" condition along with it.

# Open Questions
Should we have a follow-up PR that adds a printer column to track the state of the "HasValidBundle"?

Closes #129

### DIFF
--- a/internal/provisioner/plain/controllers/bundleinstance_controller.go
+++ b/internal/provisioner/plain/controllers/bundleinstance_controller.go
@@ -108,9 +108,9 @@ func (r *BundleInstanceReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 			bundleStatus = metav1.ConditionFalse
 		}
 		meta.SetStatusCondition(&bi.Status.Conditions, metav1.Condition{
-			Type:    "HasValidBundle",
+			Type:    rukpakv1alpha1.TypeHasValidBundle,
 			Status:  bundleStatus,
-			Reason:  "BundleLookupFailed",
+			Reason:  rukpakv1alpha1.ReasonBundleLookupFailed,
 			Message: err.Error(),
 		})
 		return ctrl.Result{}, client.IgnoreNotFound(err)
@@ -125,16 +125,16 @@ func (r *BundleInstanceReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 				reason = "BundleUnpackRunning"
 			}
 			meta.SetStatusCondition(&bi.Status.Conditions, metav1.Condition{
-				Type:   "Installed",
+				Type:   rukpakv1alpha1.TypeInstalled,
 				Status: metav1.ConditionFalse,
 				Reason: reason,
 			})
 			return ctrl.Result{}, nil
 		}
 		meta.SetStatusCondition(&bi.Status.Conditions, metav1.Condition{
-			Type:    "HasValidBundle",
+			Type:    rukpakv1alpha1.TypeHasValidBundle,
 			Status:  metav1.ConditionFalse,
-			Reason:  "BundleLoadFailed",
+			Reason:  rukpakv1alpha1.ReasonBundleLoadFailed,
 			Message: err.Error(),
 		})
 		return ctrl.Result{}, err
@@ -147,9 +147,9 @@ func (r *BundleInstanceReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		jsonData, err := yaml.Marshal(obj)
 		if err != nil {
 			meta.SetStatusCondition(&bi.Status.Conditions, metav1.Condition{
-				Type:    "InvalidBundleContent",
+				Type:    rukpakv1alpha1.TypeInvalidBundleContent,
 				Status:  metav1.ConditionTrue,
-				Reason:  "ReadingContentFailed",
+				Reason:  rukpakv1alpha1.ReasonReadingContentFailed,
 				Message: err.Error(),
 			})
 			return ctrl.Result{}, err
@@ -166,9 +166,9 @@ func (r *BundleInstanceReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	bi.SetNamespace("")
 	if err != nil {
 		meta.SetStatusCondition(&bi.Status.Conditions, metav1.Condition{
-			Type:    "Installed",
+			Type:    rukpakv1alpha1.TypeInstalled,
 			Status:  metav1.ConditionFalse,
-			Reason:  "ErrorGettingClient",
+			Reason:  rukpakv1alpha1.ReasonErrorGettingClient,
 			Message: err.Error(),
 		})
 		return ctrl.Result{}, err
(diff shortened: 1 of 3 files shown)


internal/provisioner/plain/controllers/bundleinstance_controller.go
```diff
@@ -103,12 +103,13 @@ func (r *BundleInstanceReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
-			bi.Status.InstalledBundleName = ""
+		currentInstalledCondition := meta.FindStatusCondition(bi.Status.Conditions, "Installed")
+		if apierrors.IsNotFound(err) && currentInstalledCondition != nil && currentInstalledCondition.Status != "True" {
+			bi.Status.InstalledBundleName += ""
 		}
 		meta.SetStatusCondition(&bi.Status.Conditions, metav1.Condition{
-			Type:    "Installed",
-			Status:  metav1.ConditionFalse,
+			Type:    "HasValidBundle",
+			Status:  metav1.ConditionTrue,
```
joelanford, line 112: Shouldn't this be `"False"` since we failed to lookup the bundle?
```suggestion
			Status:  metav1.ConditionFalse,
```

internal/provisioner/plain/controllers/bundleinstance_controller.go
```diff
@@ -103,12 +103,13 @@ func (r *BundleInstanceReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 
 	b := &rukpakv1alpha1.Bundle{}
 	if err := r.Get(ctx, types.NamespacedName{Name: bi.Spec.BundleName}, b); err != nil {
-		if apierrors.IsNotFound(err) {
-			bi.Status.InstalledBundleName = ""
+		currentInstalledCondition := meta.FindStatusCondition(bi.Status.Conditions, "Installed")
+		if apierrors.IsNotFound(err) && currentInstalledCondition != nil && currentInstalledCondition.Status != "True" {
+			bi.Status.InstalledBundleName += ""
```
joelanford, line 108: I'm not sure I'm following this code. It looks like a no-op?

Its possible to have both a successfully installed bundle (from a successful previous reconciliation) and an invalid bundle during this reconciliation.

Consider a scenario where the BI first has a valid bundle "foo-v1", and then is changed to a non-existent bundle "foo-v2". I'd expect to see this in the conditions:
```yaml
status:
  conditions:
  - type: Installed
    status: "True"
    reason: InstallSucceeded
  - type: HasValidBundle
    status: "False"
    reason: bundle "foo-v2" not found
  installedBundle: foo-v1
```

Before we return from the reconcile method we need to do two different things:
1. Look to see if anything is currently installed, and set the `Installed` condition accordingly (an error looking this up should result in an `Unknown` status).
2. Lookup and validate the bundle, and set the `HasValidBundle` condition accordingly (an error other than "not found" when looking up the bundle should probably result in an `Unknown` status. Any further validation errors should result in a `False` status).

Only after we've initialized both conditions should we consider ending the reconciliation by returning.

internal/provisioner/plain/controllers/bundleinstance_controller.go
tylerslaton: Yes, it should be. Good catch, I was looking at the Reason where I should've been looking at the Type.

internal/provisioner/plain/controllers/bundleinstance_controller.go
tylerslaton: Ah, I see what you are saying here. That's a really good observation. Let me iron that out.

tylerslaton: @joelanford I believe that I've covered this now. Would you mind giving it a look?

internal/provisioner/plain/controllers/bundleinstance_controller.go
```diff
@@ -103,12 +103,13 @@ func (r *BundleInstanceReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 	if err := r.Get(ctx, types.NamespacedName{Name: bi.Spec.BundleName}, b); err != nil {
+		bundleStatus := metav1.ConditionUnknown
 		if apierrors.IsNotFound(err) {
-			bi.Status.InstalledBundleName = ""
+			bundleStatus = metav1.ConditionTrue
 		}
 		meta.SetStatusCondition(&bi.Status.Conditions, metav1.Condition{
-			Type:    "Installed",
-			Status:  metav1.ConditionFalse,
+			Type:    "HasValidBundle",
```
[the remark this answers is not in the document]
tylerslaton: Yeah, that was something Joe brought up. Do you think we should handle that here or in a tracking issue?

test/e2e/plain_provisioner_test.go
```diff
@@ -317,30 +317,43 @@ var _ = Describe("plain provisioner bundleinstance", func() {
...
+				}
+
+				// Find the HasValidBundle status
+				existingBundleStatus := meta.FindStatusCondition(bi.Status.Conditions, "HasValidBundle")
+				if existingBundleStatus == nil {
+					return false
+				}
+				expectedBundleStatus := metav1.Condition{
+					Type:    "HasValidBundle",
+					Status:  metav1.ConditionStatus(corev1.ConditionTrue),
```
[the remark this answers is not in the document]
tylerslaton: I believe that this is a failure for some code that I wrote to get committed. This should be a status of False.

internal/provisioner/plain/controllers/bundleinstance_controller.go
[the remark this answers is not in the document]
exdx: I would say if you're introducing a new `Type` it should be a constant, at least so it doesn't have to be changed later. For the other `Types` I suppose that can be a follow-up.

exdx (approved):

internal/provisioner/plain/controllers/bundleinstance_controller.go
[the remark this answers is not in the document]
tylerslaton: This should now resolved, I moved them all into the api package.

Issue #153, opened 2022-03-17
Use an http server instead of pod logs for bundle unpacking


joelanford:
When unpacking a bundle, the current architecture launches a pod that reads a bundle's filesystem and writes a JSON map containing file paths and base64-encoded file content to the pod's log. Once it has completed writing the log, it exits.

This is simple and works fairly well, but it has several problems:
1. It requires a duplication of the bundle data on the node's disk. The bundle data is written once in the actual pod filesystem and another time in the pod log.
2. It is possible that the pod logs are eventually garbage collected for complete pods.

We can improve this by having the pod run a simple http server that serves a tar.gz of the bundle content. This solves both problems:
1. On each request it reads the pod filesystem to send the response, so no extra copies are written in the unpack pod
2. It serves the filesystem via http and is now running, thus completely avoiding the problem of pod log garbage collection.

In a follow-up, it might be worth considering another layer of optimization that removes the need for the intermediate configmaps. We can just have the bundle instance controller request the FS directly from the running pod.

exdx:
Moving to backlog for now since it's unclear we will merge the implementation in #155.

tylerslaton:
This is no longer relevant as the implementation of bundle unpacking has changed.

PR #155, opened 2022-03-17
serve bundle tar.gz via http instead of writing pod logs

joelanford:
Retrying #154 here (that one seemed to succumb to today's GitHub problems)

Closes https://github.com/operator-framework/rukpak/issues/153

This PR changes the unpack binary to serve the bundle directory as a tar.gz file on the /content endpoint. It also updates the bundle controller for the plain provisioner to read the bundle content via this new endpoint instead of fetching the pod logs.

### DIFF
--- a/cmd/unpack/main.go
+++ b/cmd/unpack/main.go
@@ -1,20 +1,35 @@
 package main
 
 import (
-	"encoding/json"
+	"archive/tar"
+	"bytes"
+	"compress/gzip"
+	"context"
+	"errors"
 	"fmt"
+	"io"
 	"io/fs"
 	"log"
+	"net/http"
 	"os"
+	"os/signal"
+	"syscall"
+	"time"
+
+	"github.com/gorilla/handlers"
+	"github.com/gorilla/mux"
+	"github.com/spf13/cobra"
 
 	"github.com/operator-framework/rukpak/internal/version"
-
-	"github.com/spf13/cobra"
 )
 
 func main() {
-	var bundleDir string
-	var rukpakVersion bool
+	var (
+		bundleDir  string
+		listenAddr string
+
+		rukpakVersion bool
+	)
 
 	cmd := &cobra.Command{
 		Use:  "unpack",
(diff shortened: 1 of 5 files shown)


cmd/unpack/main.go
```diff
@@ -24,36 +39,99 @@ func main() {
...
-			}); err != nil {
-				log.Fatalf("walk bundle dir %q: %v", bundleDir, err)
+			contentHandler := serveBundle(os.DirFS(bundleDir))
+			h := mux.NewRouter()
+			h.Handle("/content", contentHandler).Methods(http.MethodGet)
+			srv := http.Server{
+				Addr:         listenAddr,
+				Handler:      handlers.CombinedLoggingHandler(os.Stdout, h),
+				ReadTimeout:  1 * time.Second,
+				WriteTimeout: 10 * time.Second,
```
exdx, line 49: Does the server expect a POST/write response?

cmd/unpack/main.go
```diff
@@ -24,36 +39,99 @@ func main() {
...
+	return func(writer http.ResponseWriter, request *http.Request) {
+		buf := &bytes.Buffer{}
+		gzw := gzip.NewWriter(buf)
+		tw := tar.NewWriter(gzw)
+		if err := fs.WalkDir(bundle, ".", func(path string, d fs.DirEntry, err error) error {
+			if err != nil {
+				return err
+			}
+
+			if d.Type()&os.ModeSymlink != 0 {
```
exdx, line 92: I don't believe we support symlinks in the plain bundle spec. I'm also not sure if git repositories can support symlinks on the remote. I guess it doesn't hurt to check but maybe it's adding some unnecessary complexity.

cmd/unpack/main.go
```diff
@@ -24,36 +39,99 @@ func main() {
...
+
+			h, err := tar.FileInfoHeader(info, "")
+			if err != nil {
+				return fmt.Errorf("build tar file info header for %q: %w", path, err)
+			}
+			h.Uid = 0
+			h.Gid = 0
+			h.Uname = ""
+			h.Gname = ""
+			h.Name = path
```
exdx, line 109: are there any additional file changes that need to be made regarding permissions? I know this is an edge case, but we definitely saw cases in registry where files had strange permissions or x-attrs that causes unpacking to fail on the local filesystem.

exdx (commented):
This is cool. My main question is an architectural one -- is this change implementing an http file server for every bundle that's unpacked, and leaving it running on-cluster to continually serve content? Does this scale to users with lots of bundles? Or is it really not that much overhead for us to be concerned about now?

cmd/unpack/main.go
joelanford: I think git does support symlinks. If it is important to support them, we can do a follow-up when that time comes, IMO. There's some complexity with handling them if they point to something that would be outside the tar.

So for now, I opted to just ignore them.

cmd/unpack/main.go
joelanford: The server only accepts `GET` requests to `/content` and responds with the tar.gz of the bundle dir. That's the only possible interaction with the server.

cmd/unpack/main.go
joelanford: That's a good question. By the time we're creating this tar, the files are already on the filesystem where the pod is running. I _think_ we're okay, but if not that seems simple enough to fix later?

cmd/unpack/main.go
exdx: Ok, yes I think I didn't word that correctly -- what I meant to say was, since the server only supports GET requests, should we have a write timeout set at all? Is it being used somewhere?

cmd/unpack/main.go
exdx: Yes, I think that's fair

cmd/unpack/main.go
exdx: That could be addressed in a follow-up, sure

internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -305,20 +296,25 @@ func (r *BundleReconciler) handleCompletedPod(ctx context.Context, u *updater.Up
 }
 
 func (r *BundleReconciler) getBundleContents(ctx context.Context, pod *corev1.Pod) (fs.FS, error) {
-	bundleContentsData, err := r.getPodLogs(ctx, pod)
+	url := fmt.Sprintf("http://%s:8080/content", pod.Status.PodIP)
```
exdx, line 299: Is this a stable IP that we can rely on, or should we have a service in front of the unpacker pod?

exdx (approved):

joelanford:
> Does this scale to users with lots of bundles? Or is it really not that much overhead for us to be concerned about now?

I did a performance test locally and the server handles ~5000 requests per second on my machine. At idle, one pod is consuming 0.0% CPU and about 6-7MB of memory. Honestly 6-7MB seems like a lot when the server is doing nothing. I'll look to see where all that memory is going.

I think we need to expect there to be `2 * # of BIs` bundles at any given time. If someone had 50 operators installed, that would require 600-700MB of memory for the current and next bundle for each of those operators.

We're basically talking a 14MB overhead per installed operator. On "normal" clusters, that seems fine. But on small single node clusters that's a decent chunk.

joelanford:
I just tested a program that literally just waits for a signal and then exits. That also uses 6MB of resident memory. So we might be as optimized as possible (with a Go program at least).

internal/provisioner/plain/controllers/bundle_controller.go
joelanford: Almost positive the pod IP is fixed for the lifetime of the pod. Even if it isn't we read the pod status on every reconcile to get the IP, so I think we're good either way?

cmd/unpack/main.go
joelanford: Ah. As I understand it, ReadTimeout is the timeout for the server reading the request body. WriteTimeout is the timeout for the server writing the response back to the client.

I think its a best practice to set both timeouts to avoid issues with clients that open a connection and then don't send the request body or read the response body.

For an individual bundle, I would think 10 seconds is enough time to send the tar.gz, but I have nothing concrete to back that up. I figure its best to err on the low side and we can always increase it later if we find that 10 seconds is insufficient.

joelanford:
I think a better solution here is for the controller to:
1. Have an admission webhook to make the bundle immutable.
2. During reconciliation, check if the bundle content is already stored in storage. If so return.
3. If not 2, check to see if the bundle unpack pod is present
4. If not 3, create the bundle unpack pod, and return.
5. If 3, pull the bundle content and store it. Then delete the pod and return.

With this approach:
1. Immutability is important because we never have to worry about whether the bundle content we stored is stale.
2. If the bundle content is already stored when we reconcile, there's no need to have the backing pod that sticks around.
3. If the bundle is not stored, we create the pod, keep reconciling until its ready, and then when we find that it's ready, we pull the content, store it, and then delete the pod.

timflannagan:
> 

At a quick glance, this sounds reasonable to me. The only concern I (still) have is whether we're exposing unpacked Bundle contents through the Pod's log sub-resource despite the potential for having to workaround issues around log retention. Is the approach you outlined going in a different direction w.r.t to surfacing + storing Bundle contents asynchronously?

joelanford:
Closing this due to the aforementioned downside that each bundle would have a fairly large memory overhead just to host the unpacked contents to be pulled by the provisioner.

We might need to revisit this or come up with another solution (e.g. #157) if it turns out that reading the pod logs turns out to be unreliable.

PR #156, opened 2022-03-18
Add ValidatingWebhook to limit the bundle name to 40 letters

### LINKED ISSUE #118: Enforce name length and content validation for rukpak generated resources
Rukpak generates kubernetes resources such as pods and configmaps based on the name of the underlying Bundle provided by the user. These resources' names are by default restricted by kubernetes (based on DNS naming conventions) so there should be a way to add these checks before creating resources on-cluster. The checks can be as simple as ensuring that any metadata.name is less than 63 characters and does not contain any invalid characters. See #114 for additional discussion.

akihikokuroda:

This adds ValidatingWebHook for Bundle resource.  It limits the length of the bundle name up-to 40 letters.  The webhook has the certification manager dependency.  The Makefile is updated to install the certification manager for the e2e test.

Closes #118


Makefile
```diff
@@ -95,6 +95,9 @@ deploy: install-apis ## Deploy the operator to the current cluster
 
 run: build-local-container kind-load deploy ## Build image and run operator in-cluster
```
joelanford, line 96: Also add `cert-mgr` to the `run` target?

hack/tools/install-cert-mgr.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/bash
+
+kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.7.1/cert-manager.yaml
+while [[ $(kubectl get pods -l app=webhook -n cert-manager -o 'jsonpath={..status.conditions[?(@.type=="Ready")].status}') != "True" ]]; do echo "waiting for pod" && sleep 5; done
```
joelanford, line 4: Is the wait necessary here? Is there a reason we can't just apply the manifest and then let kubernetes' eventual consistency catch up?

hack/tools/install-cert-mgr.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/bash
+
+kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.7.1/cert-manager.yaml
```
joelanford, line 3: Could we extract the cert manager version into a make version so that we can twist that knob from there?

I also wonder if it isn't just better to inline this `kubectl apply` line in the Makefile, especially if my hunch about the following comment is true (that we don't need to actually wait for cert-manager to be ready)

internal/provisioner/plain/main.go
```diff
@@ -146,6 +146,10 @@ func main() {
 		setupLog.Error(err, "unable to create controller", "controller", "BundleInstance")
 		os.Exit(1)
 	}
+	if err = (&rukpakv1alpha1.Bundle{}).SetupWebhookWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create webhook", "webhook", "Bundle")
+		os.Exit(1)
+	}
```
joelanford, line 152: I think we unfortunately need a separate main.go and deployment for the core webhooks related to the APIs. This validation isn't a concern of the plain provisioner. Its a concern of the core APIs.

internal/provisioner/plain/manifests/07_webhook.yaml
```diff
@@ -0,0 +1,70 @@
...
+  admissionReviewVersions:
+  - v1
+  clientConfig:
+    service:
+      name: rukpak-webhook
+      namespace: rukpak-system
+      path: /validate-core-rukpak-io-v1alpha1-bundle
+      port: 443
+  failurePolicy: Fail
+  #name: vbundle.kb.io
```
joelanford, line 33: Delete this commented line?

internal/provisioner/plain/manifests/07_webhook.yaml
```diff
@@ -0,0 +1,70 @@
...
+  - apiGroups:
+    - core.rukpak.io
+    apiVersions:
+    - v1alpha1
+    operations:
+    - CREATE
+    - UPDATE
+    resources:
+    - bundles
+    #- "*/*"
```
joelanford, line 44: Delete this line?

internal/provisioner/plain/manifests/07_webhook.yaml
```diff
@@ -0,0 +1,70 @@
...
+
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: rukpak-webhook-certificate
+  namespace: rukpak-system
+spec:
+  secretName: rukpak-webhook-certificate
+  dnsNames:
+  - core.rukpak.io
```
joelanford, line 57: Do we need the cert to have the `core.rukpak.io` DNS name?

joelanford (commented):
Nice! Thanks!

If this was the only reason we needed a webhook, I'd say we should patch the CRD with the OpenAPI schema validation. But I think we also need to make the bundle immutable, so this PR will make it easier to add further validation.

internal/provisioner/plain/manifests/07_webhook.yaml
```diff
@@ -0,0 +1,70 @@
...
+---
+
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingWebhookConfiguration
+metadata:
+  name: rukpak-webhook
+  annotations:
+    cert-manager.io/inject-ca-from: rukpak-system/rukpak-webhook-certificate
+webhooks:
+- name: core.rukpak.io
```
joelanford, line 23: Does this name matter? Should it have `bundle` in the name somewhere?

Makefile
akihikokuroda: Yes, I'll add it.

hack/tools/install-cert-mgr.sh
akihikokuroda: The creation of the ValildatingWebHookConfiguration is rejected until the certification manager webhook pod becomes ready.  It takes about 30 seconds after the install in my environment.  So the apply manifests fails without this.

hack/tools/install-cert-mgr.sh
akihikokuroda: I put it the line in the Makefile first then I realized that the wait is necessary so I created a script.  I can make changes to pass the version as an argument for the script.

internal/provisioner/plain/manifests/07_webhook.yaml
akihikokuroda: This one, too.

internal/provisioner/plain/main.go
akihikokuroda: I see. I'll try.  Can it be placed at internal/core?

internal/provisioner/plain/manifests/07_webhook.yaml
```diff
@@ -0,0 +1,70 @@
+apiVersion: v1
```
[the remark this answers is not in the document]
akihikokuroda: I see.  So the api/v1alpha1/webhook.go is in the right place.

internal/provisioner/plain/manifests/07_webhook.yaml
akihikokuroda: This name probably doesn't matter.  It can serve multiple resources in the same group. So it may be better without the bundle in it.

internal/provisioner/plain/manifests/07_webhook.yaml
akihikokuroda: I'm not sure.  The runtime-controller might the group name as the domain name or the name of the webhook.  I'll figure it out.

internal/provisioner/plain/manifests/07_webhook.yaml
joelanford: Got it, makes sense.

joelanford: Looking at this closer, it seems like that while you can have multiple resources in the same group, there's only one client config and therefore only one path for the webhook server URL.

It seems like we'll want at least separation of webhook handlers by object kind, in which case it seems like the name _should_ include "bundle" somehow?

hack/tools/install-cert-mgr.sh
joelanford: I remembered there's a `kubectl` command that makes this easier:
```suggestion
kubectl wait --for=condition=Available --namespace=cert-manager deployment/cert-manager-webhook
```

akihikokuroda: It's nice.  I'll change and move it in the Make file.  I'll add --timeout=60s.  Thanks!

internal/provisioner/plain/manifests/07_webhook.yaml
akihikokuroda: It must be in domain name format but other than that it doesn't matter.  So I can put `bundle-rukpak-webhook.rukpak-system.svc`

internal/provisioner/plain/manifests/07_webhook.yaml
akihikokuroda: I took it out.  It should be `[service name].[namespace name].svc`

internal/provisioner/plain/main.go
akihikokuroda: I added the new main.  I put it in the same image and created another deployment/pod for it.  Please review.  Thanks!

internal/core/manifests/07_webhook.yaml
```diff
@@ -0,0 +1,67 @@
...
+
+---
+
+apiVersion: cert-manager.io/v1
+kind: Issuer
+metadata:
+  name: selfsigned
+  namespace: rukpak-system
+spec:
+  selfSigned: {}
```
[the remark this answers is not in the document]
akihikokuroda: It makes sense sharing the same certificate.

internal/core/main.go
```diff
@@ -0,0 +1,127 @@
...
+	mgr, err := ctrl.NewManager(cfg, ctrl.Options{
+		Scheme:                 scheme,
+		MetricsBindAddress:     metricsAddr,
+		Port:                   9443,
+		HealthProbeBindAddress: probeAddr,
+		LeaderElection:         enableLeaderElection,
+		LeaderElectionID:       "510f803c.olm.operatorframework.io",
+		NewCache: cache.BuilderWithOptions(cache.Options{
+			SelectorsByObject: cache.SelectorsByObject{
+				&rukpakv1alpha1.BundleInstance{}: {},
```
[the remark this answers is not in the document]
akihikokuroda: I don't think watch BIs is necessary.  I just left it there because it may become necessary seen.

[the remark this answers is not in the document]
exdx: Ok, I propose maybe we remove it for now then? Since it's not being used and makes the webhook seem more complex than it is. We could always add it back if we want to extend this validation to BIs as well.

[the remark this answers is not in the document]
akihikokuroda: OK.  I take it out.  Thanks!

internal/core/main.go
```diff
@@ -0,0 +1,126 @@
...
+	if rukpakVersion {
+		fmt.Printf("Git commit: %s\n", version.String())
+		os.Exit(0)
+	}
+
+	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
+	setupLog.Info("starting up the rukpak core webhook", "Git commit", version.String())
+
+	cfg := ctrl.GetConfigOrDie()
+	dependentRequirement, err := labels.NewRequirement("api.core.rukpak.io/owner-kind", selection.In, []string{"Bundle", "BundleInstance"})
```
exdx, line 79: nit: I think we can also update the label requirement to not have the owner-kind be a BundleInstance, since we are no longer watching for those.

api/v1alpha1/webhook.go
```diff
@@ -0,0 +1,71 @@
...
+		Complete()
+}
+
+//+kubebuilder:webhook:path=/validate-core-rukpak-io-v1alpha1-bundle,mutating=false,failurePolicy=fail,sideEffects=None,groups=core.rukpak.io,resources=bundles,verbs=create;update,versions=v1alpha1,name=core.rukpak.io,admissionReviewVersions=v1
+
+var _ webhook.Validator = &Bundle{}
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (r *Bundle) ValidateCreate() error {
+	bundlelog.Info("validate create", "name", r.Name)
```
exdx, line 47: nit: can we move these logs to the Debug level? I think they could potentially be noisy and are not very helpful on their own.

api/v1alpha1/webhook.go
```diff
@@ -0,0 +1,71 @@
...
+// ValidateDelete implements webhook.Validator so a webhook will be registered for the type
+func (r *Bundle) ValidateDelete() error {
+	bundlelog.Info("validate delete", "name", r.Name)
+
+	return nil
+}
+
+func checkNameLength(r *Bundle) error {
+	if len(r.Name) > maxNameLength {
+		return fmt.Errorf("bundle name is too long. max length is %d", maxNameLength)
```
exdx, line 68:
```suggestion
		return fmt.Errorf("bundle name %s is too long: maximum allowed name length is %d", bundle.GetName(), maxNameLength)
```

api/v1alpha1/webhook.go
```diff
@@ -0,0 +1,71 @@
...
+
+//+kubebuilder:webhook:path=/validate-core-rukpak-io-v1alpha1-bundle,mutating=false,failurePolicy=fail,sideEffects=None,groups=core.rukpak.io,resources=bundles,verbs=create;update,versions=v1alpha1,name=core.rukpak.io,admissionReviewVersions=v1
+
+var _ webhook.Validator = &Bundle{}
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (r *Bundle) ValidateCreate() error {
+	bundlelog.Info("validate create", "name", r.Name)
+
+	return checkNameLength(r)
```
exdx, line 49: It might be helpful to emit a log in the case where the name check returns an error, WDYT? Something like:

```
err := checkNameLength(r)
if err != nil {
  bundlelog.Infof("validating bundle create: %s")
} 
return nil
```

api/v1alpha1/webhook.go
akihikokuroda: Sure.  These are generated by the SDK.  I change them to the Debug level.

api/v1alpha1/webhook.go
akihikokuroda: It's a much helpful message.  I change it.  Thanks!

api/v1alpha1/webhook.go
akihikokuroda: I think it's generating enough information to the console and in the log now.

Console output
```
kubectl apply -f bundle_longname.yaml 
Error from server (bundle name combo-78901234567890123456789012345678901 is too long: maximum allowed name length is 40): error when creating "bundle_longname.yaml": admission webhook "bundle-rukpak-webhook.rukpak-system.svc" denied the request: bundle name combo-78901234567890123456789012345678901 is too long: maximum allowed name length is 40
```

Log entry
```
1.647888685788609e+09   DEBUG   bundle-resource validate create {"name": "combo-78901234567890123456789012345678901"}
1.647888685788845e+09   DEBUG   controller-runtime.webhook.webhooks     wrote response  {"webhook": "/validate-core-rukpak-io-v1alpha1-bundle", "code": 403, "reason": "bundle name combo-78901234567890123456789012345678901 is too long: maximum allowed name length is 40", "UID": "f238221b-e3f3-4b1c-a4e9-370f5cf05aa4", "allowed": false}
```
Are these enough or need some more?

internal/core/manifests/07_webhook.yaml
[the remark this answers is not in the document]
joelanford: We can only use the same cert if we're going to put the webhooks behind the same service, right? We basically need one cert per webhook service. If we put multiple webhooks in a single deployment at different http paths, that's one cert. If we have multiple deployments, each with a separate service, that's multiple certs.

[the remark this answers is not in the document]
akihikokuroda: That's true.  Must have the same domain name.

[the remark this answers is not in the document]
exdx: That makes sense. I expect we will have multiple certs then, but since it's so easy to manage with cert-manager it shouldn't be much of an issue.

Makefile
```diff
@@ -91,28 +92,36 @@ install-plain: install-apis ## Install the rukpak CRDs and the plain provisioner
 install: install-plain ## Install all rukpak core CRDs and provisioners
 
 deploy: install-apis ## Deploy the operator to the current cluster
+	kubectl apply -f internal/core/manifests
 	kubectl apply -f internal/provisioner/plain/manifests
 
-run: build-local-container kind-load deploy ## Build image and run operator in-cluster
+run: build-local-container kind-load cert-mgr deploy ## Build image and run operator in-cluster
+
+cert-mgr:
```
timflannagan, line 100: Add a comment about this target so it shows up in the `make help` output locally?

api/v1alpha1/webhook.go
```diff
@@ -0,0 +1,71 @@
...
+)
+
+// Maximum length of bundle name
+const maxNameLength = 40
+
+// log is for logging in this package.
+var bundlelog = logf.Log.WithName("bundle-resource")
+
+func (r *Bundle) SetupWebhookWithManager(mgr ctrl.Manager) error {
+	bundlelog.Info("SetupWebhookWithManager")
```
timflannagan, line 35: nit: I think we can drop this log statement?

api/v1alpha1/webhook.go
```diff
@@ -0,0 +1,71 @@
...
+		Complete()
+}
+
+//+kubebuilder:webhook:path=/validate-core-rukpak-io-v1alpha1-bundle,mutating=false,failurePolicy=fail,sideEffects=None,groups=core.rukpak.io,resources=bundles,verbs=create;update,versions=v1alpha1,name=core.rukpak.io,admissionReviewVersions=v1
+
+var _ webhook.Validator = &Bundle{}
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (r *Bundle) ValidateCreate() error {
+	bundlelog.V(1).Info("validate create", "name", r.Name)
```
timflannagan, line 47: Should we bump the log verbosity of this to a higher level? Is there a way we can inject a logger such that we don't rely on a global logger for this package? Can we wrap this ValidateCreate method and inject a logger as a parameter?

api/v1alpha1/webhook.go
```diff
@@ -0,0 +1,71 @@
...
+
+// ValidateDelete implements webhook.Validator so a webhook will be registered for the type
+func (r *Bundle) ValidateDelete() error {
+	bundlelog.V(1).Info("validate delete", "name", r.Name)
+
+	return nil
+}
+
+func checkNameLength(r *Bundle) error {
+	if len(r.Name) > maxNameLength {
```
timflannagan, line 67: Are Bundles that contain a metadata.Name == 40 characters invalid?

Makefile
akihikokuroda: I'll add it.  Thanks!

api/v1alpha1/webhook.go
akihikokuroda: Yes, I'll take it out.

api/v1alpha1/webhook.go
akihikokuroda: It is `DEBUG` level now.
`1.648259706745181e+09   DEBUG   bundle-resource validate create {"name": "combo-78901234567890123456789012345678901"}`
Another log entry is generated by the controller-runtime.  It's also `DEBUG` level.
`1.6482597067455616e+09  DEBUG   controller-runtime.webhook.webhooks     wrote response  {"webhook": "/validate-core-rukpak-io-v1alpha1-bundle", "code": 403, "reason": "bundle name combo-78901234567890123456789012345678901 is too long: ma\
ximum allowed name length is 40", "UID": "df80b906-3937-4668-ac87-e3b3f846e8e3", "allowed": false}`

This package imports `logf "sigs.k8s.io/controller-runtime/pkg/log"`  and create log by this `bundlelog = logf.Log.WithName("bundle-resource")`
So it's possible to change the logger statically.  I'm not sure how to wrap the ValidateCreate function.

joelanford: I believe controller-runtime injects a logger into webhooks if they implement the [`inject.Logger`](https://github.com/kubernetes-sigs/controller-runtime/blob/6627ea2a1c4a7f6741792438e8333e96364028f1/pkg/runtime/inject/inject.go#L151-L155) interface

akihikokuroda: Yes, I looked into it.  It seems that by adding `WithLogger (https://github.com/kubernetes-sigs/controller-runtime/blob/6627ea2a1c4a7f6741792438e8333e96364028f1/pkg/builder/controller.go#L152) in SetupWebhookWithManager, we can override the lodder used by the webhook controller.
What logger should be used in this pacakge?

akihikokuroda: I oversighted.  The WebwookBuilder doesn’t have the function.

akihikokuroda: I would like to handle this in a separate issue.

akihikokuroda:
I changed the  file locations according to the suggestion in the slack.

manifests/bundle-webhook/02_role.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: rulpak-core-admin
+  namespace: rukpak-system
+rules:
+- apiGroups: ["*"]
+  resources: ["*"]
+  verbs: ["*"]
```
timflannagan, line 9: Should we tone down these permissions down to what we just need to have Bundle validation? This could also be tackled in a follow-up.

manifests/bundle-webhook/01_cluster_role.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: rukpak-core-admin
+rules:
+- apiGroups: ["*"]
+  resources: ["*"]
+  verbs: ["*"]
```
timflannagan, line 8: Same comment around the permissions here.

cmd/main.go
```diff
@@ -0,0 +1,126 @@
...
+	if err != nil {
+		setupLog.Error(err, "unable to start manager")
+		os.Exit(1)
+	}
+
+	if err = (&rukpakv1alpha1.Bundle{}).SetupWebhookWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create webhook", "webhook", "Bundle")
+		os.Exit(1)
+	}
+	//+kubebuilder:scaffold:builder
```
exdx, line 110: nit: I think this kubebuilder marker snuck in here mistakenly?

api/v1alpha1/webhook.go
timflannagan: Nope - I misread the code.

cmd/main.go
```diff
@@ -0,0 +1,126 @@
...
+)
+
+var (
+	scheme   = runtime.NewScheme()
+	setupLog = ctrl.Log.WithName("setup")
+)
+
+func init() {
+	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
+	utilruntime.Must(apiextensionsv1.AddToScheme(scheme))
```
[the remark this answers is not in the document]
akihikokuroda: It is not necessary.  I take it out.  Thanks!

cmd/main.go
akihikokuroda: I'm not sure.  This is generated by SDK and all other main.gos have it.

exdx (approved):
This LGTM. Let's get one more approval and merge.

tylerslaton (approved):
Nice, LGTM. This PR got everything I was looking for in.

PR #157, opened 2022-03-19
WIP: Proof of concept to delete unpack pods, immutable bundles, and deny configmap deletes

joelanford:
This PR clears completed pods after bundle contents are stored to configmaps

It also adds two webhooks:
1. Bundles: Spec is immutable, Status is immutable once Unpacked phase is reached
2. Configmaps:
    - Any configmap in the rukpak-system namespace with a
      core.rukpak.io/owner-kind==Bundle label cannot be deleted if its
      owning Bundle exists.
    - Any configmap in the rukpak-system namespace with a
      core.rukpak.io/owner-kind==Bundle label cannot have any of its
      core.rukpak.io labels removed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,29 +1,8 @@
-FROM golang:1.17-buster AS builder
-
-WORKDIR /workspace
-# Copy the Go Modules manifests
-COPY go.mod go.mod
-COPY go.sum go.sum
-# cache deps before building and copying source so that we don't need to re-download as much
-# and so that source changes don't invalidate our downloaded layer
-RUN go mod download
-
-COPY Makefile Makefile
-COPY cmd cmd
-COPY api api
-COPY internal internal
-# copy git-related information for binary version information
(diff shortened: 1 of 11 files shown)


timflannagan (commented):
I think I'm warming up to this idea after playing around with embedded Bundles and we should start entertaining pushing this through. Can we rebase this PR since it's been sitting for a while?

timflannagan:
Oh it looks like I had some review comments pending since I had taken a look last time. I think we can disregard some of those comments.

joelanford:
This PR is somewhat outdated (we already have immutable bundle specs and we no longer store bundles in configmaps).

The piece of this that we may want to revisit is deleting the unpack pods as soon as we've saved the contents in the provisioner's bundle storage. There are implications to consider if we do that though (e.g. what happens if the bundle says its unpacked but the backing bundle contents were somehow corrupted or deleted?)

PR #160, opened 2022-03-23
Introduce a union-based Bundle source type

timflannagan:
Introduce a union-based Bundle source type approach, and refactor the Bundle spec.Image to be it's own independent content source type.


timflannagan:
cc @joelanford

internal/updater/updater_test.go
```diff
@@ -68,7 +68,11 @@ var _ = Describe("Updater", func() {
 			},
 			Spec: rukpakv1alpha1.BundleSpec{
 				ProvisionerClassName: provisionerClassName,
-				Image:                "quay.io/tflannag/olm-plain-bundle:olm-crds-v0.20.0",
+				Source: rukpakv1alpha1.BundleSource{
+					Image: rukpakv1alpha1.ImageSource{
```
akihikokuroda, line 72: Type is missing.

test/e2e/plain_provisioner_test.go
```diff
@@ -45,7 +45,12 @@ var _ = Describe("plain provisioner bundle", func() {
 				},
 				Spec: rukpakv1alpha1.BundleSpec{
 					ProvisionerClassName: plainProvisionerID,
-					Image:                "quay.io/tflannag/olm-plain-bundle:olm-crds-v0.20.0",
+					Source: rukpakv1alpha1.BundleSource{
+						Type: "image",
```
akihikokuroda, line 49: Would you add constants for types in the bundle_types.go ?

timflannagan: Yep - good catch.

api/v1alpha1/bundle_types.go
```diff
@@ -40,9 +40,20 @@ const (
...
+	// Source defines the configuration for the underlying Bundle content.
+	Source BundleSource `json:"source"`
+}
 
+type BundleSource struct {
+	// Type defines the kind of Bundle content being sourced.
+	Type string `json:"type"`
 	// Image is the bundle image that backs the content of this bundle.
-	Image string `json:"image"`
+	Image ImageSource `json:"image,omitempty"`
```
joelanford, line 51: Should this be a pointer to an `ImageSource` because we expect exactly one non-`nil` source struct?

```suggestion
	Image *ImageSource `json:"image,omitempty"`
```

Also, do we need to add any of the union type comment markers that are documented in the KEP: https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/1027-api-unions/README.md?

(I'm wondering if controller-tools has any support for these...)

exdx: I think a pointer here makes sense. 

I played around with the union type markers locally but it didn't look like the codegen was picking them up. Based on the open https://github.com/kubernetes-sigs/controller-tools/issues/461 I guess controller-gen doesn't have support for union semantics. We would have to hand-craft the CRD schema or move the union type validation to a webhook (the proposed bundle webhook would be a good candidate).

timflannagan: Yeah I _think_ this needs to be a pointer too - good catch.

timflannagan: Dan and I were playing around kubebuilder markers, and whether we can even manually make direct CRD changes (e.g. basically what I had to do with metering for an API that emulates this design), and as expected, the generation tooling stomps on those direct changes. Maybe this can be tackled in the Bundle webhook @akihikokuroda is working on, or internally in the provisioner's bundle controller?

api/v1alpha1/bundle_types.go
```diff
@@ -40,9 +42,20 @@ const (
...
 
+type BundleSource struct {
+	// Type defines the kind of Bundle content being sourced.
+	Type string `json:"type"`
 	// Image is the bundle image that backs the content of this bundle.
-	Image string `json:"image"`
+	Image *ImageSource `json:"image,omitempty"`
+}
+
+type ImageSource struct {
```
exdx, line 56: I think we have to update the corresponding printer column to reflect the new spec field on the API
```
//+kubebuilder:printcolumn:name=Image,type=string,JSONPath=`.spec.source.image.ref`
```

See https://github.com/operator-framework/rukpak/issues/163 for tracking issue on generalizing the field.

api/v1alpha1/bundle_types.go
joelanford: > the generation tooling stomps on those direct changes

Yeah, we either need the generator to understand how to add union semantics or we need a way to patch the generated CRD as part of the generation process.

SDK/Kubebuilder projects use kustomize to make it easy to integrate arbitrary patches into the manifest generation process. Is that a reasonable dependency to carry for this purpose?

internal/provisioner/plain/README.md
```diff
@@ -60,7 +60,10 @@ kind: Bundle
 metadata:
   name: my-bundle
 spec:
-  image: my-bundle@sha256:xyz123
+  source:
+    type: image
+    image:
+      ref:: my-bundle@sha256:xyz123
```
joelanford, line 66:
```suggestion
      ref: my-bundle@sha256:xyz123
```

internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -224,7 +220,7 @@ func (r *BundleReconciler) ensureUnpackPod(ctx context.Context, bundle *rukpakv1
 			pod.Spec.Containers = make([]corev1.Container, 1)
 		}
 		pod.Spec.Containers[0].Name = bundleUnpackContainerName
-		pod.Spec.Containers[0].Image = bundle.Spec.Image
+		pod.Spec.Containers[0].Image = bundle.Spec.Source.Image.Ref
 		pod.Spec.Containers[0].ImagePullPolicy = corev1.PullAlways
```
joelanford, line 224: Unrelated to this PR, but I think we should remove this line so that we inherit the default from the pod spec API, which is:
- If `image` uses a tag and the tag is `:latest`, use `Always`
- Otherwise, use `IfNotPresent`.

internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -224,7 +220,7 @@ func (r *BundleReconciler) ensureUnpackPod(ctx context.Context, bundle *rukpakv1
 			pod.Spec.Containers = make([]corev1.Container, 1)
 		}
 		pod.Spec.Containers[0].Name = bundleUnpackContainerName
-		pod.Spec.Containers[0].Image = bundle.Spec.Image
+		pod.Spec.Containers[0].Image = bundle.Spec.Source.Image.Ref
```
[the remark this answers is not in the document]
timflannagan: Yep - it seems like we may need to house a switch/case statement in the Reconcile method that basically builds up a concrete Unpacker type for unpacking/sourcing different Bundle source types. Maybe this can be done as a follow-up?

api/v1alpha1/bundle_types.go
exdx: Personally I would be interested in using kustomize more and potentially bringing it in if it helps deliver this functionality

exdx (approved):

internal/provisioner/plain/controllers/bundle_controller.go
timflannagan: Yep that's fine with me - made this change in a separate PR: https://github.com/operator-framework/rukpak/pull/165.

joelanford (approved):

PR #162, opened 2022-03-23
add unpacked object length check 

exdx:
This check ensures that in the case where no objects where unpacked as a result of the unpack pod a failure status will propagate to the Bundle. 

This prevents the case where zero objects are stored successfully and the Bundle is considered Unpacked even though no content was persisted to the cluster.

### DIFF
--- a/internal/provisioner/plain/controllers/bundle_controller.go
+++ b/internal/provisioner/plain/controllers/bundle_controller.go
@@ -273,6 +273,10 @@ func (r *BundleReconciler) handleCompletedPod(ctx context.Context, u *updater.Up
 	if err != nil {
 		return updateStatusUnpackFailing(u, fmt.Errorf("get objects from bundle manifests: %w", err))
 	}
+	if len(objects) == 0 {
+		return updateStatusUnpackFailing(u, errors.New("invalid bundle: found zero objects: "+
+			"plain+v0 bundles are required to contain at least one object"))
+	}
 
 	if err := r.Storage.Store(ctx, bundle, objects); err != nil {
 		return updateStatusUnpackFailing(u, fmt.Errorf("persist bundle objects: %w", err))


joelanford (commented):
This is interesting. On one hand, this could certainly be useful in the case where someone accidentally pushed a bundle with no manifests or if they somehow specified the wrong place for the unpacker to look for manifests.

On the other, there's nothing really invalid about a bundle with no manifests or a bundle instance that manages an empty bundle, is there?

Maybe its just a terminology thing that I'm thinking off. Unpacking didn't actually fail. It just resulted in an empty set of manifests.

timflannagan:
@joelanford The problem here (or at least this is a problem in my eyes) is we're marking the Bundle/BI's as successful unpack/installation states despite now resources being stamped out, so it's a UX problem? We could handle this at the unpacking layer, but I think I'm convinced we should make the unpacker a fairly trivial implementation, and we should let clients handle this kind of edge case. Thoughts?

joelanford:
> we're marking the Bundle/BI's as successful unpack/installation states despite now resources being stamped out, so it's a UX problem?

_Technically_ it was successful though right? We did what was asked, we didn't encounter any actual errors, and the result is an "Installed" bundle with nothing actually applied to the cluster. 

I suppose we could say that the spec of a plain manifest bundle _must_ have at least one `client.Object` in it. If that's what we're saying, then I'm on board with this change, but I'd say we should change the message to be something like `"found 0 objects: plain manifest bundles are required to contain at least one object"`

timflannagan:
Yeah that's fine with me - I think I'm mainly worried about how we're projecting this status back to the user. If I install a Bundle, and there's an issue with how I set that up, I'd expect the plain provisioner to tell me that vs. having to dig into this locally myself and figure out what went wrong. In that same vein, if I accidentally pass a container image that contains no YAML manifests, I think I'd expect there to be some sort of warning that gets surfaced back up to me?

exdx:
> `"found 0 objects: plain manifest bundles are required to contain at least one object"`

I like this but we have to be a bit more generic since the current git-based bundle implementation also creates an unpack pod and hits this codepath. Maybe something like `found 0 objects: valid plain bundle sources are required to contain at least one object`?

exdx:
> Figure out what version of the plain bundle its looking at. Right now, there's only one version of the plain bundle, but getting the machinery into the provisioner now seems low cost and it will make it much easier later if we need to add another plain bundle version.

That sounds good -- we would need a way to encode metadata into the bundle. That can be accomplished via image labels or a metadata directory. I touch on those in #149 but they are entirely optional for the v0 format.

joelanford:
> I touch on those in https://github.com/operator-framework/rukpak/pull/149 but they are entirely optional for the v0 format.

That should be fine. We can have logic in the "read the version number" function that defaults to returning `0` if it can't find the version number in the expected place. I've got a branch that does this that passes all of the existing e2e's. I'll push it up.

timflannagan (approved):

PR #165, opened 2022-03-23
plain: Infer the ImagePullPolicy from the Bundle's spec.Image tag

timflannagan:
The default image pull policy is IfNotPresent for a container image unless the container image tag is set to 'latest' or the tag is unset (e.g. ""), so we should infer the pull policy from the user-provided container image.


joelanford (approved):
/lgtm

PR #166, opened 2022-03-24
Makefile: Support overriding the container runtime

timflannagan:
Update the Makefile and add a variable for the container runtime that defaults to docker. This allows users to override the container runtime used when enacting any of the container-related targets, e.g. `make e2e CONTAINER_RUNTIME=podman`.


akihikokuroda:
It works with nerdctl, too.  Thanks!

exdx (approved):

joelanford (approved):

PR #170, opened 2022-03-25
Introduce a CODEOWNERS file

timflannagan:
Introduce a root CODEOWNERS file and remove the root OWNERS file as we're not using tide in this repository.


joelanford (approved):

CODEOWNERS
```diff
@@ -0,0 +1 @@
+* @operator-framework/rukpak-maintainers
```
exdx, line 1: Does this prohibit those from outside the operator-framework org to become approvers?

exdx: Seems like this is not an issue.

timflannagan: It's not super clear reading from https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners but I think this is fine for now until we have an explicit need to have someone listed as a maintainer that's not also in the o-f organization.

PR #172, opened 2022-03-25
WIP: Proof of concept: Define a base metadata bundle.yaml file for all bundles

joelanford:
Propose a spec (for now just via code) for bundle metadata

Add machinery to parse the bundle version from this file in the
plain provisioner with special handling to assume bundles missing this
file are plain v0 bundles (at least from the perspective of the plain
provisioner).

Lastly, add specific handling for supported and unsupported bundle
formats in the plain provisioner, and add a test to ensure the plain
provisioner correctly handles unsupported bundles.

---

We could also continue using a big chunk of the existing [Manifest Bundle spec](https://github.com/kubernetes/enhancements/tree/master/keps/sig-cluster-lifecycle/addons/2494-Manifest-Bundle), but I think there are some aspects of it that need to be tightened up and improved.

Good discussion here: https://kubernetes.slack.com/archives/C0181L6JYQ2/p1648229357116729


cmd/unpack/main.go
```diff
@@ -6,16 +6,26 @@ import (
...
 	"github.com/spf13/cobra"
+
+	"github.com/operator-framework/rukpak/internal/version"
 )
 
 func main() {
 	var bundleDir string
 	var rukpakVersion bool
 
+	skipRootPaths := []string{
```
timflannagan, line 20: nit: Make this a set for quick lookup?

internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -222,8 +221,8 @@ func (r *BundleReconciler) ensureUnpackPod(ctx context.Context, bundle *rukpakv1
 		if bundle.Spec.Source.Image != nil {
 			pod.Spec.Containers[0].Name = bundleUnpackContainerName
 			pod.Spec.Containers[0].Image = bundle.Spec.Source.Image.Ref
-			pod.Spec.Containers[0].Command = []string{"/util/unpack", "--bundle-dir", "/manifests"}
-			pod.Spec.Containers[0].VolumeMounts = []corev1.VolumeMount{{Name: "util", MountPath: "/util"}}
+			pod.Spec.Containers[0].Command = []string{"/bin/unpack", "--bundle-dir", "/"}
+			pod.Spec.Containers[0].VolumeMounts = []corev1.VolumeMount{{Name: "util", MountPath: "/bin"}}
```
timflannagan, line 225: :+1:

test/e2e/plain_provisioner_test.go
```diff
@@ -247,6 +247,109 @@ var _ = Describe("plain provisioner bundle", func() {
 			}).Should(BeTrue())
 		})
 	})
+
+	When("a bundle containing no manifests is created", func() {
```
timflannagan, line 251: This seems useful outside of the scope of this PR?

joelanford: yeah +1 i'll make a separate PR for this since we already have code that checks for this.

timflannagan: The addition of that skipRootPaths array also seems like it's applicable to the unpacker outside of these changes?

joelanford: In fact, a good chunk of this is applicable without bringing metadata stuff into the equation.

test/e2e/plain_provisioner_test.go
```diff
@@ -247,6 +247,109 @@ var _ = Describe("plain provisioner bundle", func() {
...
+				if err != nil {
+					return nil, err
+				}
+				return meta.FindStatusCondition(bundle.Status.Conditions, rukpakv1alpha1.TypeUnpacked), nil
+			}).Should(And(
+				Not(BeNil()),
+				WithTransform(func(c *metav1.Condition) string { return c.Type }, Equal(rukpakv1alpha1.TypeUnpacked)),
+				WithTransform(func(c *metav1.Condition) metav1.ConditionStatus { return c.Status }, Equal(metav1.ConditionFalse)),
+				WithTransform(func(c *metav1.Condition) string { return c.Reason }, Equal(rukpakv1alpha1.ReasonUnpackFailed)),
+				WithTransform(func(c *metav1.Condition) string { return c.Message }, ContainSubstring(`open manifests: file does not exist`)),
```
exdx, line 300: Should we have this error string be a bit more verbose like originally? For example `invalid bundle: found zero files: open manifests: file does not exist`?

exdx (commented):
This is really great. I'm wondering, if we do want to check this code in, if we want to add the optional metadata information to the plain bundle spec while it's still in code review. That way we wouldn't have to update it after the fact (if we are sure we want metadata, which probably we do just based on things like versioning)

test/e2e/plain_provisioner_test.go
joelanford: Alright broke a good chunk out into #174

test/e2e/plain_provisioner_test.go
joelanford: Check out the TODO comment for the `It`. Our unpacker prevents us from getting to the code path where we successfully read the manifests directory, but find no manifests within.

internal/bundle/plain/v0/bundle.go
```diff
@@ -0,0 +1,78 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	apimacherrors "k8s.io/apimachinery/pkg/util/errors"
+	apimachyaml "k8s.io/apimachinery/pkg/util/yaml"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const ManifestsPath = "manifests"
+
+var GVK = schema.GroupVersionKind{Group: "bundles.rukpak.io", Version: "v0", Kind: "Plain"}
```
exdx, line 20: I like this structure of optional metadata file which embeds a typemeta. I can add this to the plain bundle spec as an optional source of metadata. Would we also want to manifests location field to the metadata, or are we still mandating that the manifests are under `/manifests`. I think we had added that because we didn't have a source of metadata defined -- if we do, maybe we could relax that assumption and allow the manifests to be in any specified directory?

internal/bundle/plain/v0/bundle.go
```diff
@@ -0,0 +1,78 @@
...
+
+func getObjects(bundleFS fs.FS) ([]client.Object, error) {
+	var objects []client.Object
+	entries, err := fs.ReadDir(bundleFS, ManifestsPath)
+	if err != nil {
+		return nil, err
+	}
+	for _, e := range entries {
+		if e.IsDir() {
+			continue
```
exdx, line 57: Based on the code, it looks like it's expected for the bundle manifests to live in `/manifests` and include all manifests at the top-level (no nesting). I can update the spec to specify this more explicitly.

internal/bundle/bundle.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"errors"
+	"fmt"
+	"io/fs"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/yaml"
+)
+
+const MetadataFile = "bundle.yaml"
```
exdx, line 12: This may be not an issue, but if users define their Bundle via a bundle.yaml file somewhere in their source of content it may collide with the metadata file name? Does having something like `metadata.yaml` or `bundle-metadata.yaml` prevent any sorts of collisions in the future?

internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -266,16 +265,29 @@ func (r *BundleReconciler) handleCompletedPod(ctx context.Context, u *updater.Up
...
 	}
 
-	objects, err := getObjects(bundleFS)
+	bundleMetadata, err := bundle.MetadataFromFS(bundleFS)
 	if err != nil {
-		return updateStatusUnpackFailing(u, fmt.Errorf("get objects from bundle manifests: %w", err))
+		if !errors.Is(err, bundle.ErrBundleFileNotFound) {
+			return updateStatusUnpackFailing(u, fmt.Errorf("read bundle metadata from filesystem: %w", err))
+		}
+		// If we didn't find the bundle.yaml file, assume it's a plain+v0 bundle
```
exdx, line 273: Might be helpful to emit a log here stating that an assumption was made regarding the bundle schema?

internal/bundle/plain/v0/bundle.go
joelanford: If there's a _really_ good use case for allowing the metadata to specify the location of the manifests, I think its something we could consider. But I feel like we haven't seen a use case for that with OLM's registry+v1 bundles (someone correct me if I'm wrong), so I think we should leave the `/manifests` directory location fixed for now.

At any point in the future we could add that to the metadata without a breaking change because we could say that `/manifests` is the default if it isn't defined. If we allow the manifest dir to be specified, there's no going back.

internal/bundle/bundle.go
joelanford: IMO, `bundle.yaml` seems fine. `metadata.yaml` seems more likely to conflict, and `bundle-metadata.yaml` feels a little forced.

If any projects _do_ end up needing a separate unrelated `bundle.yaml` in their project, the advice I'd give would be to put the bundle in a subdirectory. And I think that would be my advice anyway.

How often do you see a project that is packaged via helm intermingle their chart directory in the root of the project? In pretty much every project I've seen, the helm chart is in a subdirectory. Same concept here, IMO.

joelanford:
This is no longer relevant. Closing.

PR #173, opened 2022-03-25
WIP: Add kustomize patches for updating the Bundle CRD with oneOf semantics

exdx:
This exploratory PR introduces a kustomize dependency and uses the kustomize tooling to attempt to add/update fields that enforce a union type on the Bundle CRD. After the manifests have been created with controller-gen they are patched via kustomize with a kustomization file that is checked in. 

To see what the goal for the CRD schema should look like, see [an example from the metering CRDs](https://github.com/kube-reporting/metering-operator/blob/0d7ecfb5f9ca4603990bb264eae4d998da0e1060/manifests/deploy/upstream/metering-ansible-operator/meteringconfig.crd.yaml#L138). 

Note: It looks like adding the kustomization.yaml file in the same directory as the CRD manifests makes controller-gen to fail, so they may need to be moved to a base directory in order for this to work.


joelanford:
Here's the canonical kubebuilder example of a kustomize directory for CRDs: https://github.com/kubernetes-sigs/kubebuilder/tree/master/testdata/project-v3/config/crd

Makefile
```diff
@@ -52,11 +52,16 @@ tidy: ## Update dependencies
...
-generate: controller-gen ## Generate code and manifests
+generate: c-gen patch
+
+c-gen: controller-gen ## Generate code and manifests
 	$(Q)$(CONTROLLER_GEN) crd:crdVersions=v1 output:crd:dir=./manifests paths=./api/...
 	$(Q)$(CONTROLLER_GEN) schemapatch:manifests=./manifests output:dir=./manifests paths=./api/...
 	$(Q)$(CONTROLLER_GEN) object:headerFile=./hack/boilerplate.go.txt paths=./api/...
 
+patch: kustomize ## Apply kustomize patches
+	$(Q)$(KUSTOMIZE) build -o /manifests/core.rukpak.io_bundles.yaml ./manifests
```
joelanford, line 63: Perhaps instead of committing the final manifest output, we should add kustomization targets and use kustomize in a just-in-time mode? When we're actually applying manifests to the cluster? For example, `make install-crds` could call:
```
kubectl apply -k kustomize/crd
```
(assuming we put our CRD kustomization in the kustomize/crd directory)

This approach could also mean we don't need to carry the extra kustomize binary dependency (though it does raise questions about the version of kubectl that's used)

Otherwise, we'll have to make sure things stay in sync.

exdx:
Superseded by #217

PR #174, opened 2022-03-25
make unpack ignore irrelevant root directories

joelanford:
With this change, we can now unpack the image at the root of the image
filesystem, making unpack capable of handling future bundle formats that
may include other files and directories outside of /manifests

This PR also adds a test to ensure that we correctly fail bundles that
contain no manifests


joelanford:
This PR breaks out some of what's in #172 that's actually relevant to our current codebase regardless of the metadata conversation.

timflannagan (approved):

PR #176, opened 2022-03-26
unpack: encode bundle content in JSON as tar.gz

### LINKED ISSUE #175: update the unpacker to correctly handle empty directories.
Right now it only outputs a map of filenames with their contents,

so empty directories don't show up in the FS read by the provisioner.

Once we make the unpacker support empty dirs, add another test to

handle that case.

https://github.com/operator-framework/rukpak/blob/96d7c5b53a15c0a295a041a37f5b676afa4a030e/test/e2e/plain_provisioner_test.go#L282

```golang

			}).Should(BeTrue())
		})
	})

	When("a bundle containing no manifests is created", func() {
		var (
			bundle *rukpakv1alpha1.Bundle
			ctx    context.Context
		)
		BeforeEach(func() {
			ctx = context.Background()

			By("creating the testing Bundle resource")
			bundle = &rukpakv1alpha1.Bundle{
				ObjectMeta: metav1.ObjectMeta{
					GenerateName: "olm-crds-unsupported",
				},
				Spec: rukpakv1alpha1.BundleSpec{
					ProvisionerClassName: "core.rukpak.io/plain",
					Source: rukpakv1alpha1.BundleSource{
						Type: rukpakv1alpha1.SourceTypeImage,
						Image: &rukpakv1alpha1.ImageSource{
							Ref: "testdata/bundles/plain-v0:empty",
						},
					},
				},
			}
			err := c.Create(ctx, bundle)
			Expect(err).To(BeNil())
		})
		AfterEach(func() {
			By("deleting the testing Bundle resource")
			err := c.Delete(ctx, bundle)
			Expect(err).To(BeNil())
		})
		// TODO: update the unpacker to correctly handle empty directories.
		//   Right now it only outputs a map of filenames with their contents,
		//   so empty directories don't show up in the FS read by the provisioner.
		//   Once we make the unpacker support empty dirs, add another test to
		//   handle that case.
		It("reports an unpack error when the manifests directory is missing", func() {
			By("waiting for the bundle to report back that state")
			Eventually(func() (*metav1.Condition, error) {
				err := c.Get(ctx, client.ObjectKeyFromObject(bundle), bundle)
				if err != nil {
					return nil, err
				}
				return meta.FindStatusCondition(bundle.Status.Conditions, rukpakv1alpha1.TypeUnpacked), nil
			}).Should(And(
				Not(BeNil()),
				WithTransform(func(c *metav1.Condition) string { return c.Type }, Equal(rukpakv1alpha1.TypeUnpacked)),
				WithTransform(func(c *metav1.Condition) string { return c.Reason }, Equal(rukpakv1alpha1.ReasonUnpackFailed)),
				WithTransform(func(c *metav1.Condition) string { return c.Message }, ContainSubstring(`open manifests: file does not exist`)),
			))
		})
	})
})

var _ = Describe("plain provisioner bundleinstance", func() {

```

joelanford:
This commit enables the unpack binary to correctly handle empty
directories, and to encode other filesystem metadata (e.g. file mode).

Closes #175 

### DIFF
--- a/cmd/unpack/main.go
+++ b/cmd/unpack/main.go
@@ -1,8 +1,12 @@
 package main
 
 import (
+	"archive/tar"
+	"bytes"
+	"compress/gzip"
 	"encoding/json"
 	"fmt"
+	"io"
 	"io/fs"
 	"log"
 	"os"
@@ -42,10 +46,16 @@ func main() {
 			}
 
 			bundleFS := os.DirFS(bundleDir)
-			bundleContents := map[string][]byte{}
+			buf := &bytes.Buffer{}
+			gzw := gzip.NewWriter(buf)
+			tw := tar.NewWriter(gzw)
 			if err := fs.WalkDir(bundleFS, ".", func(path string, d fs.DirEntry, err error) error {
 				if err != nil {
 					return err
+				}
+
+				if d.Type()&os.ModeSymlink != 0 {
+					return nil
 				}
 				if bundleDir == "/" {
 					// If bundleDir is the filesystem root, skip some known unrelated directories
@@ -54,22 +64,51 @@ func main() {
 						return filepath.SkipDir
 					}
 				}
+				info, err := d.Info()
+				if err != nil {
+					return fmt.Errorf("get file info for %q: %w", path, err)
+				}
+
+				h, err := tar.FileInfoHeader(info, "")
+				if err != nil {
+					return fmt.Errorf("build tar file info header for %q: %w", path, err)
+				}
+				h.Uid = 0
+				h.Gid = 0
+				h.Uname = ""
+				h.Gname = ""
+				h.Name = path
+
+				if err := tw.WriteHeader(h); err != nil {
+					return fmt.Errorf("write tar header for %q: %w", path, err)
+				}
 				if d.IsDir() {
 					return nil
 				}
-				data, err := fs.ReadFile(bundleFS, path)
+				f, err := bundleFS.Open(path)
 				if err != nil {
-					return fmt.Errorf("read file %q: %w", path, err)
+					return fmt.Errorf("open file %q: %w", path, err)
 				}
-				bundleContents[path] = data
+				if _, err := io.Copy(tw, f); err != nil {
+					return fmt.Errorf("write tar data for %q: %w", path, err)
+				}
 				return nil
 			}); err != nil {
-				log.Fatalf("walk bundle dir %q: %v", bundleDir, err)
+				log.Fatalf("generate tar.gz for bundle dir %q: %v", bundleDir, err)
+			}
+			if err := tw.Close(); err != nil {
+				log.Fatal(err)
+			}
+			if err := gzw.Close(); err != nil {
+				log.Fatal(err)
 			}
 
-			encoder := json.NewEncoder(os.Stdout)
-			if err := encoder.Encode(bundleContents); err != nil {
-				log.Fatal(err)
+			bundleMap := map[string]interface{}{
+				"content": buf.Bytes(),
+			}
+			enc := json.NewEncoder(os.Stdout)
+			if err := enc.Encode(bundleMap); err != nil {
+				log.Fatalf("encode bundle map as JSON: %v", err)
 			}
 			return nil
 		},
--- a/internal/provisioner/plain/controllers/bundle_controller.go
+++ b/internal/provisioner/plain/controllers/bundle_controller.go
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	"bytes"
+	"compress/gzip"
 	"context"
 	"encoding/json"
 	"errors"
@@ -26,8 +27,8 @@ import (
 	"io/fs"
 	"path/filepath"
 	"strings"
-	"testing/fstest"
 
+	"github.com/nlepage/go-tarfs"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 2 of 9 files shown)


cmd/unpack/main.go
```diff
@@ -42,35 +45,69 @@ func main() {
...
+				h.Uid = 0
+				h.Gid = 0
+				h.Uname = ""
+				h.Gname = ""
+				h.Name = path
+
+				if err := tw.WriteHeader(h); err != nil {
+					return fmt.Errorf("write tar header for %q: %w", path, err)
+				}
+				if !d.IsDir() {
```
timflannagan, line 84: nit: flip this conditional check and return early?

cmd/unpack/main.go
```diff
@@ -42,35 +45,69 @@ func main() {
...
-			bundleContents := map[string][]byte{}
+			buf := &bytes.Buffer{}
+			gzw := gzip.NewWriter(buf)
+			tw := tar.NewWriter(gzw)
 			if err := fs.WalkDir(bundleFS, ".", func(path string, d fs.DirEntry, err error) error {
 				if err != nil {
 					return err
 				}
+
+				if d.Type()&os.ModeSymlink != 0 {
```
timflannagan, line 56: :+1:

internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -26,8 +27,8 @@ import (
 	"io/fs"
 	"path/filepath"
 	"strings"
-	"testing/fstest"
 
+	"github.com/nlepage/go-tarfs"
```
timflannagan, line 31: Hmm, we're probably way too early in the project lifecycle to care about this but I'll mention it anyways: but do we need to introduce a new direct dependency here?

test/e2e/plain_provisioner_test.go
```diff
@@ -297,7 +292,56 @@ var _ = Describe("plain provisioner bundle", func() {
...
+		BeforeEach(func() {
+			ctx = context.Background()
+
+			By("creating the testing Bundle resource")
+			bundle = &rukpakv1alpha1.Bundle{
+				ObjectMeta: metav1.ObjectMeta{
+					GenerateName: "olm-crds-unsupported",
+				},
+				Spec: rukpakv1alpha1.BundleSpec{
+					ProvisionerClassName: "core.rukpak.io/plain",
```
timflannagan, line 314: nit: use the package-level `plainProvisionerID` constant instead of hardcoding?

testdata/bundles/plain-v0/no-manifests/.dockerignore
```diff
@@ -0,0 +1 @@
+manifests/.gitignore
```
timflannagan, line 1: nit: github complaining about newline here.

test/e2e/plain_provisioner_test.go
```diff
@@ -297,7 +292,56 @@ var _ = Describe("plain provisioner bundle", func() {
 				WithTransform(func(c *metav1.Condition) string { return c.Type }, Equal(rukpakv1alpha1.TypeUnpacked)),
 				WithTransform(func(c *metav1.Condition) metav1.ConditionStatus { return c.Status }, Equal(metav1.ConditionFalse)),
 				WithTransform(func(c *metav1.Condition) string { return c.Reason }, Equal(rukpakv1alpha1.ReasonUnpackFailed)),
-				WithTransform(func(c *metav1.Condition) string { return c.Message }, ContainSubstring(`open manifests: file does not exist`)),
+				WithTransform(func(c *metav1.Condition) string { return c.Message }, Equal(`get objects from bundle manifests: readdir manifests: file does not exist`)),
```
timflannagan, line 295: nit: this check of check still feels fairly brittle?

cmd/unpack/main.go
```diff
@@ -42,35 +45,69 @@ func main() {
...
+				log.Fatal(err)
+			}
+			if err := gzw.Close(); err != nil {
 				log.Fatal(err)
 			}
+
+			bundleMap := map[string]interface{}{
+				"content": buf.Bytes(),
+			}
+			enc := json.NewEncoder(os.Stdout)
```
timflannagan, line 107: This is probably a poor idea: should we add a flag that allows us to configure the file descriptor that's used to relay unpacked Bundle contents? Right now, we always default to stdout (e.g. the Pod logs) but I wonder if we'll ever need to write directly to a file?

timflannagan: Also: can we use io.Copy here?

timflannagan (approved):
I didn't see anything blocking from my perspective so I'll approve this ahead of time and let you decide whether you want to address any of my comments. Note: I didn't pull this down and just reviewed through the UI but you introduced an e2e test so I'm reasonable confident these changes work.

test/e2e/plain_provisioner_test.go
joelanford: I'll simplify back down to the equivalent `ContainSubstring` matcher for now to make it slightly less brittle.

cmd/unpack/main.go
joelanford: > should we add a flag that allows us to configure the file descriptor that's used to relay unpacked Bundle contents?

I figure we can add a flag for that if/when we need it?

> can we use io.Copy here?

Not sure I'm following this suggestion (though I did see that I could use `io.Copy` in the WalkDir function to write the file contents from a file handle rather than reading the file into memory and then writing that `[]byte` back out).

internal/provisioner/plain/controllers/bundle_controller.go
joelanford: Fair question. I _did_ do a little bit of due diligence. It's actively maintained and has lots of tests. It also looks complex enough that if we tried to re-implement it, we'd probably have some maintenance overhead to contend with.

The one bug that was filed was fixed by the maintainer in 4 days.

TL;DR: I personally feel alright about this one.

timflannagan: Thanks - that works for me. I don't think this is blocking anyways right now.

cmd/unpack/main.go
timflannagan: Yep sgtm. Nothing blocking here imo.

Issue #164, opened 2022-03-23
Generalize bundle digest status field for other content sources


exdx:
The Bundle status has a `status.digest` field which is used for reporting the SHA-256 hash, or the image ID, of the `plain+v0` bundle image that has content stored inside of it. This field is useful for immutability purposes and for making pivoting decisions. However, this field is not as useful for Bundle content sources which are not based on container images. This field could potentially be used for other sources (for example, a sha256 hash of some git repository file) or it may need to be renamed to something more generic to be useful for other content sources. This issue tracks how to handle the `status.digest` field when the content source is not a container image with a relevant digest.

PR #179, opened 2022-03-28
feat: add support for git-based bundles

### LINKED ISSUE #117: Add support for sourcing a git repository for Bundle content
The Bundle API right now only supports unpacking a single source type right now: a remote container image, which is configured using the spec.Image field in the Bundle CRD. We'll need to onboard new source types, which can be enabled on a per-provisioner-basis, in the Bundle's spec. The e2e strawman outlined the potential for a series of source references that can be configured using a prefix-based approach:

```yaml
kind: Bundle
metadata: 
  name: plumbus-operator.v0.9.3
spec:
  class: deppy.resolveset
  refs:
  - file://content
  volumeMounts:
  - mountPath: /content
    configMap:
      name: local 
      namespace: plumbus
```

The goal here would be able to expose and implement the ability to point to a git repository as a valid Bundle source. In the context of the plain+v0 provisioner, users could source Bundle contents that live in a repositories root manifests/ directory, and that directory contains a series of static YAML manifests, and the provisioner is responsible for sourcing and unpacking those contents on-cluster.

exdx:
Closes #117 

Adds introductory support for creating bundles backed by git repositories.


Makefile
```diff
@@ -75,8 +75,9 @@ UNIT_TEST_DIRS=$(shell go list ./... | grep -v /test/)
 test-unit: setup-envtest ## Run the unit tests
 	eval $$($(SETUP_ENVTEST) use -p env $(ENVTEST_VERSION)) && go test -count=1 -short $(UNIT_TEST_DIRS)
 
+SPECIFIC_E2E_TEST := $(if $(TEST),-focus "$(TEST)")
```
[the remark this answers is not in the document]
timflannagan: Oh, disregard, I forgot there's a -focus flag. Can we maybe rename this variable to TEST so it emulates OLM's implementation?

api/v1alpha1/bundle_types.go
```diff
@@ -51,13 +51,27 @@ type BundleSource struct {
...
+	// Git is the git repository that backs the content of this Bundle.
+	Git *GitSource `json:"git,omitempty"`
 }
 
 type ImageSource struct {
 	// Ref contains the reference to a container image containing Bundle contents.
 	Ref string `json:"ref"`
 }
 
+type GitSource struct {
```
timflannagan, line 63: Follow-up: add the comments + kubebuilder markers for these fields?

api/v1alpha1/bundle_types.go
```diff
@@ -51,13 +51,27 @@ type BundleSource struct {
...
+	Repository string `json:"repository"`
+	Directory  string `json:"directory,omitempty"`
+	Ref        GitRef `json:"ref"`
```
timflannagan, line 66: Does this need to be a pointer to a GitRef?

internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -206,26 +207,63 @@ func (r *BundleReconciler) ensureUnpackPod(ctx context.Context, bundle *rukpakv1
...
+		if bundle.Spec.Source.Type == "git" {
+			pod.Spec.InitContainers[1].Name = "clone-repository"
+			pod.Spec.InitContainers[1].Image = "bitnami/git:latest"
+			pod.Spec.InitContainers[1].ImagePullPolicy = corev1.PullIfNotPresent
+			checkoutCmd := git.NewCheckoutCmd(*bundle.Spec.Source.Git)
+			c, err := checkoutCmd.String()
+			if err != nil {
+				return err
+			}
+			pod.Spec.InitContainers[1].Command = []string{"/bin/bash", "-c", c}
```
timflannagan, line 244:
```suggestion
			pod.Spec.InitContainers[1].Command = []string{"/bin/bash", "-c", c.String()}
```

internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -206,26 +207,63 @@ func (r *BundleReconciler) ensureUnpackPod(ctx context.Context, bundle *rukpakv1
...
 		pod.Spec.InitContainers[0].VolumeMounts = []corev1.VolumeMount{{Name: "util", MountPath: "/util"}}
 
+		// Note: initContainer so we can ensure the repository has been cloned
+		// at the bundle.Spec.Source.Git.Ref before we unpack the Bundle contents
+		// that are stored in the repository.
+		if bundle.Spec.Source.Type == "git" {
+			pod.Spec.InitContainers[1].Name = "clone-repository"
+			pod.Spec.InitContainers[1].Image = "bitnami/git:latest"
+			pod.Spec.InitContainers[1].ImagePullPolicy = corev1.PullIfNotPresent
+			checkoutCmd := git.NewCheckoutCmd(*bundle.Spec.Source.Git)
```
timflannagan, line 239:
```suggestion
			cmd, err := git.NewCheckoutCmd(*bundle.Spec.Source.Git)
			if err != nil { return err }
```

internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -206,26 +207,63 @@ func (r *BundleReconciler) ensureUnpackPod(ctx context.Context, bundle *rukpakv1
...
 		if bundle.Spec.Source.Image != nil {
 			pod.Spec.Containers[0].Name = bundleUnpackContainerName
 			pod.Spec.Containers[0].Image = bundle.Spec.Source.Image.Ref
 			pod.Spec.Containers[0].Command = []string{"/bin/unpack", "--bundle-dir", "/"}
 			pod.Spec.Containers[0].VolumeMounts = []corev1.VolumeMount{{Name: "util", MountPath: "/bin"}}
 		}
+
+		if bundle.Spec.Source.Type == "git" {
+			pod.Spec.Containers[0].Name = bundleUnpackContainerName
+			pod.Spec.Containers[0].Image = "bitnami/git:latest"
```
timflannagan, line 261: follow-up: make this image configurable?

internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -206,26 +207,63 @@ func (r *BundleReconciler) ensureUnpackPod(ctx context.Context, bundle *rukpakv1
...
+			pod.Spec.Containers[0].ImagePullPolicy = corev1.PullAlways
```
timflannagan, line 262: IfNotPresent?

internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -262,11 +300,13 @@ func (r *BundleReconciler) handleCompletedPod(ctx context.Context, u *updater.Up
...
 	}
 
+	// TODO: this doesn't apply to all bundle source types
 	bundleImageDigest, err := r.getBundleImageDigest(pod)
 	if err != nil {
 		return updateStatusUnpackFailing(u, fmt.Errorf("get bundle image digest: %w", err))
 	}
-
+	// TODO: what happens when len(objects) == 0? how to avoid silently storing zero manifests
+	// despite them being populated correctly in the Pod logs?
```
timflannagan, line 309: Rebase against the latest main branch?

exdx: Hmm I thought I did rebase...checking again

api/v1alpha1/bundle_types.go
joelanford: Big +1, really good GoDoc is key on these APIs because it drives the content of `kubectl explain`

internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -206,26 +207,63 @@ func (r *BundleReconciler) ensureUnpackPod(ctx context.Context, bundle *rukpakv1
...
 			{Name: "util", VolumeSource: corev1.VolumeSource{EmptyDir: &corev1.EmptyDirVolumeSource{}}},
+			{Name: "manifests", VolumeSource: corev1.VolumeSource{EmptyDir: &corev1.EmptyDirVolumeSource{}}},
 		}
 		pod.Spec.RestartPolicy = corev1.RestartPolicyNever
-		if len(pod.Spec.InitContainers) != 1 {
-			pod.Spec.InitContainers = make([]corev1.Container, 1)
+		if bundle.Spec.Source.Type == "git" {
+			// Note: use host networking (hack) so we can resolve https://github.com references in the
+			// Bundle's spec.Source.Git.Repository. Maybe we should revisit having the user inline https://github.com?
+			pod.Spec.HostNetwork = true
```
joelanford, line 216: I'm surprised this is necessary. What error are you running into? By default, the dnsPolicy for pods is `ClusterFirst`, which seems fine? 

https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy

internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -206,26 +207,63 @@ func (r *BundleReconciler) ensureUnpackPod(ctx context.Context, bundle *rukpakv1
...
 		}
+
+		if bundle.Spec.Source.Type == "git" {
+			pod.Spec.Containers[0].Name = bundleUnpackContainerName
+			pod.Spec.Containers[0].Image = "bitnami/git:latest"
+			pod.Spec.Containers[0].ImagePullPolicy = corev1.PullAlways
+			pod.Spec.Containers[0].Command = []string{"/util/unpack", "--bundle-dir", "/manifests"}
+			pod.Spec.Containers[0].VolumeMounts = []corev1.VolumeMount{{Name: "util", MountPath: "/util"}, {Name: "manifests", MountPath: "/manifests"}}
+			return nil
+		}
```
joelanford, line 266: Should we extract the `image` pod and `git` pod into fully separate functions with just a switch statement somewhere?

I feel like that might be easier to maintain over time even if there is some duplication between them.

Thoughts?

timflannagan: Yup - agreed, this needs to be done. I was talking to dan in the #rukpak-dev about whether we should attempt a refactoring of this pod unpacking method in this PR, or punt that kind of work as a follow-up.

exdx: It totally makes sense, I was hacking around with breaking the code out and deduplicating it in https://github.com/exdx/rukpak/commit/4a2d74e4bd8f353926d3f33881c0505be9a5f556 but I'm not sure if we need something like that or it can be a follow-up. Simply breaking out into two separate functions is easier and makes more sense.

internal/provisioner/plain/controllers/bundle_controller.go
joelanford: Any specific reason we chose this one? I also see https://hub.docker.com/r/alpine/git, which looks quite a bit smaller and it supports multi-arch.

Looks like I wouldn't be able to run bitnami/git on my M1 mac without emulation since it is only a linux/amd64 machine.

Also, should we pin to a version (or version stream)?

internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -262,11 +300,13 @@ func (r *BundleReconciler) handleCompletedPod(ctx context.Context, u *updater.Up
 		return updateStatusUnpackFailing(u, fmt.Errorf("get bundle contents: %w", err))
 	}
 
+	// TODO: this doesn't apply to all bundle source types
```
joelanford, line 303: Nit: format this with a title-like thing on the first line so that the TODO issue creator makes a nicer GH issue for it when we merge?

internal/provisioner/plain/pkg/git/git.go
```diff
@@ -0,0 +1,78 @@
...
+	"errors"
+	"fmt"
+	"net/url"
+	"strings"
+
+	"github.com/operator-framework/rukpak/api/v1alpha1"
+)
+
+const (
+	defaultDirectory = "./manifests"
```
joelanford, line 13: This seems like it should refer to another constant that exists somewhere in the code that defines the `plain+v1` spec. If that constant doesn't exist yet, perhaps we should add a TODO comment here?

internal/provisioner/plain/pkg/git/git.go
```diff
@@ -0,0 +1,78 @@
...
+	defaultBranch    = "main"
```
joelanford, line 14: Is it possible to ask the git repo for it's default branch rather than hardcoding it?

internal/provisioner/plain/pkg/git/git.go
```diff
@@ -0,0 +1,78 @@
...
+)
+
+type CheckoutCmd struct {
+	v1alpha1.GitSource
+}
```
joelanford, line 19: Did you consider a single exported function (e.g. `CloneCommandFor(v1alpha1.GitSource) (string, error)` that combines the validation and command building into a single entrypoint with no need for an extra struct?

internal/provisioner/plain/pkg/git/git.go
```diff
@@ -0,0 +1,78 @@
+package git
```
joelanford, line 1: One issue that the location of this implementation raises is that it is coupled fairly closely to the plain provisioner's use of an unpack pod, which on the surface makes it less suitable for use in other provisioners.

This reminds me that we may want to create a Source interface that hides the details of the source from the provisioner so that other provisioners could easily reuse the implementation.

internal/provisioner/plain/pkg/git/git.go
```diff
@@ -0,0 +1,78 @@
...
+
+	if branch == "" {
+		branch = defaultBranch
+	}
+
+	if commit != "" {
+		checkoutCommand = fmt.Sprintf("git clone %s && cd %s && git checkout %s && cp -r %s/* /manifests",
+			repository, repositoryName, commit, directory)
+		return checkoutCommand, nil
+	}
```
joelanford, line 50: It seems like this code would let me specify `branch`, `commit`, and `tag`, and then would silently ignore `branch` and `tag`.

Can we validate that exactly one of `branch` or `commit` or `tag` is set?

internal/provisioner/plain/controllers/bundle_controller.go
joelanford: I think we should also consider a pure in-process Go implementation to avoid having to worry about this extra image. But I think we can defer that until/after we get a `Source` interface in place.

internal/provisioner/plain/pkg/git/git.go
joelanford: IMO, we should do this as part of reconcile, and then also layer this same validation into the CRD schema or a webhook.

api/v1alpha1/bundle_types.go
joelanford: For `Ref`, we should definitely include a statement in the GoDoc about what's valid. For example:
```go
// Ref configures the git source to clone a specific branch, tag, or commit 
// from the specified repo. Ref is required, and exactly one field within Ref
// error
```

internal/provisioner/plain/pkg/git/git.go
```diff
@@ -0,0 +1,78 @@
...
+	checkoutCommand = fmt.Sprintf("git clone --depth 1 --branch %s %s && cd %s && git checkout %s && cp -r %s/* /manifests",
+		branch, repository, repositoryName, branch, directory)
+	return checkoutCommand, nil
+}
+
+func (c CheckoutCmd) ValidateRepo() (string, error) {
+	u, err := url.Parse(c.Repository)
+	if err != nil {
+		return "", fmt.Errorf("provided git repository url malformed: %w", err)
+	}
```
joelanford, line 67: Does this work for the repo syntax that looks like this: `git@github.com:operator-framework/combo.git`?

internal/provisioner/plain/pkg/git/git.go
```diff
@@ -0,0 +1,78 @@
...
+	if directory == "" {
+		directory = defaultDirectory
+	}
+
+	if branch == "" {
+		branch = defaultBranch
+	}
+
+	if commit != "" {
+		checkoutCommand = fmt.Sprintf("git clone %s && cd %s && git checkout %s && cp -r %s/* /manifests",
```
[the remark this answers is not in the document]
joelanford: How confident are we in extracting the `repositoryName`? Should we instead just create a static dir?

```bash
git clone <repoUrl> repo && cd repo && ...
```

internal/provisioner/plain/pkg/git/git.go
joelanford: I feel pretty strongly we should just hand the repo string off to git as is and let git return any error it finds.

Makefile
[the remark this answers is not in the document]
timflannagan: We could also deflate this kind of change to it's own PR as it seems valuable outside of these changes?

internal/provisioner/plain/controllers/bundle_controller.go
exdx: We can use the alpine git based image sure -- and I agree it should be configurable. 

Doing the git unpacking in process is also doable once we have the abstractions in place.

internal/provisioner/plain/controllers/bundle_controller.go
exdx: Looks like I did but the comment came through anyway...will remove

internal/provisioner/plain/pkg/git/git.go
exdx: I looked and it was a bit non-trivial. Requiring to specify a branch makes sense IMO.

internal/provisioner/plain/pkg/git/git.go
exdx: I think that approach also works, I had something like that at one point before moving to this struct-based implementation.

internal/provisioner/plain/pkg/git/git.go
exdx: Will add additional validation here, sure. 

Looking to add some union semantics to the git fields as well to help with validation at Bundle apply-time

internal/provisioner/plain/pkg/git/git.go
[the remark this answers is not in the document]
exdx: That seems a lot safer, was not aware of that option 👍

internal/provisioner/plain/pkg/git/git.go
exdx: This is largely to safely extract the name of the repository. If we have the ability to clone to a static repo name, we can remove this logic entirely.

api/v1alpha1/bundle_types.go
timflannagan: I think we're in the clear here - sorry for the noise.

api/v1alpha1/bundle_types.go
```diff
@@ -51,13 +51,27 @@ type BundleSource struct {
...
+}
+
+type GitRef struct {
```
timflannagan, line 69: follow-up: validate (either in code or schema validation?) that only a single branch/tag/commit has been specified?

internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -206,26 +207,63 @@ func (r *BundleReconciler) ensureUnpackPod(ctx context.Context, bundle *rukpakv1
...
 			pod.Spec.Containers[0].Image = bundle.Spec.Source.Image.Ref
 			pod.Spec.Containers[0].Command = []string{"/bin/unpack", "--bundle-dir", "/"}
 			pod.Spec.Containers[0].VolumeMounts = []corev1.VolumeMount{{Name: "util", MountPath: "/bin"}}
 		}
+
+		if bundle.Spec.Source.Type == "git" {
+			pod.Spec.Containers[0].Name = bundleUnpackContainerName
+			pod.Spec.Containers[0].Image = "bitnami/git:latest"
+			pod.Spec.Containers[0].ImagePullPolicy = corev1.PullAlways
+			pod.Spec.Containers[0].Command = []string{"/util/unpack", "--bundle-dir", "/manifests"}
```
timflannagan, line 263: See: #174 

```suggestion
			pod.Spec.Containers[0].Command = []string{"/util/unpack", "--bundle-dir", "/"}
```

internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -206,26 +207,63 @@ func (r *BundleReconciler) ensureUnpackPod(ctx context.Context, bundle *rukpakv1
...
+			pod.Spec.Containers[0].VolumeMounts = []corev1.VolumeMount{{Name: "util", MountPath: "/util"}, {Name: "manifests", MountPath: "/manifests"}}
```
timflannagan, line 264: Let's make this consistent with the container image volume mount naming (see #174)

```suggestion
			pod.Spec.Containers[0].VolumeMounts = []corev1.VolumeMount{{Name: "util", MountPath: "/bin"}, {Name: "manifests", MountPath: "/manifests"}}
```

internal/provisioner/plain/pkg/git/git.go
timflannagan: I think my only hesitation is that we can easily index something like a tag/commit/directory/etc. with a custom type, while reduce the coupling with an individual bundle type? It's also entirely possible that we don't have any need for this custom type implementation right now, and all we're interested in functionality centered around building up a formatted strings that's responsible for sourcing git bundle contents.

api/v1alpha1/bundle_types.go
exdx: I added some comments for each of the fields in the newly introduced git structs. They should be much more clear now. 

Regarding kubebuilder markers, I'm not sure which markers to add -- I know we wanted to enforce some union semantics but I don't think kubebuilder support those (since controller-gen does not). We have to manually modify the Bundle CRD manifest to get that functionality.

internal/provisioner/plain/controllers/bundle_controller.go
exdx: I broke the large function out into two smaller pieces. There is still some abstraction to be desired but I think it's good enough for now.

internal/provisioner/plain/controllers/bundle_controller.go
exdx: We actually have https://github.com/operator-framework/rukpak/issues/164 open for this so I decided to put a link to there instead

internal/provisioner/plain/pkg/git/git.go
exdx: According to the spec (which we just merged) we will default to ./manifests for bundle git repos. So I propose leaving this as-is for now. Unless I'm misunderstanding what's being asked in which case we can revisit.

internal/provisioner/plain/pkg/git/git.go
exdx: I decided to combine both implementations: return a (string, error) in the main codepath and use an internal type under the hood, which helps with building the correct command.

api/v1alpha1/bundle_types.go
exdx: Yes -- I have it specified in the code and would like to do it via schema validation as well in #173

internal/provisioner/plain/pkg/git/git.go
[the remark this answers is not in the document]
exdx: I'm not sure about the use of %q since it seems to add quotation marks where we want to pass unquoted strings in the clone command. We could trim the fields of any whitespace before building the command? Although I'm not sure if the api-server does that when it unmarshals and stores the Bundle to disk.

internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -206,27 +208,24 @@ func (r *BundleReconciler) ensureUnpackPod(ctx context.Context, bundle *rukpakv1
...
+			return nil
+		case rukpakv1alpha1.SourceTypeGit:
+			var err error
+			pod, err = r.bundleGitRepoPod(pod, *bundle.Spec.Source.Git)
+			if err != nil {
+				return err
+			}
+			return nil
+		default:
+			return errors.New("unsupported bundle source type")
```
timflannagan, line 227: nit: fmt.Errorf and print out the bundle source type that's unsupported?

api/v1alpha1/bundle_types.go
```diff
@@ -51,13 +52,41 @@ type BundleSource struct {
...
+	Directory string `json:"directory,omitempty"`
+	// Ref configures the git source to clone a specific branch, tag, or commit
+	// from the specified repo. Ref is required, and exactly one field within Ref
+	// is required. Setting more than one field or zero fields will result in an
+	// error.
+	Ref GitRef `json:"ref"`
+}
+
+type GitRef struct {
+	// Branch refers to the branch to checkout from the repository.
```
tylerslaton, line 79: > follow-up: validate (either in code or schema validation?) that only a single branch/tag/commit has been specified?

In a similar vein - should we also make sure that at least one is specified? As it currently sits it will pass if none of these are set - right?

internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -392,3 +393,67 @@ func (r *BundleReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	pod.Spec.Containers[0].Name = bundleUnpackContainerName
+	pod.Spec.Containers[0].Image = source.Ref
+	pod.Spec.Containers[0].Command = []string{"/bin/unpack", "--bundle-dir", "/"}
+	pod.Spec.Containers[0].VolumeMounts = []corev1.VolumeMount{{Name: "util", MountPath: "/bin"}}
+
+	return pod
+}
+
+func (r *BundleReconciler) bundleGitRepoPod(pod *corev1.Pod, source rukpakv1alpha1.GitSource) (*corev1.Pod, error) {
+	// Note: use host networking (hack) so we can resolve https://github.com references in the
```
tylerslaton, line 421: nit: Should we make this a todo comment so an issue gets created?

internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -392,3 +393,67 @@ func (r *BundleReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&corev1.ConfigMap{}).
 		Complete(r)
 }
+
+func (r *BundleReconciler) bundleImagePod(pod *corev1.Pod, source rukpakv1alpha1.ImageSource) *corev1.Pod {
+	if len(pod.Spec.InitContainers) != 1 {
+		pod.Spec.InitContainers = make([]corev1.Container, 1)
+	}
+
+	pod.Spec.InitContainers[0].Name = "install-unpacker"
```
tylerslaton, line 402: nit: Should we make a function that is specifically for setting up the `UnpackImage`? We call this logic twice.

test/e2e/plain_provisioner_test.go
```diff
@@ -345,6 +345,120 @@ var _ = Describe("plain provisioner bundle", func() {
...
+			By("eventually writing a non-empty list of unpacked objects to the status", func() {
+				for _, bundle := range bundles {
+					Eventually(func() bool {
+						if err := c.Get(ctx, client.ObjectKeyFromObject(bundle), bundle); err != nil {
+							return false
+						}
+						if bundle.Status.Info == nil {
+							return false
+						}
+						/*
```
tylerslaton, line 446: I like the idea of including what we're expecting a bundle too look like after being unpacked. Makes this much more readable.

internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -392,3 +393,67 @@ func (r *BundleReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	pod.Spec.InitContainers[0].Image = r.UnpackImage
+	pod.Spec.InitContainers[0].ImagePullPolicy = corev1.PullIfNotPresent
+	pod.Spec.InitContainers[0].Command = []string{"cp", "-Rv", "/unpack", "/util/unpack"}
+	pod.Spec.InitContainers[0].VolumeMounts = []corev1.VolumeMount{{Name: "util", MountPath: "/util"}}
+
+	// Note: initContainer so we can ensure the repository has been cloned
+	// at the bundle.Spec.Source.Git.Ref before we unpack the Bundle contents
+	// that are stored in the repository.
+	pod.Spec.InitContainers[1].Name = "clone-repository"
+	pod.Spec.InitContainers[1].Image = r.GitClientImage
```
tylerslaton, line 439: nit: I think that the `GitClientImage` could end up feeling like some hidden configuration given that it is a command line flag for a binary most people won't be running directly. Maybe some documentation in the code could clear that up? We'll be updating the docs in a later PR right?

internal/provisioner/plain/controllers/bundle_controller.go
exdx: I took it out and the e2e still passed so I assume we are fine without this setting.

tylerslaton (commented):
This is great work. Main question is around the user-experience of editing `GitClientImage` should they need. Even though it unlikely to be used, I'm curious if this is the end UX for it and if that makes sense to proceed with.

api/v1alpha1/bundle_types.go
exdx: That's a good point -- I do verify that in the code here: https://github.com/operator-framework/rukpak/pull/179/files#diff-fe812db0d942803e1f21fe81bd691d1c62d0a438ba3f35309bb26cb5339970f9R66 but we would want to enforce that at the schema level as well if possible. I'm not sure how well we can mix oneOf with an at least one idea but we should try. 

We also have the Bundle webhook as a good place to drop some of this validation in as well.

internal/provisioner/plain/controllers/bundle_controller.go
exdx: We ended up pulling this code out and the e2e still passed, so we are removing it altogether

internal/provisioner/plain/controllers/bundle_controller.go
exdx: Sure, good catch. No need to duplicate code.

internal/provisioner/plain/controllers/bundle_controller.go
exdx: That's true that this field is a bit hidden but it's largely an implementation detail that I'm not sure we want to leak in the user documentation. I feel like this field will be largely used when we start hitting rate-limiting issues from dockerhub and have to mirror the git image to quay. 

This git-pod based implementation is pretty heavily coupled to the plain provisioner so it's not clear this would have immediate value to other provisioners either. 

I can add a comment in the code to provide more context though 👍

api/v1alpha1/bundle_types.go
tylerslaton: I like the idea of the Bundle webhook handling that logic.

joelanford: IMO, anything that _can_ be expressed in CRD schema should go in CRD schema. There are features coming to kubernetes that will make it easier to include logic in CRD validations via CEL expressions, so once that lands it would be nice if we could get rid of whatever's left in our validation webhooks.

internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -392,3 +393,68 @@ func (r *BundleReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&corev1.ConfigMap{}).
 		Complete(r)
 }
+
+func (r *BundleReconciler) bundleImagePod(pod *corev1.Pod, source rukpakv1alpha1.ImageSource) *corev1.Pod {
```
timflannagan, line 397: Does this need to be a method on the Bundle controller?

internal/provisioner/plain/pkg/git/git.go
```diff
@@ -0,0 +1,83 @@
...
+		return "", err
+	}
+	return cmd.String(), nil
+}
+
+func (c *checkoutCmd) String() string {
+	var checkoutCommand string
+	var repository = c.Repository
+	var directory = c.Directory
+	var branch = c.Ref.Branch
```
timflannagan, line 32: Can ref be nil here?

test/e2e/plain_provisioner_test.go
```diff
@@ -345,6 +345,120 @@ var _ = Describe("plain provisioner bundle", func() {
...
+	When("Bundles are backed by a git repository", func() {
+		var (
+			bundles []*rukpakv1alpha1.Bundle
+			ctx     context.Context
+		)
+		BeforeEach(func() {
+			ctx = context.Background()
+
+			By("creating the git based Bundles")
+			bundles = []*rukpakv1alpha1.Bundle{
```
timflannagan, line 358: I think my preference would be to split out each of these Bundle resources into their own nested When containers, e.g. When("a Bundle contains a reference to a git repository tag"), and handle any of the requisite checks within those individual test containers. Thoughts?

internal/provisioner/plain/pkg/git/git.go
```diff
@@ -0,0 +1,83 @@
...
+			tag, repository, repositoryName, repositoryName, tag, directory)
+		return checkoutCommand
+	}
+
+	checkoutCommand = fmt.Sprintf("git clone --depth 1 --branch %s %s %s && cd %s && git checkout %s && cp -r %s/* /manifests",
+		branch, repository, repositoryName, repositoryName, branch, directory)
+	return checkoutCommand
+}
+
+func (c *checkoutCmd) Validate() error {
```
timflannagan, line 57: I think my preference would be to handle these kinds of validation checks in the CRD schema validation. Thoughts?

internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -392,3 +393,68 @@ func (r *BundleReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	}
+	pod.Spec.InitContainers[1].Command = []string{"/bin/sh", "-c", cmd}
+	pod.Spec.InitContainers[1].VolumeMounts = []corev1.VolumeMount{{Name: "util", MountPath: "/util"}, {Name: "manifests", MountPath: "/manifests"}}
+
+	if len(pod.Spec.Containers) != 1 {
+		pod.Spec.Containers = make([]corev1.Container, 1)
+	}
+
+	pod.Spec.Containers[0].Name = bundleUnpackContainerName
+	pod.Spec.Containers[0].Image = r.GitClientImage
```
timflannagan, line 443: Do we need to use the r.GitClientImage as the container image here? We should already have access to the /manifests directory as we populated that volume in an initContainer, and the r.UnpackImage should be smaller in size compared to any git container image we provide to this provisioner.

internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -392,3 +393,68 @@ func (r *BundleReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	pod.Spec.InitContainers[1].Name = "clone-repository"
+	// r.GitClientImage configures which git-based container image to use to clone the provided repository
+	// r.GitClientImage currently defaults to alpine/git:v2.32.0
+	pod.Spec.InitContainers[1].Image = r.GitClientImage
+	pod.Spec.InitContainers[1].ImagePullPolicy = corev1.PullIfNotPresent
+	cmd, err := git.CloneCommandFor(source)
+	if err != nil {
+		return nil, err
+	}
+	pod.Spec.InitContainers[1].Command = []string{"/bin/sh", "-c", cmd}
```
timflannagan, line 435: I realize this is one of the downsides when I had suggested we support configuring the git container image is that we have less control over images that contain multiple shells. That said, I _think_ /bin/sh should be fine for most container images?

internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -42,6 +42,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	rukpakv1alpha1 "github.com/operator-framework/rukpak/api/v1alpha1"
+	"github.com/operator-framework/rukpak/internal/provisioner/plain/pkg/git"
```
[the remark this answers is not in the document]
exdx: Ok -- we can move it out for the sake of consistency. We don't really have a `pkg` anywhere but maybe we can introduce one down the line.

internal/provisioner/plain/controllers/bundle_controller.go
exdx: we can make it standalone

internal/provisioner/plain/pkg/git/git.go
exdx: Ref is a non-pointer type so I think we are safe to reference it?

internal/provisioner/plain/controllers/bundle_controller.go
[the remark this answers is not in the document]
timflannagan: We could just move it to internal/git/git.go?

internal/provisioner/plain/controllers/bundle_controller.go
exdx: Yes, for portability, I think we can get away with /bin/sh as most distros recognize this or have a symlink to another shell

internal/provisioner/plain/pkg/git/git.go
exdx: I think since Ref is required on the API we can get away with this type of blind indexing, although I agree it looks a bit strange. We could alternatively update Ref to be a pointer type.

test/e2e/plain_provisioner_test.go
timflannagan: We can hold off on fixing this in this initial implementation but I'd like to see us refactor these e2e test cases in a follow-up.

api/v1alpha1/bundle_types.go
```diff
@@ -51,13 +52,41 @@ type BundleSource struct {
...
+	Git *GitSource `json:"git,omitempty"`
 }
 
 type ImageSource struct {
 	// Ref contains the reference to a container image containing Bundle contents.
 	Ref string `json:"ref"`
 }
 
+type GitSource struct {
+	// Repository is a URL link to the git repository containing the bundle.
```
akihikokuroda, line 65: nit: it may not be clear enough.  My suggestion is
`Repository is a HTTPS URL to clone the git repository containing the bundle.`

timflannagan: I can understand the perspective that this description may need to be revisited, but we could in theory add support for any git-based protocol (ssh/https/etc.) that's supported on their side.

internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -206,27 +208,24 @@ func (r *BundleReconciler) ensureUnpackPod(ctx context.Context, bundle *rukpakv1
...
-		}
-		if bundle.Spec.Source.Image != nil {
-			pod.Spec.Containers[0].Name = bundleUnpackContainerName
-			pod.Spec.Containers[0].Image = bundle.Spec.Source.Image.Ref
-			pod.Spec.Containers[0].Command = []string{"/bin/unpack", "--bundle-dir", "/"}
-			pod.Spec.Containers[0].VolumeMounts = []corev1.VolumeMount{{Name: "util", MountPath: "/bin"}}
+
+		switch bundle.Spec.Source.Type {
+		case rukpakv1alpha1.SourceTypeImage:
+			pod = bundleImagePod(pod, *bundle.Spec.Source.Image, r.UnpackImage)
```
akihikokuroda, line 217: It's possible that the `bundle.Spec.Source.Image` is not set.  It may be better that `bundleImagePod` validates the source and return error if it's not set.

akihikokuroda (commented):
Looks good. I have some nit comments.  Thanks!

internal/provisioner/plain/controllers/bundle_controller.go
timflannagan: I think this is something I'd expect to be enforced in CRD schema validation (which doesn't exist right now but we're targeting that work in a follow-up), but maybe we should double dip into validation in code too?

api/v1alpha1/bundle_types.go
akihikokuroda: I see.  I see only the HTTPS URLs in the unit tests so I guessed that that's the only supported format.

internal/provisioner/plain/controllers/bundle_controller.go
akihikokuroda: The `bundleGitRepoPod` does validation but `bundleImagePod` doesn't.  I see a little inconsistency.

api/v1alpha1/bundle_types.go
exdx: Currently yes -- we only support publicly available git repositories in this iteration. But we would like to support additional formats as well once we have some form of secret-sharing/authentiation in place.

internal/provisioner/plain/controllers/bundle_controller.go
exdx: That's true, I would expect this validation at the schema level but we can add some validation here as well. I think it should be a follow-up since it's not directly related to this PR.

tylerslaton (approved):
I'm going to go ahead and send out the first approval here. Re-reviewing nothing is immediately jumping out to me as blocking thought I do want to make sure we get another approval before moving forward.

timflannagan:
#201 just merged - re-spinning the linting check.

exdx:
rebased off main -- lint should pass now I think.

exdx:
Follow-ups in #204

timflannagan (approved):

Issue #53, opened 2022-02-23
Build rukpak images for multiple architectures


joelanford:
It'd be great to go ahead and setup multi-arch image builds. It will help us avoid (or at least be more mindful about) changes that break multi-arch image builds (e.g. accidental reliance on CGO)

Possible tools to use:
- https://github.com/marketplace/actions/docker-buildx-bake
- https://goreleaser.com/

References:
 - operator-lifecycle-manager [release job](https://github.com/openshift/operator-framework-olm/blob/master/staging/operator-lifecycle-manager/.goreleaser.yml)

**Acceptance Criteria:**
- release jobs for: amd64, ppc64Ie, s390x, arm64 architectures
- ensure compilation, environment docker images are updated
- binaries of all architectures go through the pipeline (get tested, etc.)

timflannagan:
Does this need to be a 0.0.1 release blocker?

joelanford:
not in my opinion

exdx:
0.3 candidate

PR #181, opened 2022-03-29
Add a new test-e2e-local target 

exdx:
Adds a new makefile target that that builds binaries locally, deploys rukpak, and then runs the e2e suite. This target is helpful for local testing, because the existing e2e target builds the binaries in a container and takes several minutes to just get started.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,6 +78,8 @@ test-unit: setup-envtest ## Run the unit tests
 test-e2e: ginkgo ## Run the e2e tests
 	$(GINKGO) -v -trace -progress test/e2e
 
+test-e2e-local: kind-cluster run kind-load-bundles test-e2e ## Deploys rukpak onto a new local cluster and runs e2e tests
+
 ###################
 # Install and Run #
 ###################


joelanford:
I ran into this too. My initial thought was "rather than adding yet another make target that does something different than what CI does, can we use `Dockerfile.local` as _the_ Dockerfile for `make run` and `make e2e`.

Basically `git mv Dockerfile.local Dockerfile`, consolidate `build-local-container` and `build-container` make targets, and then fix whatever else up. 

I started looking into that and I think I ran into an issue with how that would work in a platform agnostic way.  It should be doable, though I think. I just didn't have the time to chase it.

joelanford:
Another potential option to consolidate even further is to use `GOOS=linux goreleaser build --single-target` to build the binaries and then see if `build-container` could make use of the goreleaser Dockerfile.

joelanford:
Alright check out #186 as an alternative.

That PR highlights to me that we need to:
- consolidate the `deploy` and `goreleaser` workflows (covered by #187)
- add build targets to `goreleaser` for multi-arch support (covered by #53)

exdx:
I think consolidating down to one Dockerfile per #186 is the better approach. Closing this one out in favor of that PR.

PR #186, opened 2022-03-29
Consolidate Dockerfiles down to one

### LINKED ISSUE #112: Revisit concerns about 2 Docker files for CI and local dev
The concerns are discussed here:
https://github.com/operator-framework/rukpak/pull/102

joelanford:
This PR:
- Deletes `Dockerfile` and `Dockerfile.goreleaser`
- Moves `Dockerfile.local` to `Dockerfile` (with a few minor tweaks)
- Removes the old `build-container` make target
- Renames `build-local-container` to `build-container` and changes all references to `build-container`
- Updates the `.github/workflows/deploy.yaml` workflow based on the above changes and configures it to run (but not push) for PR CI.

Closes #112


joelanford:
/hold cancel

Pretty sure the `deploy` workflow should work with the latest force push.

.github/workflows/deploy.yaml
```diff
@@ -4,24 +4,32 @@ on:
...
+    env:
+      IMAGE: quay.io/operator-framework/plain-provisioner:main
     steps:
       - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '~1.17'
+
+      - name: Build container
+        run: make build-container
```
timflannagan, line 22: Do we need to cache the build cache throughout this job?

joelanford: +1

timflannagan: Just create a quick issue so I don't forgot about this again.

Dockerfile
```diff
@@ -1,31 +1,9 @@
...
-COPY --from=builder /workspace/bin/core .
-EXPOSE 8080
 
+COPY plain plain
+COPY unpack unpack
+COPY core core
+
+EXPOSE 8080
 ENTRYPOINT ["/plain"]
-CMD ["run"]
```
exdx, line 31: This doesn't break anything? I guess this argument was never doing anything?

Makefile
```diff
@@ -108,31 +109,28 @@ cert-mgr: ## Install the certification manager
...
-build: bin/plain bin/unpack bin/core
-
-bin/plain:
-	CGO_ENABLED=0 go build $(VERSION_FLAGS) -o $@$(BIN_SUFFIX) ./internal/provisioner/plain
+build: plain unpack core
 
-bin/unpack:
-	CGO_ENABLED=0 go build $(VERSION_FLAGS) -o $@$(BIN_SUFFIX) ./cmd/unpack/...
+plain:
+	mkdir -p $(BIN_DIR)
```
exdx, line 121: should this command be a defined somewhere since we call it for each of the binary builds?

Dockerfile
timflannagan: I had the same initial reaction as well but we should be good to go with removing this?

joelanford: Correct, we don't have a `run` subcommand in the plain provisioner.

Makefile
joelanford: I could add a `$(BIN_DIR)` target that does `mkdir -p $(BIN_DIR)`? I'm always a little gun-shy about adding more targets because I feel like that makes the Makefile harder to understand and mentally parse.

But I'm up for it if that's there's consensus to do that.

timflannagan: Potentially dumb question: do we even need to create this bin/ directory? I was under the impression that go build would handle that for us implicitly.

joelanford: Good catch. Removed these because `go build` takes care of it.

timflannagan (approved):

PR #190, opened 2022-03-29
add validating webhook e2e test

### LINKED ISSUE #189: Add e2e test for webhook
Validating webhook for bundle api in https://github.com/operator-framework/rukpak/pull/156.  The webhook function must be tested in the e2e test.

akihikokuroda:

Add simple tests for bundle name length check.
Closes #189

### DIFF
--- a/test/e2e/webhook_test.go
+++ b/test/e2e/webhook_test.go
@@ -1,0 +1,87 @@
+package e2e
+
+import (
+	"context"
+	"fmt"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	rukpakv1alpha1 "github.com/operator-framework/rukpak/api/v1alpha1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ = Describe("bundle api validating webhook", func() {
+	When("Bundle is valid", func() {
+		var (
+			bundle *rukpakv1alpha1.Bundle
+			ctx    context.Context
+			err    error
+		)
+
+		BeforeEach(func() {
+			By("creating the valid Bundle resource")
+			ctx = context.Background()
+
+			bundle = &rukpakv1alpha1.Bundle{
+				ObjectMeta: metav1.ObjectMeta{
+					GenerateName: "valid-bundle-",
+				},
+				Spec: rukpakv1alpha1.BundleSpec{
+					ProvisionerClassName: plainProvisionerID,
+					Source: rukpakv1alpha1.BundleSource{
+						Type: rukpakv1alpha1.SourceTypeImage,
+						Image: &rukpakv1alpha1.ImageSource{
+							Ref: "testdata/bundles/plain-v0:valid",
+						},
+					},
+				},
+			}
+			err = c.Create(ctx, bundle)
+		})
+		AfterEach(func() {
+			By("deleting the testing Bundle resource")
+			err := c.Delete(ctx, bundle)
+			Expect(err).To(BeNil())
+		})
+		It("should create the bundle resource", func() {
+			Expect(err).To(BeNil())
+		})
+	})
+	When("the bundle name is too long", func() {
+		var (
+			bundle *rukpakv1alpha1.Bundle
+			ctx    context.Context
+			err    error
+		)
+		BeforeEach(func() {
+			By("creating the Bundle resource with a long name")
+			ctx = context.Background()
+
+			bundle = &rukpakv1alpha1.Bundle{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "bundlename-0123456789012345678901234567891",
+				},
+				Spec: rukpakv1alpha1.BundleSpec{
+					ProvisionerClassName: plainProvisionerID,
+					Source: rukpakv1alpha1.BundleSource{
+						Type: rukpakv1alpha1.SourceTypeImage,
+						Image: &rukpakv1alpha1.ImageSource{
+							Ref: "testdata/bundles/plain-v0:valid",
+						},
+					},
+				},
+			}
+			err = c.Create(ctx, bundle)
+		})
+		AfterEach(func() {
+			By("deleting the testing Bundle resource for failure case")
+			err := c.Delete(ctx, bundle)
+			Expect(err).To(WithTransform(apierrors.IsNotFound, BeTrue()))
+		})
+		It("should fail the long name bundle creation", func() {
+			Expect(err).To(MatchError(ContainSubstring(fmt.Sprintf("bundle name %s is too long: maximum allowed name length is 40", bundle.GetName()))))
+		})
+	})
+})


test/e2e/webhook_test.go
```diff
@@ -0,0 +1,82 @@
...
+						Image: &rukpakv1alpha1.ImageSource{
+							Ref: "testdata/bundles/plain-v0:valid",
+						},
+					},
+				},
+			}
+			err := c.Create(ctx, bundle)
+			Expect(err).To(BeNil())
+		})
+	})
```
timflannagan, line 47: Do we need a test case for valid Bundles? Is there too much overlap with the core testing suite?

test/e2e/webhook_test.go
```diff
@@ -0,0 +1,82 @@
...
+					Source: rukpakv1alpha1.BundleSource{
+						Type: rukpakv1alpha1.SourceTypeImage,
+						Image: &rukpakv1alpha1.ImageSource{
+							Ref: "testdata/bundles/plain-v0:valid",
+						},
+					},
+				},
+			}
+			err := c.Create(ctx, bundle)
+			Expect(err).NotTo(BeNil())
```
timflannagan, line 78: Is there any way we can perform this logic in a BeforeEach clause, and the It spec compares the error with the ContainSubstring gomega matcher?

timflannagan: Alternatively: we just deflate this test case entirely as a top-level It spec if we don't have the need for more test cases right now?

test/e2e/webhook_test.go
akihikokuroda: I see.  I take this out.

test/e2e/webhook_test.go
timflannagan: Another thing that comes to mind: should we be testing this workflow from the kubectl [create,apply] -f perspective? Is there any potential for skew between using the dynamic client, and a local kubectl CLI experience?

akihikokuroda: Yes, I'll try that.  Thanks!

test/e2e/webhook_test.go
joelanford: I actually think this is valuable enough to leave it. It does overlap with the plain provisioner tests, but it probably makes sense to have positive and negative tests for the webhook logic even if there is overlap with the provisioner e2es.

test/e2e/webhook_test.go
akihikokuroda: OK.  I'll just restructure the test.  Thanks!

joelanford: > Is there any way we can perform this logic in a BeforeEach clause, and the It spec compares the error with the ContainSubstring gomega matcher?

+1 from me on this suggestion. The `BeforeEach` setup should match the text of what the parent `When` is describing.

test/e2e/webhook_test.go
```diff
@@ -0,0 +1,82 @@
...
+			Expect(err.Error()).To(ContainSubstring("bundle name bundlename-0123456789012345678901234567891 is too long: maximum allowed name length is 40"))
```
joelanford, line 79: Pretty sure you can eliminate the nil check with this:
```suggestion
			Expect(err).To(MatchError(ContainSubstring("bundle name bundlename-0123456789012345678901234567891 is too long: maximum allowed name length is 40")))
```

akihikokuroda: Thanks!  I update.

test/e2e/webhook_test.go
akihikokuroda: OK.  I'll change to put it into BeforeEach

test/e2e/webhook_test.go
```diff
@@ -0,0 +1,47 @@
...
+		ctx    context.Context
+		err    error
+	)
+	BeforeEach(func() {
+		By("creating the Bundle resource with a long name")
+		ctx = context.Background()
+
+		bundle = &rukpakv1alpha1.Bundle{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "bundlename-0123456789012345678901234567891",
```
timflannagan, line 25: nit: use genName instead of hardcoding a Bundle name which might make tests more brittle over time?

test/e2e/webhook_test.go
```diff
@@ -0,0 +1,47 @@
...
+		}
+		err = c.Create(ctx, bundle)
+	})
+	AfterEach(func() {
+		By("deleting the testing Bundle resource for failure case")
+		err := c.Delete(ctx, bundle)
+		Expect(err).NotTo(BeNil())
+	})
+	It("should fail the long name bundle creation", func() {
+		Expect(err).To(MatchError(ContainSubstring("bundle name bundlename-0123456789012345678901234567891 is too long: maximum allowed name length is 40")))
```
timflannagan, line 45: same comment as above: use fmt.Sprintf to inject the bundle.GetName() instead of hardcoding the Bundle's metadata.Name directly into the error message we expect?

timflannagan:
I think we're getting pretty close after the latest force push (assuming those changes pass the e2e checks that are running now).

test/e2e/webhook_test.go
timflannagan: Okay can we use bundle.GetName() instead of hardcoding the Bundles name for the error message check, so we avoid hardcoding in two places?

akihikokuroda: Yes, it's better.  I make the change.  Thanks!

test/e2e/webhook_test.go
```diff
@@ -0,0 +1,48 @@
...
+		err = c.Create(ctx, bundle)
+	})
+	AfterEach(func() {
+		By("deleting the testing Bundle resource for failure case")
+		err := c.Delete(ctx, bundle)
+		Expect(err).NotTo(BeNil())
+	})
+	It("should fail the long name bundle creation", func() {
+		Expect(err).To(MatchError(ContainSubstring(fmt.Sprintf("bundle name %s is too long: maximum allowed name length is 40", bundle.GetName()))))
+	})
```
joelanford, line 47: Enclose all of this in a `When("the bundle name is too long")` block?

test/e2e/webhook_test.go
```diff
@@ -0,0 +1,48 @@
...
+					},
+				},
+			},
+		}
+		err = c.Create(ctx, bundle)
+	})
+	AfterEach(func() {
+		By("deleting the testing Bundle resource for failure case")
+		err := c.Delete(ctx, bundle)
+		Expect(err).NotTo(BeNil())
```
joelanford, line 43: We always expect this delete to fail because the bundle wasn't ever created, right? Can we be more explicit about the error we expect?
```suggestion
		Expect(err).To(MatchError(WithTransform(apierrors.IsNotFound), Equal(true))
```

akihikokuroda: Yes, that's right.  I'll update.  Thanks!

timflannagan (approved):

joelanford (approved):
Looks great! Thanks!

PR #191, opened 2022-03-30
Introduce Source interface with Image implementation

### LINKED ISSUE #164: Generalize bundle digest status field for other content sources

joelanford:
This PR introduces a new `Source` interface that defines a method that various source types can implement to abstract fetching bundle content.

```go
type Source interface {
	SourceContent(context.Context, *rukpakv1alpha1.Bundle) (*Result, error)
}
```

The design is a light abstraction around the current async process. It defines a `Result` type, which contains a state field that can carry `Pending`|`Unpacking`|`Unpacked` states. 

The provisioner can populate the bundle status and make decisions about re-reconciling and/or returning an error based on the returned Result and error values.

Lastly this PR changes the `bundle.status.digest` field to a `bundle.status.resolvedSource`, which the source is able to populate via the `Result` type.

Closes #164 


api/v1alpha1/bundle_types.go
```diff
@@ -48,11 +48,17 @@ type BundleSpec struct {
...
+	Type BundleSourceType `json:"type"`
 	// Image is the bundle image that backs the content of this bundle.
 	Image *ImageSource `json:"image,omitempty"`
 }
 
+type BundleSourceType string
+
+const (
+	BundleSourceTypeImage BundleSourceType = "image"
+)
```
timflannagan, line 60: Nice - these changes seem reasonable outside of the scope of this PR.

api/v1alpha1/bundle_types.go
```diff
@@ -64,7 +70,7 @@ type ProvisionerID string
 type BundleStatus struct {
 	Info               *BundleInfo        `json:"info,omitempty"`
 	Phase              string             `json:"phase,omitempty"`
-	Digest             string             `json:"digest,omitempty"`
+	ResolvedSource     *BundleSource      `json:"resolvedSource,omitempty"`
```
timflannagan, line 73: #164

internal/source/image.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+	defer logReader.Close()
+	buf := &bytes.Buffer{}
+	if _, err := io.Copy(buf, logReader); err != nil {
+		return nil, err
+	}
+	return buf.Bytes(), nil
+}
+
+func (i *Image) handleUnexpectedPod(ctx context.Context, pod *corev1.Pod) error {
```
timflannagan, line 185: nit: DI the client and make this a function?

internal/source/source.go
```diff
@@ -0,0 +1,67 @@
...
+	"sync"
+
+	rukpakv1alpha1 "github.com/operator-framework/rukpak/api/v1alpha1"
+)
+
+type Source interface {
+	SourceContent(context.Context, *rukpakv1alpha1.Bundle) (*Result, error)
+}
+
+type Result struct {
```
timflannagan, line 16: How do we feel about SourceResult (or anything slightly more explicit that could potentially improve readability)?

internal/source/source.go
```diff
@@ -0,0 +1,67 @@
...
+	Message        string
+}
+
+type State string
+
+const (
+	StatePending   State = "Pending"
+	StateUnpacking State = "Unpacking"
+	StateUnpacked  State = "Unpacked"
+)
```
timflannagan, line 29: Do we need an internal/source/types.go?

internal/source/source.go
```diff
@@ -0,0 +1,67 @@
...
+		return ErrSourceAlreadyAdded
+	}
+
+	s.m.Lock()
+	defer s.m.Unlock()
+	s.sources[typ] = source
+	return nil
+}
+
+func (s *Sourcer) For(typ rukpakv1alpha1.BundleSourceType) (Source, error) {
```
timflannagan, line 59: Just to clarify my understanding: we're trying to check whether the type parameter we'd passed is a valid, supported Bundle source type for this provisioner. Is that correct? Is there a better name for this method?

internal/source/image.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+
+	gzr, err := gzip.NewReader(bytes.NewReader(bd.Content))
+	if err != nil {
+		return nil, fmt.Errorf("read bundle content gzip: %w", err)
+	}
+	return tarfs.New(gzr)
+}
+
+func (i *Image) getBundleImageDigest(pod *corev1.Pod) (string, error) {
```
timflannagan, line 163: nit: Does this need to be a method?

internal/source/image.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+
+	return &Result{Bundle: bundleFS, ResolvedSource: resolvedSource, State: StateUnpacked}, nil
+}
+
+func (i *Image) getBundleContents(ctx context.Context, pod *corev1.Pod) (fs.FS, error) {
+	bundleData, err := i.getPodLogs(ctx, pod)
+	if err != nil {
+		return nil, fmt.Errorf("get bundle contents: %w", err)
+	}
```
timflannagan, line 147: nit: Pass the `bundleData []byte` as a parameter instead?

internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -92,64 +78,74 @@ func (r *BundleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ c
...
+	switch res.State {
+	case source.StatePending:
+		updateStatusUnpackPending(&u, res)
+		return ctrl.Result{}, nil
+	case source.StateUnpacking:
+		updateStatusUnpacking(&u)
 		return ctrl.Result{}, nil
-	case corev1.PodRunning:
-		r.handleRunningPod(&u)
+	case source.StateUnpacked:
```
timflannagan, line 98: Nice - this seems much more streamlined. Any value in moving the logic for handling an unpacked source to it's own Bundle controller method?

internal/provisioner/plain/main.go
```diff
@@ -123,13 +127,25 @@ func main() {
...
+		ProvisionerName: plainBundleProvisionerName,
+		PodNamespace:    ns,
+		UnpackImage:     unpackImage,
+	}
+
+	sourcer := source.NewSourcer()
+	if err := sourcer.Add(rukpakv1alpha1.BundleSourceTypeImage, imageSource); err != nil {
+		setupLog.Error(err, "unable add bundle image sourcer")
+		os.Exit(1)
+	}
```
timflannagan, line 142: I think this is fine for now, but this main.go could balloon in potential complexity as we introduce new supported sourcing types. I had a comment earlier that it would be nice if the sourcer.Add method had support for variadic source implementations, but it's possible I'm thinking about this in the wrong light.

internal/source/image.go
```diff
@@ -0,0 +1,188 @@
...
+
+	digest, err := i.getBundleImageDigest(pod)
+	if err != nil {
+		return nil, fmt.Errorf("get bundle image digest: %w", err)
+	}
+
+	resolvedSource := &rukpakv1alpha1.BundleSource{
+		Type:  rukpakv1alpha1.BundleSourceTypeImage,
+		Image: &rukpakv1alpha1.ImageSource{Ref: digest},
+	}
```
timflannagan, line 138: I'm curious on why you settled for this kind of reference tracking vs. tracking something like a hash of the Bundle in it's status block if we're mainly interested right now in making the bundle.Status.Digest field more agnostic to Bundle source types.

internal/source/source.go
timflannagan: Also: nit: godoc to implement readability?

internal/source/source.go
timflannagan: I think I'm slowly coming back around to this naming after the first pass through these changes. I'd still like to see a comment that helps explain the functionality so it's a bit easier to reason about when interacting with the codebase, but this isn't blocking anyways.

internal/source/source.go
joelanford: We _could_ I suppose. I stuffed all the general "applies to all sources" stuff into this file, which is still pretty small. Seemed unnecessary to me to split things out.

internal/source/source.go
joelanford: It's effectively a map lookup. "Get the `Source` for this type". `Has` wouldn't make sense (as that would intuitively return a `bool`). `Get` could make sense. I tend to shy away from functions called `Get` since they're so generic sounding.

Seeing something like `src, err := sourcer.For(rukpakv1alpha1.BundleSourceTypeImage)` reads pretty nicely to me.

Regardless, +1 on GoDoc.

internal/source/source.go
```diff
@@ -0,0 +1,67 @@
...
+var (
+	ErrSourceAlreadyAdded = errors.New("source type already added")
+	ErrSourceNotSupported = errors.New("source type not supported")
+)
+
+func NewSourcer() *Sourcer {
+	return &Sourcer{sources: map[rukpakv1alpha1.BundleSourceType]Source{}}
+}
+
+func (s *Sourcer) Add(typ rukpakv1alpha1.BundleSourceType, source Source) error {
```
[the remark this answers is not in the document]
joelanford: Making it variadic without a key/value struct would sacrifice compile-time checking since we'd have to make the function `Add(keysAndValues ...interface{})`, which has three runtime ways of going wrong instead of just one:
1. The key already exists (that's a check we would need no matter what)
2. The length of `keysAndValues` is odd
3. The values at even indices are not bundle source types or the values add odd indices are not `Source`s.

There are two other options:
1. Change this to accept a map that we can iterate.
2. Change this to accept a list of `SourcerMapping` objects (or some variation of that name).

internal/source/image.go
joelanford: Two reasons:
1. It would enable the user to have a way to go fetch the original source and (if they needed to) create a new pinned bundle.
2. It would potentially give us a way to re-pull bundle content more deterministically if the storage for that bundle was lost. We could always re-pull from the pinned source in the status.

internal/source/image.go
joelanford: Main reason I didn't do that is because I expect this package to contain at least a handful of other sources and I wanted to keep helper functions scoped to the specific source they're for.

I can change it to `handleUnexpectedImageUnpackPod` though?

joelanford:
/hold

I realized this still doesn't fully decouple the provisioner from the source implementation. The provisioner sets up watches for pods owned by Bundles, but that detail should be the responsibility of the source not the provisioner.

I think I need to iterate on the `Source` and `Sourcer` types a bit more.

internal/source/source.go
```diff
@@ -0,0 +1,67 @@
...
+	s.m.RLock()
+	_, ok := s.sources[typ]
+	s.m.RUnlock()
+	if ok {
+		return ErrSourceAlreadyAdded
+	}
+
```
akihikokuroda, line 52: nit: I assume that this mutex for thread safe.  If 2 threads for the same `typ` come here at the same time and both get the `ok` false, the source of one thread is overwritten by the other thread without an error.

joelanford:
Updated to rebase onto `main` and pick up (and move) the new git source to a new `Source` implementation

internal/source/git.go
```diff
@@ -0,0 +1,219 @@
...
+}
+
+func (g *Git) failedPodResult(ctx context.Context, pod *corev1.Pod) error {
+	logs, err := g.getPodLogs(ctx, pod)
+	if err != nil {
+		return fmt.Errorf("unpack failed: failed to retrieve failed pod logs: %w", err)
+	}
+	_ = g.Client.Delete(ctx, pod)
+	return fmt.Errorf("unpack failed: %v", string(logs))
+}
```
joelanford, line 153: I think this function needs to be improved for the git source. I tried a negative test by specifying a non-existent directory and the failure happened in the `clone-repository` init container. Problem is, we don't pull the logs from that container, we always pull the logs from the bundle container. In this failure case, the bundle container never even runs because an init container failed, so the error message that gets propagated to the bundle is along the lines of "failed to get logs from pod because bundle container is terminated"

Note: I think this is an issue in the current implementation. It's not a huge problem, because the bundle is still correctly showing an unpack failure. Its just that you can't tell what the root cause was from looking at the bundle, because we could have propagated a better error message if we had better logic for checking which containers within the pod failed.

internal/source/source.go
joelanford: I updated this logic to use a write lock around the entire function. That way by the time the second thread enters the critical section, the map will include the contents of the first thread's source and that will result in the second thread getting the error.

internal/source/git.go
exdx: Yes, this is definitely a limitation with the current design, and I ran into it quite a lot when developing it. I probably should have called it out more. 

We could potentially make this more robust by checking the status of the Pod before grabbing the logs, and if it shows an InitContainer error, grab the logs of the failed init container and pass those back as part of an error.

timflannagan: Yep I just hit this playing around with rukpak locally and having to manually view the clone-repository init container's logs to figure out what's happening vs. interacting directly with the Bundle and/or BundleInstance APIs seems like a potentially large blindspot. As always, it might be more beneficial to fix this outside of any of Joe's expiremental PRs (although I think this one is close).

joelanford: The other issue with a pod-based git clone is that there isn't an easy way to get the resolved commit hash back out. Its easy to get the image sha for image sources because the digest is on the pod API. But the only way to get the commit sha of a git branch/tag is to extract it from a log written by the git clone container.

I've got a branch built on top of this PR that subs the pod-based git clone out for an in-process clone. And its trivial in that case to resolve the commit hash and propagate it back to the bundle.

internal/source/image.go
```diff
@@ -0,0 +1,198 @@
...
+		return nil, i.failedPodResult(ctx, pod)
+	case corev1.PodSucceeded:
+		return i.succeededPodResult(ctx, pod)
+	default:
+		return nil, i.handleUnexpectedPod(ctx, pod)
+	}
+}
+
+func (i *Image) RegisterWatches(c controller.Controller) error {
+	return c.Watch(&source.Kind{Type: &corev1.Pod{}}, &handler.EnqueueRequestForOwner{
```
timflannagan, line 69: Do we need to watch for changes to the object/metadata ConfigMaps(s) for either the git or image based Source implementations?

internal/source/image.go
timflannagan: Okay that's fine with me. Let's try and land the core functionality first, so I don't think we need to rename this method right now.

internal/source/image.go
timflannagan: I'm just going to mark this as resolved as this is a nit and an issue with the existing implementation.

api/v1alpha1/bundle_types.go
timflannagan: @joelanford Following back up on this - is it worth splitting this change out into it's own PR while we continue to have reviewers parse through this and finalize on an implementation?

internal/source/source.go
timflannagan: Yep I think I'm fine with the current naming after a couple of read through(s) with these changes, and I think my first read would've been easier if we integrated godocs for that method but this isn't blocking anyways so feel free to resolve.

internal/provisioner/plain/controllers/bundle_controller.go
timflannagan: (we can mark this as resolved - this feels like it's in nit territory anyways)

internal/source/git.go
timflannagan: Okay as always with your POC/larger effort PRs (:laughing:): is it worth trying to split out these changes gradually and seeing what sticks? It sounds like if you've already have an implementation aimed an in-process git unpacking process that's based on top of this source interface. Would it be easier to have multiple people shore up capacity and try and push these changes through, or re-implement using the existing unpacking process and risk implementation churn?

internal/source/image.go
joelanford: No. The `ConfigMap` watch is the domain of the bundle controller because it is the thing creating/updating config maps after the `Source` has unpacked the content and handed it to the controller in the from of an `fs.FS`.

internal/source/git.go
joelanford: I just pushed a new commit to this PR to add GoDoc and address a few more comments and other concerns I had.

I have a gut instinct that the `Source` interface still isn't quite right, but I'm not _too_ worried about that at this stage (we can always refactor that interface later if it becomes more obvious that there are shortcomings with it).

I'm happy to break this up into smaller chunks if that would help get it through, though it isn't immediately obvious to me what the cut lines are, other than a few really small changes we could make separately.

internal/source/source.go
joelanford: I've completely removed the mutex in the latest commit, which is made possible by forcing the set of sources to be provided as a map only during initialization of the object. New sources can no longer be added later.

internal/source/source.go
joelanford: I changed this up one more time, such that `Sourcer` is now an unexported struct that implements the `Source` interface. So now the controller just immediately calls `Unpack` on it's `Source` rather than getting the underlying source and then calling `Unpack`. I think this is a bit cleaner.

internal/provisioner/plain/main.go
joelanford: Resolved this by making `source.New` take a map of `BundleSourceType -> Source`. It also no longer returns an error because there's no longer a chance for collision.

api/v1alpha1/bundle_types.go
```diff
@@ -109,10 +109,10 @@ type BundleStatus struct {
 //+kubebuilder:object:root=true
 //+kubebuilder:resource:scope=Cluster
 //+kubebuilder:subresource:status
-//+kubebuilder:printcolumn:name=Reference,type=string,JSONPath=`.spec.source.image.ref`,priority=1
 //+kubebuilder:printcolumn:name=Type,type=string,JSONPath=`.spec.source.type`
 //+kubebuilder:printcolumn:name=Phase,type=string,JSONPath=`.status.phase`
 //+kubebuilder:printcolumn:name=Age,type=date,JSONPath=`.metadata.creationTimestamp`
+//+kubebuilder:printcolumn:name=Resolved Source,type=string,JSONPath=`.status.resolvedSource`,priority=1
```
joelanford, line 115: This is hack-ish, but it gets around the problem of there not being a single string field that every source config could fit into to define how that source "resolved".

```
$ k get bundles -o wide
NAME                  TYPE    PHASE      AGE   RESOLVED SOURCE
memcached-api-0.1.0   image   Unpacked   4s    {"image":{"ref":"quay.io/joelanford/memcached-api-bundle@sha256:992ca690771a03dfc552c8f8c0d549f8143f09f3c3d8fbc99e48943b7400415b"},"type":"image"}
memcached-api-0.2.0   image   Unpacked   4s    {"image":{"ref":"quay.io/joelanford/memcached-api-bundle@sha256:7f7de1072ab64df45e9aed826b45b4de46e9c1ed44d40f7e7ab35b645941b8cc"},"type":"image"}
```

timflannagan: Can we simply index into the resolvedSource.Ref field here instead of printing out the entire BundleSource object?

joelanford: Not if we also want to support git sources since the path is different.

internal/provisioner/plain/main.go
```diff
@@ -165,15 +166,31 @@ func main() {
...
+		},
+		rukpakv1alpha1.SourceTypeGit: &source.Git{
+			Client:          mgr.GetClient(),
+			KubeClient:      kubeClient,
+			ProvisionerName: plainBundleProvisionerName,
+			PodNamespace:    ns,
+			UnpackImage:     unpackImage,
+			GitClientImage:  gitClientImage,
+		},
+	})
```
timflannagan, line 186: Follow-up: any value in updating the NewUnpacker constructor to work with the options pattern?

joelanford: Hadn't thought too much about that option, but I'm not opposed!

Issue #183, opened 2022-03-29
Introduce an e2e test for a BundleInstance that depends on another BundleInstance


timflannagan:
Goal: introduce a new BI e2e test that creates a BI that contains Bundle contents that rely on Bundle contents that don't exist on-cluster, then create a BI that contains those requisite resources, and waits until the original BI reports a successful state.

joelanford:
The idea here is that we should keep retrying BIs that fail to rollout due to some missing prereq?

timflannagan:
Yep that's what I had in mind - basically we're interested in eventually consistency here?

PR #193, opened 2022-03-30
e2e: Introduce an e2e test for a BI that's dependent on another BI resource

timflannagan:
Introduce a BI e2e test that creates a set of Bundle/BI resources that reference namespace-scoped CRs, and the corresponding CRD is not present on cluster, wait until that BI reports a failed installation state, and then create the requisite Bundle/BI resources that contain the CRD definition and ensure the original BI eventually reports a successful state.

Closes: #183 


test/e2e/plain_provisioner_test.go
```diff
@@ -544,6 +544,130 @@ var _ = Describe("plain provisioner bundleinstance", func() {
...
+				}
+				return meta.FindStatusCondition(bi.Status.Conditions, rukpakv1alpha1.TypeInstalled), nil
+			}).Should(And(
+				Not(BeNil()),
+				WithTransform(func(c *metav1.Condition) string { return c.Type }, Equal(rukpakv1alpha1.TypeInstalled)),
+				WithTransform(func(c *metav1.Condition) metav1.ConditionStatus { return c.Status }, Equal(metav1.ConditionFalse)),
+				WithTransform(func(c *metav1.Condition) string { return c.Reason }, Equal(rukpakv1alpha1.ReasonInstallFailed)),
+				WithTransform(func(c *metav1.Condition) string { return c.Message }, ContainSubstring(`uninstall failed`)),
+			))
+		})
```
joelanford, line 607: Perhaps put this in a `When("the dependent BundleInstance does not exist")` block?

Makefile
```diff
@@ -136,10 +136,14 @@ build-local-container: build ## Builds the provisioner container image using loc
 
 kind-load-bundles:
 	$(CONTAINER_RUNTIME) build $(TESTDATA_DIR)/bundles/plain-v0/valid -t testdata/bundles/plain-v0:valid
+	$(CONTAINER_RUNTIME) build $(TESTDATA_DIR)/bundles/plain-v0/dependent -t testdata/bundles/plain-v0:dependent
+	$(CONTAINER_RUNTIME) build $(TESTDATA_DIR)/bundles/plain-v0/provides -t testdata/bundles/plain-v0:provides
```
timflannagan, line 140: Note: I could've used the valid testdata manifests and achieved the same behavior, but I was slightly concerned with breaking this test over time if we chose to refactor those manifests.

test/e2e/plain_provisioner_test.go
```diff
@@ -544,6 +544,130 @@ var _ = Describe("plain provisioner bundleinstance", func() {
...
+			})
+			AfterEach(func() {
+				By("deleting the testing provides Bundle resource")
+				Expect(client.IgnoreNotFound(c.Delete(ctx, providesBundle))).To(BeNil())
+
+				By("deleting the testing provides BundleInstance resource")
+				Expect(client.IgnoreNotFound(c.Delete(ctx, providesBI))).To(BeNil())
+
+			})
+			It("should eventually project a successful installation", func() {
```
joelanford, line 655:
```suggestion
			It("should eventually project a successful installation for the dependent BundleInstance", func() {
```

test/e2e/plain_provisioner_test.go
```diff
@@ -544,6 +544,130 @@ var _ = Describe("plain provisioner bundleinstance", func() {
 		})
 	})
 
+	When("a BundleInstance is dependent on another BundleInstance resource", func() {
+		var (
+			ctx    context.Context
+			bundle *rukpakv1alpha1.Bundle
+			bi     *rukpakv1alpha1.BundleInstance
```
joelanford, line 551:
```suggestion
			dependentBundle *rukpakv1alpha1.Bundle
			dependentBI     *rukpakv1alpha1.BundleInstance
```

test/e2e/plain_provisioner_test.go
```diff
@@ -544,6 +544,130 @@ var _ = Describe("plain provisioner bundleinstance", func() {
...
+				Eventually(func() (*metav1.Condition, error) {
+					if err := c.Get(ctx, client.ObjectKeyFromObject(bi), bi); err != nil {
+						return nil, err
+					}
+					return meta.FindStatusCondition(bi.Status.Conditions, rukpakv1alpha1.TypeInstalled), nil
+				}).Should(And(
+					Not(BeNil()),
+					WithTransform(func(c *metav1.Condition) string { return c.Type }, Equal(rukpakv1alpha1.TypeInstalled)),
+					WithTransform(func(c *metav1.Condition) metav1.ConditionStatus { return c.Status }, Equal(metav1.ConditionTrue)),
+					WithTransform(func(c *metav1.Condition) string { return c.Reason }, Equal(rukpakv1alpha1.ReasonInstallationSucceeded)),
```
joelanford, line 665: Should we explicitly check for an empty message? Not sure if we do this in other places, but maybe we should. That would add a guarantee that we don't accidentally leave a stale message in the status once an install succeeds

test/e2e/plain_provisioner_test.go
```diff
@@ -544,6 +544,134 @@ var _ = Describe("plain provisioner bundleinstance", func() {
...
+					if err := c.Get(ctx, client.ObjectKeyFromObject(dependentBI), dependentBI); err != nil {
+						return nil, err
+					}
+					return meta.FindStatusCondition(dependentBI.Status.Conditions, rukpakv1alpha1.TypeInstalled), nil
+				}).Should(And(
+					Not(BeNil()),
+					WithTransform(func(c *metav1.Condition) string { return c.Type }, Equal(rukpakv1alpha1.TypeInstalled)),
+					WithTransform(func(c *metav1.Condition) metav1.ConditionStatus { return c.Status }, Equal(metav1.ConditionFalse)),
+					WithTransform(func(c *metav1.Condition) string { return c.Reason }, Equal(rukpakv1alpha1.ReasonInstallFailed)),
+					WithTransform(func(c *metav1.Condition) string { return c.Message }, ContainSubstring(`uninstall failed`)),
```
exdx: This is a strange message -- should it be `install failed`? If we aren't uninstalling in this context?

test/e2e/plain_provisioner_test.go
```diff
@@ -544,6 +544,135 @@ var _ = Describe("plain provisioner bundleinstance", func() {
 		})
 	})
 
+	When("a BundleInstance is dependent on another BundleInstance", func() {
```
exdx, line 547: nit: does it make it more clear to have the test name be slightly more specific ie `a BundleInstance is dependent on APIs provided by another BundleInstance?` Since BIs are independent of one another. I didn't fully understand the context of the test until I looked at the manifests in the testdata directory. 

Or is the intention here that this test is generalizable to all cases? Like for example if one BI needs a namespace to install a resource, and another BI comes along and installs that namespace, and then the first BI succeeds?

test/e2e/plain_provisioner_test.go
timflannagan: Yep - It looks like the failure we get back isn't always deterministic which is slightly annoying. I ended up changing this to check the error message is complaining about the OG CRD being missing on cluster, which should be a bit more readable and easier to reason about.

test/e2e/plain_provisioner_test.go
timflannagan: What I had in mind was closer to the latter, and we can easily emulate trying to instantiate a Bundle that contains a namespaced-scoped resource and explicitly defines a metadata.Namespace that doesn't exist on-cluster, and then create the providing Bundle/BI resource that rolls out that namespace.

joelanford (approved):

Makefile
timflannagan: Definitely - I can already see myself getting the itch to comb down the test data directory sizes at some point in the future.

Issue #159, opened 2022-03-22
Support nested bundle directory structures


exdx:
The current iteration of the unpacker does not support nested directories -- if the file encountered by the `WalkDirFunc` is a directory, [it gets skipped](https://github.com/operator-framework/rukpak/blob/bb79ed5efd7e7c7a9c8ff5c83fdb10697ecee2a5/cmd/unpack/main.go#L33).

Therefore the contents of that directory do not get applied to the cluster. This is evident when unpacking the [combo manifests directory](https://github.com/operator-framework/combo/tree/main/manifests), where only 5 resources get unpacked (the ones at the top-level) and the CRD resources located in the crd subdirectory are not applied.  

Since git-backed bundle sources can have arbitrary directory structures, the unpacker should be able to fully unpack all manifests from a specified folder that has a nested directory structure.

joelanford:
> if the file encountered by the WalkDirFunc is a directory, [it gets skipped](https://github.com/operator-framework/rukpak/blob/bb79ed5efd7e7c7a9c8ff5c83fdb10697ecee2a5/cmd/unpack/main.go#L33).

Returning `nil` early in the unpacker doesn't mean we don't walk the files in that directory. I'm fairly sure you'll see the combo CRDs in the unpacker log output.

However, the plain+v0 manifest reader specifically only reads the [manifests in the /manifests directory](https://github.com/operator-framework/rukpak/blob/3c6539db652ca020bfc47cc963070b27ce36c30a/internal/provisioner/plain/controllers/bundle_controller.go#L351), as per the (current, possibly underdefined) plain+v0 spec.

exdx:
Oh ok, that makes more sense. In that case it should be more straightforward to implement support for nested directories by updating the manifest reader to read manifests from all the directories that exist?

exdx:
I think this is a good thing to work on -- I would double check that this is still an issue, I know we refactored some of the relevant code recently. 

I believe we do want to support an arbitrary nested directory structure inside of the `/manifests` directory. In the plain bundle spec we mandated a flat directory but I would like to remove this requirement.

timflannagan:
Can we enforce this behavior with an e2e test?

akihikokuroda:
I'm thinking about adding the directories with a valid yaml file in it and an empty directory bundle in the e2e test.

timflannagan:
> Can we enforce this behavior with an e2e test?

[...]

PR #198, opened 2022-03-31
make sure nested manifest directories don't cause any affects

### LINKED ISSUE #195: Introduce an e2e test for plain bundles that contain nested manifest directories
Goal: introduce an e2e test that instantiates a Bundle resource that sources from a `manifests` directory that contains nested directories, and ensure the unpacker rolls out the expected number of resources.

See the [plain bundle format documentation](https://github.com/operator-framework/rukpak/blob/main/docs/plain-bundle-spec.md) or more details.

akihikokuroda:

Add a new bundle that has subdirectories in the manifests directory.  The yaml files in the subdirectory should be ignored.

Closes #195

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -145,12 +145,14 @@ kind-load-bundles: ## Load the e2e testdata container images into a kind cluster
 	$(CONTAINER_RUNTIME) build $(TESTDATA_DIR)/bundles/plain-v0/empty -t testdata/bundles/plain-v0:empty
 	$(CONTAINER_RUNTIME) build $(TESTDATA_DIR)/bundles/plain-v0/no-manifests -t testdata/bundles/plain-v0:no-manifests
 	$(CONTAINER_RUNTIME) build $(TESTDATA_DIR)/bundles/plain-v0/invalid-crds-and-crs -t testdata/bundles/plain-v0:invalid-crds-and-crs
+	$(CONTAINER_RUNTIME) build $(TESTDATA_DIR)/bundles/plain-v0/subdir -t testdata/bundles/plain-v0:subdir
 	${KIND} load docker-image testdata/bundles/plain-v0:valid --name $(KIND_CLUSTER_NAME)
 	${KIND} load docker-image testdata/bundles/plain-v0:dependent --name $(KIND_CLUSTER_NAME)
 	${KIND} load docker-image testdata/bundles/plain-v0:provides --name $(KIND_CLUSTER_NAME)
 	${KIND} load docker-image testdata/bundles/plain-v0:empty --name $(KIND_CLUSTER_NAME)
 	${KIND} load docker-image testdata/bundles/plain-v0:no-manifests --name $(KIND_CLUSTER_NAME)
 	${KIND} load docker-image testdata/bundles/plain-v0:invalid-crds-and-crs --name $(KIND_CLUSTER_NAME)
+	${KIND} load docker-image testdata/bundles/plain-v0:subdir --name $(KIND_CLUSTER_NAME)
 
 kind-load: ## Load-image loads the currently constructed image onto the cluster
 	${KIND} load docker-image $(IMAGE) --name $(KIND_CLUSTER_NAME)
--- a/internal/provisioner/plain/controllers/bundle_controller.go
+++ b/internal/provisioner/plain/controllers/bundle_controller.go
@@ -359,7 +359,7 @@ func getObjects(bundleFS fs.FS) ([]client.Object, error) {
 	}
 	for _, e := range entries {
 		if e.IsDir() {
-			continue
+			return nil, fmt.Errorf("subdirectories are not allowed within the %q directory of the bundle image filesystem: found %q", manifestsDir, filepath.Join(manifestsDir, e.Name()))
 		}
 		fileData, err := fs.ReadFile(bundleFS, filepath.Join(manifestsDir, e.Name()))
 		if err != nil {
--- a/test/e2e/plain_provisioner_test.go
+++ b/test/e2e/plain_provisioner_test.go
@@ -3,6 +3,7 @@ package e2e
 import (
 	"context"
 	"fmt"
+	"path/filepath"
 	"strings"
 
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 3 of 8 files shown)


akihikokuroda:
I'm not sure what is wrong with the sanity / lint.  Where should I look at?

akihikokuroda:
I'm not sure if the lint error is related to https://github.com/golangci/golangci-lint/issues/2374

Meuko:
> I'm not sure if the lint error is related to [golangci/golangci-lint#2374](https://github.com/golangci/golangci-lint/issues/2374)

This happens if you upgrade go to 1.18, it seems that under the hood go(ci) has been bumped up overnight.

akihikokuroda:
I see.  It seems good idea to me.  I'll try.  Thanks!

test/e2e/plain_provisioner_test.go
```diff
@@ -345,6 +345,70 @@ var _ = Describe("plain provisioner bundle", func() {
...
+			err := c.Delete(ctx, bundle)
+			Expect(err).To(BeNil())
+		})
+		It("should eventually report a successful state", func() {
+			By("eventually reporting an Unpacked phase", func() {
+				Eventually(func() (string, error) {
+					if err := c.Get(ctx, client.ObjectKeyFromObject(bundle), bundle); err != nil {
+						return "", err
+					}
+					return bundle.Status.Phase, nil
```
joelanford, line 385: IMO, this should actually fail (and we should tighten up the spec and code to call this out as invalid).

Ignoring subdirectories will be a poor UX.

test/e2e/plain_provisioner_test.go
```diff
@@ -345,6 +345,70 @@ var _ = Describe("plain provisioner bundle", func() {
 			))
 		})
 	})
+	When("a bundle containing nested directry is created", func() {
```
joelanford, line 348:
```suggestion
	When("a bundle containing nested directory is created", func() {
```

Makefile
```diff
@@ -135,18 +135,21 @@ build-local-container: build ## Builds the provisioner container image using loc
 	$(CONTAINER_RUNTIME) build -f Dockerfile.local -t $(IMAGE) .
 
 kind-load-bundles:
+	-mkdir $(TESTDATA_DIR)/bundles/plain-v0/subdir/manifests/emptydir
```
joelanford, line 138: Since we already have `subdir/manifests/subdir` checked in, do we actually need `emptydir` to exercise the logic we expect? I'm curious if we can remove this line?

If we _do_ need the emptydir, I'd suggest checking it in with a `.gitignore` and a `.dockerignore` like we have with the [no-manifests](https://github.com/operator-framework/rukpak/tree/main/testdata/bundles/plain-v0/no-manifests) bundle.

akihikokuroda: Thank! I thought that It's better checking the empty directory doesn't cause any issue.  I can take it out.  
I tried checking in the `emptydir` in the `.gitignore` but it actually caused failure.  The manifest reader reads it and errored.  I'm thinking about opening an issue for it.

test/e2e/plain_provisioner_test.go
akihikokuroda: Thanks!  I fix it.

test/e2e/plain_provisioner_test.go
akihikokuroda: I agree.  It should be handled in https://github.com/operator-framework/rukpak/issues/159

Makefile
joelanford: > The manifest reader reads it and errored. I'm thinking about opening an issue for it.

I might be misunderstanding the issue, but I don't think the .gitignore file should be showing up in the container that the unpacker sees if you exclude the .gitignore file from the docker build with a .dockerignore file.

test/e2e/plain_provisioner_test.go
joelanford: #159 says we should support nested directories, but IMO, there's no reason to do that until there's a specific use case where nested directories are the most logical solution.  

> we should tighten up the spec and code to call [nested manifests directories] out as invalid

Do we need another issue that talks about that?

akihikokuroda: I thought #195 is only for the e2e test.  I'll include the manifest reader change in this PR.  Thanks!

Makefile
akihikokuroda: I missed the `.dockerignore` to remove the `.gitignore`.  I'll try.  Thanks!

akihikokuroda:
I updated the manifest reader to fail if the manifest directory has directories.

internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -359,7 +359,7 @@ func getObjects(bundleFS fs.FS) ([]client.Object, error) {
 	}
 	for _, e := range entries {
 		if e.IsDir() {
-			continue
+			return nil, fmt.Errorf("directory is not allowed: %q", e.Name())
```
exdx, line 362: nit: should this error string be referring to subdirectories to be more clear? for example `fmt.Errorf("subdirectories are not allowed in the bundle image filesystem: found %q", e.Name())`

akihikokuroda: It's a clearer message.  I'll update.  Thanks!

test/e2e/plain_provisioner_test.go
```diff
@@ -459,6 +459,55 @@ var _ = Describe("plain provisioner bundle", func() {
...
+			bundle *rukpakv1alpha1.Bundle
+			ctx    context.Context
+		)
+		BeforeEach(func() {
+			ctx = context.Background()
+
+			By("creating the testing Bundle resource")
+			bundle = &rukpakv1alpha1.Bundle{
+				ObjectMeta: metav1.ObjectMeta{
+					GenerateName: "olm-crds-unsupported",
```
timflannagan, line 474: super nit: can we change the name as there's nothing related to OLM in the testdata manifests?

akihikokuroda: Yes, I'll update.  Thanks!

internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -359,7 +359,7 @@ func getObjects(bundleFS fs.FS) ([]client.Object, error) {
 	}
 	for _, e := range entries {
 		if e.IsDir() {
-			continue
+			return nil, fmt.Errorf("subdirectories are not allowed in the bundle image filesystem: found %q", e.Name())
```
joelanford, line 362: One more nit on this error message: we're specifically disallowing subdirectories _within the manifests directory_.

```suggestion
			return nil, fmt.Errorf("subdirectories are not allowed within the %q directory of the bundle image filesystem: found %q", manifestsDir, filepath.Join(manifestsDir, e.Name()))
```

akihikokuroda: OK.  It's clearer.  I'll update. Thanks!

joelanford (approved):
Nice, thanks!

PR #199, opened 2022-04-01
[WIP] feat(bundleinstance): add call to crd validator inside of controller

tylerslaton:
## Summary
Doing a few things here that we should talk through:

1. Moving the `crd` package into its own separate one. Additionally remove its ability to execute a creation/upgrade.
2. Calling the `crd`.`Validate()` function on any incoming CRD's to ensure the upgrade won't cause any destructive actions.
3. Added a new `ReasonUnsafeCRDUpgrade` status that gets created on failure to `Validate` the `crd`

## Working On
This is not complete yet, there are still a few things I'm working through but wanted to get the approach into code review while working through theses:

1. Adding a label/annotation/spec field that turns this functionality off
2. Adding a few E2E test cases for this logic

### DIFF
--- a/api/v1alpha1/bundleinstance_types.go
+++ b/api/v1alpha1/bundleinstance_types.go
@@ -20,9 +20,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
-// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
-
 const (
 	TypeHasValidBundle       = "HasValidBundle"
 	TypeInvalidBundleContent = "InvalidBundleContent"
@@ -38,6 +35,7 @@ const (
 	ReasonReconcileFailed          = "ReconcileFailed"
 	ReasonCreateDynamicWatchFailed = "CreateDynamicWatchFailed"
 	ReasonInstallationSucceeded    = "InstallationSucceeded"
+	ReasonUnsafeCRDUpgrade         = "UnsafeCRDUpgrade"
 )
 
 // BundleInstanceSpec defines the desired state of BundleInstance
--- a/internal/crd/crd.go
+++ b/internal/crd/crd.go
@@ -1,4 +1,4 @@
-package util
+package crd
 
 import (
 	"context"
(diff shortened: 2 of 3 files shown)


api/v1alpha1/bundleinstance_types.go
```diff
@@ -38,6 +35,7 @@ const (
 	ReasonReconcileFailed          = "ReconcileFailed"
 	ReasonCreateDynamicWatchFailed = "CreateDynamicWatchFailed"
 	ReasonInstallationSucceeded    = "InstallationSucceeded"
+	ReasonUnsafeCRDUpgrade         = "UnsafeCRDUpgrade"
```
exdx, line 38: Nice 👍

internal/provisioner/plain/controllers/bundleinstance_controller.go
```diff
@@ -154,6 +156,32 @@ func (r *BundleInstanceReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+			TODO (tylerslaton): Convert crd validation into an AdmissionWebhook
+
+			Eventually we want to convert this code segment into an webhook that validates
+			this logic agnostic of any one provisioner. There was some contention around.
+			Eventually it was settled on implenting this interface but should not be the end
+			product for this behavior.
+		*/
+		var unmarshalledCRD apiextensionsv1.CustomResourceDefinition
+		err = yaml.Unmarshal(jsonData, &unmarshalledCRD)
+		if err == nil && unmarshalledCRD.GroupVersionKind() == apiextensionsv1.SchemeGroupVersion.WithKind("CustomResourceDefinition") {
```
exdx, line 173: This type check is a bit awkward -- I'm not sure but since we are unmarshaling to an `apiextensionsv1.CustomResourceDefinition` concrete type I imagine the GVK for that type will always be the v1 apiextensions GVK regardless of whether the underlying JSON represents a v1 CRD in reality. If it doesn't, then an error occurs (probably some field is not able to be unmarshaled successfully) which doesn't seem to be handled here. 

I would recommend as an intermediate step unmarshaling the jsonData into an unstructured.Unstructured object, which has a `GroupVersionKind()` method. From there we can do the GVK equality check and be more confident that the object we are handling is actually a CRD. It also shouldn't error since we are just converting JSON into another JSON-based type.

internal/crd/crd.go
```diff
@@ -13,49 +13,29 @@ import (
...
 
-func CreateOrUpdateCRD(ctx context.Context, cl client.Client, crd *apiextensionsv1.CustomResourceDefinition) (controllerutil.OperationResult, error) {
-	createCRD := crd.DeepCopy()
-	createErr := cl.Create(ctx, createCRD)
-	if createErr == nil {
-		return controllerutil.OperationResultCreated, nil
-	}
-	if !apierrors.IsAlreadyExists(createErr) {
-		return controllerutil.OperationResultNone, createErr
+func Validate(ctx context.Context, cl client.Client, crd *apiextensionsv1.CustomResourceDefinition) error {
```
exdx, line 19: nit: might make this function clearer to call the crd argument `newCRD` to differentiate it more from the `currentCRD` instantiated within

internal/provisioner/plain/controllers/bundleinstance_controller.go
exdx: Alternatively we can move this check up to L146 to do a `obj.GetObjectKind()` when we are handling the client.Object directly and handle the CRD from there..it makes the code a bit less clear though.

tylerslaton:
Closed in favor of #213

PR #201, opened 2022-04-01
deps: update golangci-lint version used in CI

exdx:
Fixes the perma-failing sanity/lint action which is failing due to https://github.com/golangci/golangci-lint/issues/2374


.github/workflows/sanity.yaml
```diff
@@ -24,4 +24,4 @@ jobs:
       - name: Run linting checks
         uses: "golangci/golangci-lint-action@v2"
         with:
-          version: "v1.43"
```
joelanford, line 27: Can we update the version in the makefile as well to match?

joelanford: Actually, I guess its `hack/tools/go.mod`

exdx: Sure 👍 I updated it and ran `go mod tidy` on the tools go.mod as well

akihikokuroda (approved):
Thanks for fixing this.  Looks good to me!

timflannagan (approved):

joelanford (approved):

PR #202, opened 2022-04-01
Makefile: Include cert-manager installation on `install`

anik120:
When using the target `install` on a plain kind cluster, I ran into the
following error:

```
validatingwebhookconfiguration.admissionregistration.k8s.io/rukpak-webhook created
unable to recognize "manifests/bundle-webhook/07_webhook.yaml": no matches for kind
"Certificate" in version "cert-manager.io/v1"
unable to recognize "manifests/bundle-webhook/07_webhook.yaml": no matches for kind
"Issuer" in version "cert-manager.io/v1"
```

Looks like the cert-manager needs to be installed on cluster to support PR #156

This PR includes the installation of the cert-manager to the `install` target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,7 +87,7 @@ test-e2e: ginkgo ## Run the e2e tests
 
 ##@ install/run:
 
-install-apis: generate ## Install the core rukpak CRDs
+install-apis: cert-mgr generate ## Install the core rukpak CRDs
 	kubectl apply -f manifests
 	kubectl apply -f manifests/bundle-webhook
 


tylerslaton (approved):
Nice, I like it! Thanks for the contribution.

timflannagan (approved):
Nice - we might need to consolidate the install/deploy target at some point as we should be installing cert-manager when running the deploy target.

Makefile
```diff
@@ -94,7 +94,7 @@ install-apis: generate ## Install the core rukpak CRDs
 install-plain: install-apis ## Install the rukpak CRDs and the plain provisioner
 	kubectl apply -f internal/provisioner/plain/manifests
 
-install: install-plain ## Install all rukpak core CRDs and provisioners
+install: cert-mgr install-plain ## Install all rukpak core CRDs and provisioners
```
akihikokuroda, line 97: Sorry for missed in #156.  I think the install-api requires it.

joelanford: +1 to move this to the `install-apis` target. That's what actually depends on cert manager.

anik120:
/hold 

Till I can address the feedback

anik120:
> Nice - we might need to consolidate the install/deploy target at some point as we should be installing cert-manager when running the deploy target.

@timflannagan looks like `deploy` is doing the same thing as `install-plain`. Since `install-plain` is already being used in other targets, is there any objections to just getting rid of the `deploy` target, so that it's not causing more confusion?

timflannagan:
@anik120 I _think_ that sounds reasonable to me, but we can iron that out in another PR anyways.

tylerslaton:
> > Nice - we might need to consolidate the install/deploy target at some point as we should be installing cert-manager when running the deploy target.
> 
> @timflannagan looks like `deploy` is doing the same thing as `install-plain`. Since `install-plain` is already being used in other targets, is there any objections to just getting rid of the `deploy` target, so that it's not causing more confusion?

I would be fine with remove it or making it more into an aggregate of all the provisioners and core api's related to RukPak. Something like this:

```make
deploy: install-apis install-plain ## Deploy the operator to the current cluster
```

Makefile
anik120: @akihikokuroda Done. Thanks for pointing that out!

anik120:
Cool yea that sounds reasonable to hash it out in another PR. 

/hold cancel

PR #206, opened 2022-04-01
(Makefile): Remove install and deploy Makefile targets

anik120:
This PR:

1) Uses the `local` tag for building local images that are used in Makefile targets
2) Cleans up Makefile targets by removing the `install` and `deploy` targets, and consolidates it all in the `run-local` target instead(addresses #209).    
 
Signed-off-by: Anik Bhattacharjee <anikbhattacharya93@gmail.com>

### DIFF
--- a/internal/provisioner/plain/main.go
+++ b/internal/provisioner/plain/main.go
@@ -53,6 +53,11 @@ func init() {
 	//+kubebuilder:scaffold:scheme
 }
 
+const (
+	defaultProvisionerImage    = "quay.io/operator-framework/plain-provisioner"
+	defaultProvisionerImageTag = "main"
+)
+
 func main() {
 	var metricsAddr string
 	var enableLeaderElection bool
@@ -64,7 +69,7 @@ func main() {
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.StringVar(&systemNamespace, "system-namespace", "rukpak-system", "Configures the namespace that gets used to deploy system resources.")
-	flag.StringVar(&unpackImage, "unpack-image", "quay.io/operator-framework/plain-provisioner:latest", "Configures the container image that gets used to unpack Bundle contents.")
+	flag.StringVar(&unpackImage, "unpack-image", defaultProvisionerImage+":"+defaultProvisionerImageTag, "Configures the container image that gets used to unpack Bundle contents.")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
--- a/internal/provisioner/plain/manifests/06_deployment.yaml
+++ b/internal/provisioner/plain/manifests/06_deployment.yaml
@@ -18,7 +18,7 @@ spec:
       serviceAccountName: plain-provisioner-admin
       containers:
         - name: plain-provisioner
-          image: quay.io/operator-framework/plain-provisioner:latest
+          image: quay.io/operator-framework/plain-provisioner:local
           imagePullPolicy: IfNotPresent
           ports:
             - containerPort: 8080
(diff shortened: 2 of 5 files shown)


internal/provisioner/plain/manifests/06_deployment.yaml
```diff
@@ -18,7 +18,7 @@ spec:
       serviceAccountName: plain-provisioner-admin
       containers:
         - name: plain-provisioner
-          image: quay.io/operator-framework/plain-provisioner:latest
+          image: quay.io/operator-framework/plain-provisioner:main
```
timflannagan, line 21: I realize we also default to the latest tag for the plain provisioner's --unpack-image flag: https://github.com/operator-framework/rukpak/blob/main/internal/provisioner/plain/main.go#L66.

anik120: @timflannagan ah nice thanks for pointing that out. I've added some constants to take care of that https://github.com/operator-framework/rukpak/pull/206/files#diff-00e176bec896af2959cbd2132aedac49f892da315eca7c81de39287295c3605dR56-R60

timflannagan: I like the idea of a constant: can we consolidate that into a single variable though?

timflannagan:
I think this lgtm from my perspective but let's get two approvals here before merging.

timflannagan (approved):
Note: I didn't pull down these changes, but I think this lgtm from my perspective. Let's get two approvals here before merging?

joelanford:
/hold

One question I'm trying to form...

anik120:
Yea we can wait. I've got things working finally locally so I'm good 👍🏽

internal/provisioner/plain/main.go
```diff
@@ -64,7 +69,7 @@ func main() {
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.StringVar(&systemNamespace, "system-namespace", "rukpak-system", "Configures the namespace that gets used to deploy system resources.")
-	flag.StringVar(&unpackImage, "unpack-image", "quay.io/operator-framework/plain-provisioner:latest", "Configures the container image that gets used to unpack Bundle contents.")
+	flag.StringVar(&unpackImage, "unpack-image", defaultProvisionerImage+":"+defaultProvisionerImageTag, "Configures the container image that gets used to unpack Bundle contents.")
```
exdx, line 72: Does changing the unpacker default image to point to the main tag by default result in always pulling down that image even when running locally? Should we instead be running the locally-built image when using `make run-local`? 

For example, say we are developing the unpacker locally and want to deploy it on the cluster. We would need some way to override this default to point to the `local` tag.  Is that scenario worth introducing a kustomize patch that would set the image to point to the local tag, when `make run-local` is used?

exdx, line 72: Also, as an aside, I'm not sure how much value the variables `defaultProvisionerImage` and `defaultProvisionerImageTag` are providing here since they are only called once and don't seem resuable. My preference would be not to introduce extra variables and simply update the embedded image name to point to `local` instead.

exdx, line 72: Also, as an aside, I'm not sure how much value the variables `defaultProvisionerImage` and `defaultProvisionerImageTag` are providing here since they are only called once and don't seem resuable. My preference would be not to introduce extra variables and simply update the embedded image name to point to `local` instead.

exdx (approved):
Looks good, just a thought

joelanford:
@anik120 Here's what I was planning to do after #235 and #236 merge: https://github.com/operator-framework/rukpak/pull/234/commits/b60d6e52d4e12192bd5ffddae014c32534166bce

Can we separate the "consolidate make targets" from the "how we coherently deal with the default unpack image"?

For the question about "how we coherently deal with the default unpack image". I'm thinking that the right thing to do is use a build variable and have the build inject the value of the default image name and/or tag. That way:
- when we build locally, we use `dev` as the tag.
- when we build the `main` branch, we use `main` as the tag.
- when we build a tag, we use the tag as the the tag.

PR #211, opened 2022-04-03
git bundle source: avoid copying full contents of unpack container

joelanford:
I was debugging the git source flake and I noticed that the log output
for the unpack pod was unexpectedly long. Turns out, the "contents" JSON
included the full filesystem of that container, which includes a few large
binaries. All we actually need is the bundle directory.

This commit resolves this with a few tweaks:
- Change the volume mount from "/manifests" to "/bundle".
- Just before copying, delete the repo's ".git" directory. This avoids
  copying that directory into /bundle when we're treating the entire git
  repository as the bundle.
- Copy the entire specified directory within the git repo into the
  /bundle directory (this results in "/bundle/manifests" for valid
  bundles.
- The unpack binary unpacks the "/bundle" directory rather than "/"

I'm not sure if this is the root cause of the flake we've been seeing,
but this fix _drastically_ decreases the size of the logs produced by
the unpack bundle, so it seems feasible and is a good change to make
regardless.

### DIFF
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -8,7 +8,7 @@ import (
 )
 
 const (
-	defaultDirectory = "./manifests"
+	defaultDirectory = "./"
 	repositoryName   = "repo"
 )
 
@@ -38,18 +38,18 @@ func (c *checkoutCmd) String() string {
 	}
 
 	if commit != "" {
-		checkoutCommand = fmt.Sprintf("git clone %s %s && cd %s && git checkout %s && cp -r %s/* /manifests",
+		checkoutCommand = fmt.Sprintf("git clone %s %s && cd %s && git checkout %s && rm -r .git && cp -r %s /bundle",
 			repository, repositoryName, repositoryName, commit, directory)
 		return checkoutCommand
 	}
 
 	if tag != "" {
-		checkoutCommand = fmt.Sprintf("git clone --depth 1 --branch %s %s %s && cd %s && git checkout tags/%s && cp -r %s/* /manifests",
+		checkoutCommand = fmt.Sprintf("git clone --depth 1 --branch %s %s %s && cd %s && git checkout tags/%s && rm -r .git && cp -r %s /bundle",
 			tag, repository, repositoryName, repositoryName, tag, directory)
 		return checkoutCommand
 	}
 
-	checkoutCommand = fmt.Sprintf("git clone --depth 1 --branch %s %s %s && cd %s && git checkout %s && cp -r %s/* /manifests",
+	checkoutCommand = fmt.Sprintf("git clone --depth 1 --branch %s %s %s && cd %s && git checkout %s && rm -r .git && cp -r %s /bundle",
 		branch, repository, repositoryName, repositoryName, branch, directory)
 	return checkoutCommand
 }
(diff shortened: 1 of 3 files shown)


internal/git/git.go
```diff
@@ -38,18 +38,18 @@ func (c *checkoutCmd) String() string {
 	}
 
 	if commit != "" {
-		checkoutCommand = fmt.Sprintf("git clone %s %s && cd %s && git checkout %s && cp -r %s/* /manifests",
```
timflannagan, line 41: I wonder whether we should just be copying the configured manifests directory (and any nested directories?) instead of the root of the repository? We could also be filtering for files that have the .yml or .yaml file extension to help alleviate a bloated volume. Any obvious downsides of this approach?

joelanford: In my mind, this code is meant to unpack arbitrary bundle content without any idea what that bundle content is. So I'm thinking that none of the source implementations should know anything about a `manifests` directory (or anything else).

Consider a helm chart provisioner. It should be able to re-use this git source implementation to unpack a chart directory, which doesn't contain a `manifests` directory and has a bunch of other important files.

If the entire repo is meant to be a bundle, then it makes sense to me that we'd copy the root of the repo. If a subdirectory is meant to be the bundle (e.g. `<repoDir>/bundle`), then the user would specify `bundle` as the repo directory, and we'd copy the full contents of that directory.

timflannagan: Thanks for the explanation - I think that sounds reasonable to me. One thing to note is whether we'll continue to see the unpacker component continue to be agnostic, in addition to acting as the entrypoint for sourcing Bundle contents from various source types. 

We've thrown around the idea on an in-process g/go-containerregistry implementation for image-based source types, but I think it's entirely possible we'll play with a git Go client to handle the checking out process (e.g. tag/branch/etc. concrete implementations), and if that's the case, it feels like the unpacking component doesn't add a ton of value in the context of the overall sourcing process.

I'm fine with pushing this through though as it seems like a short-term win while we continue to iterate on the underlying sourcing implementation, and removing potentially bloated directories like .git/ works with me.

timflannagan (approved):

internal/git/git.go
joelanford: > it feels like the unpacking component doesn't add a ton of value in the context of the overall sourcing process.

The way I've been thinking about this is that the `Bundle` source API is something we'll want all (or at least almost all) provisioners to implement, so I think we should build our source implementations to be agnostic to bundle types to decrease the effort required to build new provisioners.  If every provisioner has to write custom code to implement generic "pull a directory out of this artifact" code, that seems like it will result in every provisioner behaving slightly differently for really no reason other than it was implemented twice.

Another benefit of agnostic source unpackers is that we can keep tweaking/improving our plain+v0 spec without changing this implementation. And we'll be able to reuse it to unpack registry+v1 bundles.

joelanford:
I'll leave this open until tomorrow to give @exdx a chance to review.

internal/git/git.go
joelanford: And I agree that `go-git` could be a better implementation for git sources, but I'm not sure I understand how that changes the value of an agnostic unpacker.

timflannagan: Somewhat related: would it make sense to purge the vendor/ directory if it's present to help cut down on the size of the repository?

Issue #163, opened 2022-03-23
Generalize Bundle Image printcolumn for non-image based content sources


exdx:
Currently the printcolumn for a Bundle includes an `Image` field backed by the `spec.image` based on kubebuilder markers:
```
//+kubebuilder:printcolumn:name=Image,type=string,JSONPath=`.spec.image`
```
This works well for Bundles backed by an image, but results in a blank, empty column for non-image backed bundles.

As support is added for Bundles backed by additional content sources, this column should be generalized to output helpful reference information for the provided type of content. This custom column can be something generic like `Reference` and output different fields depending on the type of Bundle.  For example:
```
Type                                    Reference
Git-based bundle (tag)                  my-repo@v0.0.1
Git-based bundle (sha)                  my-repo@ae123xyz
Image-based bundle                      image:v0.0.1
```

timflannagan:
Interesting - I knew this issue existed but I was also thinking of updating the IMAGE -> Reference custom column, and changing the status.Digest field to status.Reference which the bundle controller is responsible for propagating down the stack.

timflannagan:
I was briefly playing around with this locally but I think we might run into some UX issues if we simply update that printer column to point to a status.Ref string that the bundle controller is responsible for populating, which might be fine but it's a change in behavior we've seen so far with a static spec.Image field implementation.

timflannagan:
@joelanford @exdx any idea on whether this could be solved using a more complex jsonpath query? I wasn't able to get something sketched out when I tried but that could also be a short term win while we iron out the underlying UX.

timflannagan:
Coming back to this again: I think we'd still run into this issue pushing through Joe's resolved source implementation in #191 if I'm reading the code correctly (and my brain isn't entirely dead for the day) where there's no guarantee that this field is set, especially in error paths, which is probably fine as long as we're all okay with the potential gap in UX. Maybe we can move the image custom column to -o wide output until we have a long term fix?

exdx:
> Maybe we can move the image custom column to -o wide output until we have a long term fix?

I think that's reasonable. If we don't want to have a default value for this printer field, and we don't want it to come up empty, we can make it part of -o wide output for now.

exdx:
Looking at this more today and it doesn't look like JSONPath supports conditional statements within one expression, so we couldn't do something like `spec.source.image.ref ? spec.source.image.ref : spec.source.git.repository` which I think was the intention.

PR #212, opened 2022-04-04
api,manifests: Introduce a 'Type' custom column to the Bundle CRD

timflannagan:
Purely exploratory right now but this seemed like it was a nice addition in my eyes when managing multiple Bundles locally:

```console
$ k get bundles
NAME           IMAGE                                           TYPE    PHASE       AGE
combo                                                          git     Failing     17m
combo-v0.0.1   quay.io/tflannag/bundle:combo-operator-v0.0.1   image   Unpacking   6s
```

We'll still have the problem around the IMAGE custom column (tracked in #163) but I think we can work around that limitation in parallel.

### DIFF
--- a/api/v1alpha1/bundle_types.go
+++ b/api/v1alpha1/bundle_types.go
@@ -117,6 +117,7 @@ type BundleObject struct {
 //+kubebuilder:resource:scope=Cluster
 //+kubebuilder:subresource:status
 //+kubebuilder:printcolumn:name=Image,type=string,JSONPath=`.spec.source.image.ref`
+//+kubebuilder:printcolumn:name=Type,type=string,JSONPath=`.spec.source.type`
 //+kubebuilder:printcolumn:name=Phase,type=string,JSONPath=`.status.phase`
 //+kubebuilder:printcolumn:name=Age,type=date,JSONPath=`.metadata.creationTimestamp`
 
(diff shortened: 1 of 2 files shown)


api/v1alpha1/bundle_types.go
```diff
@@ -117,6 +117,7 @@ type BundleObject struct {
 //+kubebuilder:resource:scope=Cluster
 //+kubebuilder:subresource:status
 //+kubebuilder:printcolumn:name=Image,type=string,JSONPath=`.spec.source.image.ref`
+//+kubebuilder:printcolumn:name=Type,type=string,JSONPath=`.spec.source.type`
```
timflannagan, line 120: Alternatively: we propagate this information from the status where we have the chance for more granularity, e.g. git -> git/tag, git/commit, etc.

timflannagan: There might be a tradeoff here though where you're forcing provisioner implementations to also have to propagate this to the Bundle's status vs. referencing this type information directly from the spec, where we'll eventually require that this field gets set.

tylerslaton (approved):
Nice UX improvement!

joelanford (approved):
As a follow-on, is it worth writing a test for the print columns?

timflannagan:
> As a follow-on, is it worth writing a test for the print columns?

This seems reasonable - do you have any concrete references that test printer columns?

joelanford:
> This seems reasonable - do you have any concrete references that test printer columns?

I don't :(

PR #213, opened 2022-04-04
feat(crdvalidator): adding webhook to ensure safety of crd create/upgrades

tylerslaton:
_NOTES_ 
- This is an alternative approach to #199
- Adding a few E2E test cases for this logic
- This webhook is intended to eventually live in its own repository separate from RukPak components. The implementation has been done with that in mind.

# Summary
Doing a few things here that we should talk through:

- Creating a new webhook (`crdvalidator`) under the `cmd` directory. There has been talk of moving this webhook into its own repo in the future and this should make that very easy.
- Adding Makefile targets to deploy the webhook
- Moving the `crd` util package along side the webhook code.

# Open Questions
- Should this have an option to force `crd` applies through or would that option just be removing this webhook?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,9 +102,12 @@ install-apis: cert-mgr generate kustomize ## Install the core rukpak CRDs
 install-plain: install-apis ## Install the rukpak CRDs and the plain provisioner
 	$(KUSTOMIZE) build manifests/provisioners/plain | kubectl apply -f -
 
+install-crdvalidator: cert-mgr kustomize ## Install the crdvalidator webhook manifests to the current cluster.
+	$(KUSTOMIZE) build manifests/crdvalidator | kubectl apply -f -
+
 install: install-plain ## Install all rukpak core CRDs and provisioners
 
-deploy: install-apis ## Deploy the operator to the current cluster
+deploy: install-apis install-crdvalidator ## Deploy the operator to the current cluster
 	$(KUSTOMIZE) build manifests/provisioners/plain | kubectl apply -f -
 
 run: build-container kind-load cert-mgr deploy ## Build image and run operator in-cluster
@@ -122,7 +125,7 @@ cert-mgr: ## Install the certification manager
 
 # Binary builds
 VERSION_FLAGS=-ldflags "-X $(VERSION_PATH).GitCommit=$(GIT_COMMIT)"
-build: plain unpack core
+build: plain unpack core crdvalidator
 
 plain:
 	CGO_ENABLED=0 go build $(VERSION_FLAGS) -o $(BIN_DIR)/$@ ./internal/provisioner/plain
@@ -132,6 +135,9 @@ unpack:
 
 core:
 	CGO_ENABLED=0 go build $(VERSION_FLAGS) -o $(BIN_DIR)/$@ ./cmd/core/...
+
+crdvalidator: 
+	CGO_ENABLED=0 go build $(VERSION_FLAGS) -o $(BIN_DIR)/$@ ./cmd/crdvalidator
 
 build-container: export GOOS=linux
 build-container: BIN_DIR:=$(BIN_DIR)/$(GOOS)
@@ -156,7 +162,7 @@ kind-load-bundles: ## Load the e2e testdata container images into a kind cluster
 	${KIND} load docker-image testdata/bundles/plain-v0:invalid-crds-and-crs --name $(KIND_CLUSTER_NAME)
 	${KIND} load docker-image testdata/bundles/plain-v0:subdir --name $(KIND_CLUSTER_NAME)
 
-kind-load: ## Load-image loads the currently constructed image onto the cluster
+kind-load: ## Loads the currently constructed image onto the cluster
 	${KIND} load docker-image $(IMAGE) --name $(KIND_CLUSTER_NAME)
 
 ################
--- a/cmd/crdvalidator/main.go
+++ b/cmd/crdvalidator/main.go
@@ -1,0 +1,75 @@
+/*
+Copyright 2022.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"os"
+
+	"github.com/operator-framework/rukpak/cmd/crdvalidator/handlers"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+)
+
+var (
+	scheme   = runtime.NewScheme()
+	entryLog = log.Log.WithName("crdvalidator")
+)
+
+const defaultCertDir = "/etc/admission-webhook/tls"
+
+func init() {
+	if err := apiextensionsv1.AddToScheme(scheme); err != nil {
+		entryLog.Error(err, "unable to set up crd scheme")
+		os.Exit(1)
+	}
+}
+
+func main() {
+	// Setup a Manager
+	entryLog.Info("setting up manager")
+	mgr, err := manager.New(config.GetConfigOrDie(), manager.Options{Scheme: scheme})
+	if err != nil {
+		entryLog.Error(err, "unable to set up overall controller manager")
+		os.Exit(1)
+	}
+
+	entryLog.Info("setting up webhook server")
+	hookServer := mgr.GetWebhookServer()
+
+	// Point to where cert-mgr is placing the cert
+	hookServer.CertDir = defaultCertDir
+
+	// Register CRD validation handler
+	entryLog.Info("registering webhooks to the webhook server")
+	crdValidatorHandler := handlers.NewCrdValidator(entryLog, mgr.GetClient())
+	hookServer.Register("/validate-crd", &webhook.Admission{
+		Handler: &crdValidatorHandler,
+	})
+
+	entryLog.Info("starting manager")
+	if err := mgr.Start(signals.SetupSignalHandler()); err != nil {
+		entryLog.Error(err, "unable to run manager")
+		os.Exit(1)
+	}
+}
--- a/test/e2e/crdvalidator_test.go
+++ b/test/e2e/crdvalidator_test.go
@@ -1,0 +1,198 @@
+package e2e
+
+import (
+	"context"
+	"fmt"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/operator-framework/rukpak/internal/util"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	defaultTestingCrdName  = "samplecrd"
+	defaultTestingCrdGroup = "e2e.io"
+)
+
+var _ = Describe("crd validation webhook", func() {
+	When("a crd event is emitted", func() {
+		var ctx context.Context
+
+		BeforeEach(func() { ctx = context.Background() })
+		AfterEach(func() { ctx.Done() })
+
+		When("an incoming crd event is safe", func() {
+			var crd *apiextensionsv1.CustomResourceDefinition
+
+			BeforeEach(func() {
+				crd = newTestingCRD(
+					util.GenName(defaultTestingCrdName),
+					defaultTestingCrdGroup,
+					[]apiextensionsv1.CustomResourceDefinitionVersion{
+						{
+							Name:    "v1alpha1",
+							Served:  true,
+							Storage: true,
+							Schema: &apiextensionsv1.CustomResourceValidation{
+								OpenAPIV3Schema: &apiextensionsv1.JSONSchemaProps{
+									Type:        "object",
+									Description: "my crd schema",
+								},
+							},
+						},
+					},
+				)
+
+				Eventually(func() error {
+					return c.Create(ctx, crd)
+				}).Should(Succeed(), "should be able to create a safe crd but was not")
+			})
+
+			AfterEach(func() {
+				By("deleting the testing crd")
+				Expect(c.Delete(ctx, crd)).To(BeNil())
+			})
+
+			It("should allow the crd update event to occur", func() {
+				Eventually(func() error {
+					if err := c.Get(ctx, client.ObjectKeyFromObject(crd), crd); err != nil {
+						return err
+					}
+
+					crd.Spec.Versions[0].Storage = false
+
+					crd.Spec.Versions = append(crd.Spec.Versions, apiextensionsv1.CustomResourceDefinitionVersion{
+						Name:    "v1alpha2",
+						Served:  true,
+						Storage: true,
+						Schema: &apiextensionsv1.CustomResourceValidation{
+							OpenAPIV3Schema: &apiextensionsv1.JSONSchemaProps{
+								Type:        "object",
+								Description: "my crd schema",
+							},
+						},
+					})
+
+					return c.Update(ctx, crd)
+				}).Should(Succeed())
+			})
+		})
+
+		// TODO (tylerslaton): Check CRDValidator safe storage logic
+		//
+		// This test is currently trying to simulate a situtation where an incoming
+		// CRD removes a stored version. However, it does not work as expected because
+		// something (potentially the apiserver) is intervening first and not allowing
+		// it through. This is fine and ultimately what the safe storage logic of the
+		// CRDValidator was designed to prevent but is unknown why it is occurring. We
+		// Should come back to this test case, figure out what is preventing it from
+		// hitting the webhook and decide if we want to keep that logic or not.
+		PWhen("an incoming crd event removes a stored version", func() {
+			var crd *apiextensionsv1.CustomResourceDefinition
+
+			BeforeEach(func() {
+				crd = newTestingCRD(
+					util.GenName(defaultTestingCrdName),
+					defaultTestingCrdGroup,
+					[]apiextensionsv1.CustomResourceDefinitionVersion{
+						{
+							Name:    "v1alpha1",
+							Served:  true,
+							Storage: true,
+							Schema: &apiextensionsv1.CustomResourceValidation{
+								OpenAPIV3Schema: &apiextensionsv1.JSONSchemaProps{
+									Type:        "object",
+									Description: "my crd schema",
+								},
+							},
+						},
+						{
+							Name:    "v1alpha2",
+							Served:  true,
+							Storage: false,
+							Schema: &apiextensionsv1.CustomResourceValidation{
+								OpenAPIV3Schema: &apiextensionsv1.JSONSchemaProps{
+									Type:        "object",
+									Description: "my crd schema",
+								},
+							},
+						},
+					},
+				)
+
+				Eventually(func() error {
+					return c.Create(ctx, crd)
+				}).Should(Succeed(), "should be able to create a safe crd but was not")
+			})
+
+			AfterEach(func() {
+				By("deleting the testing crd")
+				Expect(c.Delete(ctx, crd)).To(BeNil())
+			})
+
+			It("should deny admission", func() {
+				Eventually(func() string {
+					if err := c.Get(ctx, client.ObjectKeyFromObject(crd), crd); err != nil {
+						return err.Error()
+					}
+
+					newCRD := newTestingCRD(
+						crd.Spec.Names.Singular,
+						defaultTestingCrdGroup,
+						[]apiextensionsv1.CustomResourceDefinitionVersion{
+							{
+								Name:    "v1alpha2",
+								Served:  true,
+								Storage: true,
+								Schema: &apiextensionsv1.CustomResourceValidation{
+									OpenAPIV3Schema: &apiextensionsv1.JSONSchemaProps{
+										Type:        "object",
+										Description: "my crd schema",
+									},
+								},
+							},
+							{
+								Name:    "v1alpha3",
+								Served:  true,
+								Storage: false,
+								Schema: &apiextensionsv1.CustomResourceValidation{
+									OpenAPIV3Schema: &apiextensionsv1.JSONSchemaProps{
+										Type:        "object",
+										Description: "my crd schema",
+									},
+								},
+							},
+						},
+					)
+
+					newCRD.SetResourceVersion(crd.ResourceVersion)
+					newCRD.Status.StoredVersions = []string{"v1alpha2"}
+
+					return c.Update(ctx, newCRD).Error()
+				}).Should(ContainSubstring("cannot remove stored versions"))
+			})
+		})
+	})
+})
+
+func newTestingCRD(name, group string, versions []apiextensionsv1.CustomResourceDefinitionVersion) *apiextensionsv1.CustomResourceDefinition {
+	return &apiextensionsv1.CustomResourceDefinition{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: fmt.Sprintf("%v.%v", name, group),
+		},
+		Spec: apiextensionsv1.CustomResourceDefinitionSpec{
+			Scope:    apiextensionsv1.ClusterScoped,
+			Group:    group,
+			Versions: versions,
+			Names: apiextensionsv1.CustomResourceDefinitionNames{
+				Plural:   name,
+				Singular: name,
+				Kind:     name,
+				ListKind: name + "List",
+			},
+		},
+	}
+}
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -35,6 +35,15 @@ builds:
     - amd64
     ldflags:
     - -X {{ .Env.PKG }}.GitCommit={{ .ShortCommit }}
+  - id: crdvalidator
+    main: ./cmd/crdvalidator
+    binary: crdvalidator
+    goos:
+    - linux
+    goarch:
+    - amd64
+    ldflags:
+    - -X {{ .Env.PKG }}.GitCommit={{ .ShortCommit }}
 dockers:
 - image_templates:
   - "{{ .Env.IMAGE_REPO }}:{{ .Tag }}-amd64"
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,6 +4,7 @@ WORKDIR /
 COPY plain plain
 COPY unpack unpack
 COPY core core
+COPY crdvalidator crdvalidator
 
 EXPOSE 8080
 ENTRYPOINT ["/plain"]
--- a/cmd/crdvalidator/handlers/crd.go
+++ b/cmd/crdvalidator/handlers/crd.go
@@ -1,0 +1,75 @@
+/*
+Copyright 2022.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package handlers
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	"github.com/go-logr/logr"
+	"github.com/operator-framework/rukpak/internal/crd"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+// +kubebuilder:webhook:path=/validate-crd,mutating=false,failurePolicy=fail,groups="",resources=customresourcedefinitions,verbs=create;update,versions=v1,name=crd-validation-webhook.io
+
+// CrdValidator houses a client, decoder and Handle function for ensuring
+// that a CRD create/update request is safe
+type CrdValidator struct {
+	log     logr.Logger
+	client  client.Client
+	decoder *admission.Decoder
+}
+
+func NewCrdValidator(log logr.Logger, client client.Client) CrdValidator {
+	return CrdValidator{
+		log:    log.V(1).WithName("crdhandler"), // Default to non-verbose logs
+		client: client,
+	}
+}
+
+// Handle takes an incoming CRD create/update request and confirms that it is
+// a safe upgrade based on the crd.Validate() function call
+func (cv *CrdValidator) Handle(ctx context.Context, req admission.Request) admission.Response {
+	incomingCrd := &apiextensionsv1.CustomResourceDefinition{}
+
+	err := cv.decoder.Decode(req, incomingCrd)
+	if err != nil {
+		message := fmt.Sprintf("failed to decode CRD %q", req.Name)
+		cv.log.V(0).Error(err, message)
+		return admission.Errored(http.StatusBadRequest, fmt.Errorf("%s: %w", message, err))
+	}
+
+	err = crd.Validate(ctx, cv.client, incomingCrd)
+	if err != nil {
+		message := fmt.Sprintf("failed to validate safety of %s for CRD %q: %s", req.Operation, req.Name, err)
+		cv.log.V(0).Info(message)
+		return admission.Denied(message)
+	}
+
+	cv.log.Info("admission allowed for %s of CRD %q", req.Name, req.Operation)
+	return admission.Allowed("")
+}
+
+// InjectDecoder injects a decoder for the CrdValidator.
+func (cv *CrdValidator) InjectDecoder(d *admission.Decoder) error {
+	cv.decoder = d
+	return nil
+}
--- a/internal/crd/crd.go
+++ b/internal/crd/crd.go
@@ -1,4 +1,17 @@
-package util
+/*
+Copyright 2022 The Kubernetes Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package crd
 
 import (
 	"context"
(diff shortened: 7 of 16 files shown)


cmd/crdvalidator/manifests/00_service_account.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  namespace: default
```
exdx, line 4: I'm not sure how I feel about deploying all these resources to the default namespace, generally it's considered best practice not to deploy to the default namespace. 

An alternative is to deploy to the rukpak-system namespace that we are using currently, but we are coupling this webhook to rukpak more in that case. 

WDYT about shipping this with it's own namespace so that all the resources are self-contained in that namespace?

cmd/crdvalidator/internal/handlers/crd.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// Handle takes an incoming CRD create/update request and confirms that it is
+// a safe upgrade based on the crd.Validate() function call
+func (v *CrdValidator) Handle(ctx context.Context, req admission.Request) admission.Response {
+	incomingCrd := &apiextensionsv1.CustomResourceDefinition{}
+
+	err := v.decoder.Decode(req, incomingCrd)
+	if err != nil {
+		return admission.Errored(http.StatusBadRequest, err)
```
exdx, line 43: could we be more descriptive with the error? It's unlikely this should fail, but having something like `fmt.Errorf("failed to decode CRD %q: %s", req.Name, err)` would be more helpful.

cmd/crdvalidator/internal/handlers/crd.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+
+	err = crd.Validate(ctx, v.Client, incomingCrd)
+	if err != nil {
+		return admission.Denied(fmt.Sprintf("failed to validate safety of CRD %s: %v", req.Operation, err))
```
exdx, line 48: nit: I would also maybe add `req.Name` here in the reason string as well, and format the error with a `%s` since we know it's going to be in a string.

cmd/crdvalidator/internal/handlers/crd.go
```diff
@@ -0,0 +1,61 @@
...
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+// +kubebuilder:webhook:path=/validate-crd,mutating=false,failurePolicy=fail,groups="",resources=customresourcedefinitions,verbs=create;update,versions=v1,name=crd-validation-webhook.acme.com
+
+// CrdValidator houses a client, decoder and Handle function for ensuring
+// that a CRD create/update request is safe
+type CrdValidator struct {
```
exdx, line 31: Should we pass the CRDValidator a logger as well to log out the results of the validation? Might be helpful if you're trying to see what the webhook is doing. From what I can tell right now, the validation results aren't being logged anywhere on the server, only sent back to the client.

cmd/crdvalidator/manifests/04_webhook.yaml
```diff
@@ -0,0 +1,41 @@
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingWebhookConfiguration
+metadata:
+  name: crd-validation-webhook # TODO (tylerslaton): do we want this domain?metadata:
+  annotations:
+    cert-manager.io/inject-ca-from: default/crd-validation-webhook-certificate
+webhooks:
+  - name: "crd-validation-webhook.acme.com"
```
exdx, line 8: WDYT about "crd-validation-webhook.rukpak.io"? I know we want to keep this webhook as it's own component, but we are housing it in rukpak initially and that's its origin essentially.

exdx (commented):
Very nice work! This looks really good. 

For an example idea of how an e2e could look like, check out https://github.com/operator-framework/operator-lifecycle-manager/blob/ac3aa278ddc8d59add546bf9e33e583f17d62bb0/test/e2e/crd_e2e_test.go#L115

cmd/crdvalidator/manifests/00_service_account.yaml
tylerslaton: I like the concept of a `crdvalidator-system` (or something in that vein) namespace. I didn't do it to begin with to make sure that was something we wanted to go with but it makes sense. In the theme of not coupling this with Rukpak I'd prefer it to be in its own - would you agree?

cmd/crdvalidator/internal/handlers/crd.go
tylerslaton: Good call, I'll add that.

cmd/crdvalidator/manifests/04_webhook.yaml
tylerslaton: Maybe it makes sense to create its own namespace? Potentially something like `crd-validation-webhook.io`. My main hesitation with doing anything `rukpak` specific to it is that we'll have to undo it later. Granted, its trivial work to fix so I'm fine either way.

cmd/crdvalidator/internal/handlers/crd.go
tylerslaton: Yep, that's a great callout that I will add now.

cmd/crdvalidator/internal/handlers/crd.go
tylerslaton: Great call, on it.

cmd/crdvalidator/manifests/04_webhook.yaml
tylerslaton: Added the new address in the most recent commit.

cmd/crdvalidator/manifests/00_namespace.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: crdvalidator-system
```
akihikokuroda, line 4: I wonder why this webhook is in a separate namespace.  Can it be in the same namespace as other rukpak components

akihikokuroda: I see.  I didn't know the intention.  I just wondered.  Thanks!

cmd/crdvalidator/internal/crd/crd.go
```diff
@@ -0,0 +1,172 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func Validate(ctx context.Context, cl client.Client, newCrd *apiextensionsv1.CustomResourceDefinition) error {
+	oldCRD := &apiextensionsv1.CustomResourceDefinition{}
+	if err := cl.Get(ctx, client.ObjectKeyFromObject(newCrd), oldCRD); err != nil && !apierrors.IsNotFound(err) {
```
perdasilva, line 21:
```suggestion
	if err := client.IgnoreNotFound(cl.Get(ctx, client.ObjectKeyFromObject(newCrd), oldCRD)); err != nil {
```

cmd/crdvalidator/internal/crd/crd.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	return nil
+}
+
+func keys(m map[string]apiextensionsv1.CustomResourceDefinitionVersion) sets.String {
+	return sets.StringKeySet(m)
+}
+
+func validateCRDCompatibility(ctx context.Context, cl client.Client, oldCRD *apiextensionsv1.CustomResourceDefinition, newCRD *apiextensionsv1.CustomResourceDefinition) error {
```
perdasilva, line 45: should we check upfront if oldCRD is "empty" and just return before performing the whole check?

cmd/crdvalidator/internal/crd/crd.go
```diff
@@ -0,0 +1,172 @@
+package crd
```
perdasilva, line 1: License header?

cmd/crdvalidator/internal/crd/crd.go
```diff
@@ -0,0 +1,172 @@
...
+	"k8s.io/apiextensions-apiserver/pkg/apiserver/validation"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func Validate(ctx context.Context, cl client.Client, newCrd *apiextensionsv1.CustomResourceDefinition) error {
```
perdasilva, line 19: I think it would be helpful to have a comment here that explains what the validation does. It seems pretty involved. So having this upfront might help when reading the code for the first time (even if after a long time)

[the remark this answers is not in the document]
tylerslaton: Right now I'm adding an E2E test that will check out a lot of this logic already. I think a unit test would be a good follow-up item from this as it would be pretty involved to add given the amount of mocking/setup needed. Thoughts?

tylerslaton: Agreed, I'll add something more descriptive.

cmd/crdvalidator/internal/crd/crd.go
timflannagan, line 1: Should we move this to the internal/crd and have the cmd/crdvalidator executable package import that?

Makefile
```diff
@@ -152,9 +152,29 @@ kind-load-bundles: ## Load the e2e testdata container images into a kind cluster
...
+CRD_WEBHOOK_IMAGE_REPO=quay.io/operator-framework/crd-validation-webhook
+CRD_WEBHOOK_IMAGE=$(CRD_WEBHOOK_IMAGE_REPO):main
+build-crdvalidator-image: ## Builds crd-validation-webhook container image locally
+	docker build -f Dockerfile.crdvalidator -t $(CRD_WEBHOOK_IMAGE) .
+
+kind-load-crdvalidator: ## Loads the currently constructed crd-validation-webhook onto the cluster
+	${KIND} load docker-image $(CRD_WEBHOOK_IMAGE) --name $(KIND_CLUSTER_NAME)
+
+deploy-crdvalidator: build-crdvalidator-image kind-load-crdvalidator cert-mgr ## Deploy the crdvalidator manifests to the current cluster. Make sure cert-manager is installed prior.
+	kubectl apply -f cmd/crdvalidator/manifests
```
timflannagan, line 176: Do we need to add one of these targets to `make run` in some capacity? Right now, that's the Makefile target we document for quickstart purposes, and reading through these changes, it's unclear to me whether I'd see this functionality present on a local cluster if I pulled down these changes and ran `make run` locally.

cmd/crdvalidator/internal/crd/crd.go
tylerslaton: I have two main reasons that I moved this package into the `cmd/crdvalidator` package:

1. The understanding that eventually this webhook will broken out into its own repository and it requires this package
2. Nothing else but this webhook is using it currently.

With that in mind, I'm fine with moving it back if we feel that something else in Rukpak will eventually use it.

Makefile
exdx: My impression is that we should probably include this is the default rukpak installation. If we introduce a webhook into the codebase we should be the ones running and testing against it. The caveat being there should be an easy way to disable the actual CRD validation checks, which I think we have a ticket for somewhere (this should be a follow-up).

tylerslaton: Okay got it. In that case I'll have this included in `make run` and we can action on that follow up item.

Makefile
```diff
@@ -156,9 +156,36 @@ kind-load-bundles: ## Load the e2e testdata container images into a kind cluster
...
+
+kind-load-crdvalidator: ## Loads the currently constructed crd-validation-webhook onto the cluster
+	${KIND} load docker-image $(CRD_WEBHOOK_IMAGE) --name $(KIND_CLUSTER_NAME)
+
+install-crdvalidator: build-crdvalidator-image kind-load-crdvalidator ## Install the crdvalidator manifests to the current cluster. Requires that cert-manager is installed prior.
+	kubectl apply -f cmd/crdvalidator/manifests
+
+deploy-crdvalidator: cert-mgr install-crdvalidator ## Install the crdvalidator manifests to the current cluster and ensures that cert-manager is installed prior.
+
+FOCUS := $(if $(TEST),-focus "$(TEST)")
```
exdx, line 185: nit: this variable is defined in L80 -- could we use that definition in this context as well?

Makefile
```diff
@@ -156,9 +156,36 @@ kind-load-bundles: ## Load the e2e testdata container images into a kind cluster
...
+test-crdvalidator-e2e: ginkgo ## Run the crdvalidator e2e tests. Assumes you have already installed the crdvalidator webhook.
+	$(GINKGO) -v -trace -progress $(FOCUS) cmd/crdvalidator/test/e2e
```
exdx, line 187: nit: we decided to remove the verbosity from the default e2e runs, and moved the `-v` to only the case where the test is specified via `$(FOCUS)`

cmd/crdvalidator/internal/crd/crd.go
```diff
@@ -0,0 +1,188 @@
...
+)
+
+// Validate is a wrapper for doing four things:
+// 	1. Retrieving the existing version of the specified CRD where it exists.
+// 	2. Calling validateCRDCompatibility() on the newCrd.
+// 	3. Calling safeStorageVersionUpgrade() on the newCrd.
+// 	4. Reporting any errors that it encounters along the way.
+func Validate(ctx context.Context, cl client.Client, newCrd *apiextensionsv1.CustomResourceDefinition) error {
+	oldCRD := &apiextensionsv1.CustomResourceDefinition{}
+	if err := client.IgnoreNotFound(cl.Get(ctx, client.ObjectKeyFromObject(newCrd), oldCRD)); err != nil {
```
exdx, line 38: By ignoring not found errors, are we skipping the case where there is no prior version of the CRD on-cluster? 
We could have the following and return early:
```
err := cl.Get(ctx, client.ObjectKeyFromObject(newCrd), oldCRD)
if apierrors.IsNotFound(err) {
// no prior version of the CRD
// safe to install new version
  return nil
} 
```

cmd/crdvalidator/manifests/01_cluster_role.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: ClusterRole
+metadata:
+  name: crd-validation-webhook
+rules:
+- apiGroups: ["apiextensions.k8s.io"]
+  resources: ["customresourcedefinitions"]
+  verbs: ["get", "watch", "list"]
+- apiGroups: ["*"]
+  resources: ["*"]
+  verbs: ["list"]
```
exdx, line 11: nit: new-line

cmd/crdvalidator/test/e2e/crdvalidator_test.go
```diff
@@ -0,0 +1,150 @@
...
+	"k8s.io/apiserver/pkg/storage/names"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ = Describe("crd validation webhook", func() {
+	When("a crd event is emitted", func() {
+		var ctx context.Context
+
+		BeforeEach(func() { ctx = context.Background() })
+		AfterEach(func() { ctx.Done() })
```
exdx, line 21: 👍

cmd/crdvalidator/internal/crd/crd.go
```diff
@@ -0,0 +1,188 @@
...
+	return sets.StringKeySet(m)
+}
+
+// validateCRDCompatibility runs through the following cases to test:
+//   1. New CRD removes version that Old CRD had => Must ensure nothing is stored at removed version
+//   2. New CRD changes a version that Old CRD has => Must validate existing CRs with new schema
+//   3. New CRD adds a version that Old CRD does not have =>
+//      - If conversion strategy is None, ensure existing CRs validate with new schema.
+//      - If conversion strategy is Webhook, allow update (assume webhook handles conversion correctly)
+func validateCRDCompatibility(ctx context.Context, cl client.Client, oldCRD *apiextensionsv1.CustomResourceDefinition, newCRD *apiextensionsv1.CustomResourceDefinition) error {
```
exdx, line 68: nit: all the code in this function and following it is fairly in-depth, and could use some unit tests as a follow-up. Since it came from elsewhere, largely from OLM, it's fairly safe to rely on, but we could test it more on the rukpak side as a follow-up.

tylerslaton: I agree, I'll make a follow-up item for this PR that adds unit testing for this particular package.

Makefile
tylerslaton: Ah right, good catch.

Dockerfile.crdvalidator
```diff
@@ -0,0 +1,23 @@
...
+# cache deps before building and copying source so that we don't need to re-download as much
+# and so that source changes don't invalidate our downloaded layer
+RUN go mod download
+
+COPY Makefile Makefile
+COPY cmd/crdvalidator cmd/crdvalidator
+
+RUN make bin/crdvalidator
+
+FROM gcr.io/distroless/static:debug
```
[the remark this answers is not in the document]
tylerslaton: 🤔  It doesn't, you're right. I was mainly copying this over from the root Dockerfile. Do you think that means we should be moving that container away from having a shell as well?

cmd/crdvalidator/internal/crd/crd.go
tylerslaton: Hm, that's a really good point. If the CRD is not on the cluster there is no real need to go through and try to validate it.

exdx (approved):
Looks great, thanks for all the work on this. Let's get one more approval before merging.

Makefile
```diff
@@ -156,9 +156,35 @@ kind-load-bundles: ## Load the e2e testdata container images into a kind cluster
...
 
+##################################
+# CRDValidator Webhook Targets #
+##################################
+.PHONY: bin/crdvalidator build-crdvalidator-image kind-load-crdvalidator install-crdvalidator deploy-crdvalidator
+
+##@ crdvalidator:
+
+bin/crdvalidator: 
+	CGO_ENABLED=0 go build $(VERSION_FLAGS) -o $@$(BIN_SUFFIX) ./cmd/crdvalidator
```
timflannagan, line 170: Should we make this kind of target consistent with the plain provisioner + core webhook listed above?

```makefile
crdvalidator:
	CGO_ENABLED=0 go build $(VERSION_FLAGS) -o $(BIN_DIR)/$@ ./cmd/crdvalidator
```

cmd/crdvalidator/internal/handlers/crd.go
timflannagan: CrdValidator -> CRDValidator?

cmd/crdvalidator/internal/handlers/crd.go
```diff
@@ -0,0 +1,69 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+// +kubebuilder:webhook:path=/validate-crd,mutating=false,failurePolicy=fail,groups="",resources=customresourcedefinitions,verbs=create;update,versions=v1,name=crd-validation-webhook.io
+
+// CrdValidator houses a client, decoder and Handle function for ensuring
+// that a CRD create/update request is safe
+type CrdValidator struct {
+	Log     logr.Logger
```
timflannagan, line 36: Does this need to be exported?

cmd/crdvalidator/internal/handlers/crd.go
```diff
@@ -0,0 +1,69 @@
...
+	Client  client.Client
```
timflannagan, line 37: Does this need to be exported?

cmd/crdvalidator/internal/handlers/crd.go
```diff
@@ -0,0 +1,69 @@
...
+
+	log.Info("admission allowed for %s of CRD %q", req.Name, req.Operation)
+	return admission.Allowed("")
+}
+
+// InjectDecoder injects a decoder for the CrdValidator.
+func (v *CrdValidator) InjectDecoder(d *admission.Decoder) error {
+	v.decoder = d
+	return nil
+}
```
timflannagan, line 69: How do we feel about an options pattern implementation, .e.g `v.WithDecoder(d)` here?

cmd/crdvalidator/main.go
```diff
@@ -0,0 +1,74 @@
...
+	if err != nil {
+		entryLog.Error(err, "unable to set up overall controller manager")
+		os.Exit(1)
+	}
+
+	entryLog.Info("setting up webhook server")
+	hookServer := mgr.GetWebhookServer()
+
+	// Point to where cert-mgr is placing the cert
+	hookServer.CertDir = defaultCertDir
```
timflannagan, line 55: Any idea on whether this needs to be configurable?

cmd/crdvalidator/main.go
```diff
@@ -0,0 +1,74 @@
...
+
+	// Need manager to have scheme of CRDs
+	utilruntime.Must(apiextensionsv1.AddToScheme(mgr.GetScheme()))
```
timflannagan, line 58: nit: Can we setup the scheme before instantiating the manager instance, and handle the error returned vs. relying on the utilruntime.Must (which I believe panics under-the-hood). You would then be able to pass the built up scheme variable to manager.New in the manager.Options (e.g. manager.Options{ Scheme: scheme })

cmd/crdvalidator/manifests/00_namespace.yaml
timflannagan: I think I'd prefer landing with rukpak-system until we have a concrete want/need to move this outside of rukpak but I won't hold this up while we continue iterating.

cmd/crdvalidator/test/e2e/crdvalidator_test.go
```diff
@@ -0,0 +1,150 @@
...
+
+		When("incoming crd event is safe", func() {
```
timflannagan, line 23: nit: When("an incoming CRD event is safe", ...)

cmd/crdvalidator/test/e2e/crdvalidator_test.go
```diff
@@ -0,0 +1,150 @@
...
+					}
+
+					err := c.Create(ctx, crd)
+					if err == nil {
+						return false
+					}
+
+					return errorsLooselyEqual(err, errors.New("cannot remove stored versions"))
+				}).Should(BeTrue())
+			})
```
timflannagan, line 113: A couple of things:

- I'm slightly confused whether we should be firing off an update request vs. a create one
- Can we return an error here and use gomega matchers to enforce the expected error message, e.g. https://github.com/operator-framework/rukpak/blob/bf174efdd76fb100a0d818cbc895f49b7f85b440/test/e2e/plain_provisioner_test.go#L838

cmd/crdvalidator/test/e2e/crdvalidator_test.go
```diff
@@ -0,0 +1,150 @@
...
+import (
+	"context"
+	"errors"
+	"strings"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apiserver/pkg/storage/names"
```
timflannagan, line 12: Note: this is probably way too early in this project's lifecycle to care about this - can we use another package for our CRD naming needs so we can eliminate a dependency from our root module?

go.mod
```diff
@@ -16,6 +16,7 @@ require (
 	k8s.io/api v0.23.1
 	k8s.io/apiextensions-apiserver v0.23.1
 	k8s.io/apimachinery v0.23.1
+	k8s.io/apiserver v0.23.1
```
timflannagan, line 19: Same comment in the CRD validator e2e suite: can we eliminate this dependency (maybe in a follow-up?)?

Dockerfile.crdvalidator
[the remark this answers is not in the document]
timflannagan: I think @exdx brings up a good point but I wonder whether it's good to use a debug tag in the short term while we continue to iterate, and then later on work towards removing it.

cmd/crdvalidator/test/e2e/crdvalidator_test.go
```diff
@@ -0,0 +1,150 @@
...
+				err := c.Create(ctx, crd)
+				Expect(err).To(BeNil())
+
+				Eventually(func() error {
+					return c.Get(ctx, client.ObjectKeyFromObject(crd), crd)
+				}).Should(Succeed(), "should be able to create a safe crd but was not")
+			})
+
+			AfterEach(func() {
+				By("deleting the testing Bundle resource")
```
timflannagan, line 38: s/Bundle/CRD

cmd/crdvalidator/test/e2e/crdvalidator_test.go
```diff
@@ -0,0 +1,150 @@
...
```
timflannagan, line 82: s/Bundle/CRD

timflannagan (commented):
Okay I just did a quick passthrough of these changes. I think my main takeaways were the following:

- I'm not a huge fan of the current directory structure as it's difficult to discover where packages live in what directories and we don't have a concrete want/need to split this kind of component into it's own repository, or attempt to introduce it to upstream k/k.
- The `make run` command doesn't install this component like I would've expected
- It's not entirely evident how we could disable this behavior dynamically (or whether that's something we'd need to support)
- We can punt on the debug/static distroless tag discussion
- I think we can drop that k8s.io/apiserver dependency that was introduced with these changes
- We're potentially introducing a confusing installation UX by having a deploy and install Makefile targets
- It doesn't look like the CRD validator e2e test Makefile target loads the kind-load-crdvalidator target unless I'm reading these changes wrong
- We'll need to update goreleaser to build this new executable package

I also ran these changes locally and ran into e2e failures. I haven't dived too deep into those yet though.

timflannagan:
FYI - It looks like I ran into a container image build warning locally too:

```
Step 3/14 : COPY go.mod go.mod
 ---> 73952c5724e3
Step 4/14 : COPY go.sum go.sum
 ---> 358d31c7b255
Step 5/14 : RUN go mod download
 ---> Running in eeb714ec03fd
Removing intermediate container eeb714ec03fd
 ---> d370149889ea
Step 6/14 : COPY Makefile Makefile
 ---> ddfb93c559ff
Step 7/14 : COPY cmd/crdvalidator cmd/crdvalidator
 ---> 1ba657d6e072
Step 8/14 : RUN make bin/crdvalidator
 ---> Running in 49439f4699bf
fatal: not a git repository (or any of the parent directories): .git
CGO_ENABLED=0 go build -ldflags "-X github.com/operator-framework/rukpak/internal/version.GitCommit=" -o bin/crdvalidator ./cmd/crdvalidator
```

tylerslaton:
> Okay I just did a quick passthrough of these changes. I think my main takeaways were the following:
> 
> * The `make run` command doesn't install this component like I would've expected
> * It doesn't look like the CRD validator e2e test Makefile target loads the kind-load-crdvalidator target unless I'm reading these changes wrong
> 
> I also ran these changes locally and ran into e2e failures. I haven't dived too deep into those yet though.

Yep, so I recently rebased this PR and it removed my addition to `make run` that had the `crdvalidator` being deployed. This resulted in the `crdvalidators` e2e suite not passing as a side effect. I'll work to address the other issues you listed.

cmd/crdvalidator/main.go
tylerslaton: I suppose we could set it via an environment variable and that would make sense. That way if there is a non-default configuration of the `cert-manager` then a user can edit it easily. Thoughts?

cmd/crdvalidator/test/e2e/crdvalidator_test.go
tylerslaton: Yeah so the `Update` here was attempted and continuously failed. If you have the `Create` it succeeds - however, this is not what we should really be expecting and just obfuscated the issue.  Something under-the-hood in Kube is now blocking unsafe storage removals and thus the error that we are expecting is not coming through. This is because the `Update` request does not seem to even hit the webhook and Kube returns its own error. I'm making some changes in an upcoming commit along with a `todo` comment to address this in the future by reevaluating if we need to ensure safe storage removals or if Kube does it by default now.

cmd/crdvalidator/internal/handlers/crd.go
tylerslaton: Yes as it is being set outside of this package in the `main` for the `crdvalidator`. Thoughts around that?

cmd/crdvalidator/internal/handlers/crd.go
tylerslaton: So the `InjectDecoder` actually implements an admission.DecoderInjector so that the decoder gets automatically injected. I got it from the `controller-runtime` example found [here](https://github.com/kubernetes-sigs/controller-runtime/blob/188b48630db275131d0fef1818933d8c2da96753/examples/builtins/validatingwebhook.go#L58).

cmd/crdvalidator/main.go
tylerslaton: So the newest commit will be adding this fix. One thing to ask is if we want to stop using `utilruntime.Must` inside of the `plain` provisioner as well for a follow-up:

https://github.com/operator-framework/rukpak/blob/455aead0f9a3687bdee334d27483ac23d1219a4e/internal/provisioner/plain/main.go#L50-L52

cmd/crdvalidator/internal/handlers/crd.go
timflannagan: You could just create a simple constructor for this type then, e.g. NewCRDValidator which takes a client.Client parameter and sets the internal client struct field.

cmd/crdvalidator/internal/handlers/crd.go
timflannagan: Alright if there's already an existing pattern in the community, then sticking with that is fine with me.

cmd/crdvalidator/main.go
timflannagan: Yep that's what I had in mind but this can be a follow-up anyways.

timflannagan: Maybe we don't derive this through an environment variable though, but we can hash that outside of this PR.

cmd/crdvalidator/main.go
timflannagan: Yeah that seems like a reasonable cleanup PR in my eyes.

cmd/crdvalidator/internal/handlers/crd.go
tylerslaton: Makes sense, I was thinking of that as well 👍🏻 Let me get that typed up.

Makefile
```diff
@@ -156,9 +156,38 @@ kind-load-bundles: ## Load the e2e testdata container images into a kind cluster
...
+
+install-crdvalidator: build-crdvalidator-image kind-load-crdvalidator ## Install the crdvalidator manifests to the current cluster. Requires that cert-manager is installed prior.
+	kubectl apply -f cmd/crdvalidator/manifests
+
+deploy-crdvalidator: cert-mgr install-crdvalidator ## Install the crdvalidator manifests to the current cluster and ensures that cert-manager is installed prior.
+
+test-crdvalidator-e2e: ginkgo ## Run the crdvalidator e2e tests. Assumes you have already installed the crdvalidator webhook.
+	$(GINKGO) -trace -progress $(FOCUS) cmd/crdvalidator/test/e2e
+
+crdvalidator-e2e: KIND_CLUSTER_NAME=crdvalidator-e2e
```
timflannagan, line 188: This seems problematic at a first glance, but it's possible I'm reading this wrong: I don't think we create this kind cluster anywhere in the target prerequisite chain. I also don't think we need to separate out kind clusters when running the plain and crdvalidator e2e suites, at least for right now.

cmd/crdvalidator/test/e2e/crdvalidator_test.go
```diff
@@ -0,0 +1,197 @@
...
+				}).Should(ContainSubstring("cannot remove stored versions"))
+			})
+		})
+	})
+})
+
+func newTestingCRD(name string, versions []apiextensionsv1.CustomResourceDefinitionVersion) *apiextensionsv1.CustomResourceDefinition {
+	return &apiextensionsv1.CustomResourceDefinition{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: name + ".e2e.io",
```
timflannagan, line 183: nit: I think this is slightly confusing from the consumer aspect where I pass an explicit name parameter, but it gets mutated under-the-hood. Maybe we should pass an explicit group parameter too?

tylerslaton: Yeah, that's a fair improvement.

Makefile
```diff
@@ -83,7 +83,7 @@ test-e2e: ginkgo ## Run the e2e tests
 	$(GINKGO) -trace -progress $(FOCUS) test/e2e
 
 e2e: KIND_CLUSTER_NAME=rukpak-e2e
-e2e: build-container kind-cluster kind-load cert-mgr kind-load-bundles deploy test-e2e ## Run e2e tests against a kind cluster
+e2e: build-container kind-cluster kind-load cert-mgr kind-load-bundles deploy test-e2e test-crdvalidator-e2e ## Run e2e tests against a kind cluster
```
timflannagan, line 86: It doesn't look like `test-crdvalidator-e2e` exists anymore?

timflannagan (commented):
I think we're super close and then I'm ready to approve:

- Can we remove that test-crdvalidator-e2e target which appears to be unused?
- Can we remove that unused Dockerfile.crdvalidator now that we've consolidated this component in the current repository structure?

timflannagan (approved):

timflannagan:
Note: it looks like this branch references out-of-date files compared to the main branch. It might be worth rebasing against the main branch, and respinning e2e once more to make sense we're not potentially silently regressing, but I'll leave that up to you as I just approved this.

timflannagan:
Really nice work btw @tylerslaton

tylerslaton:
> Note: it looks like this branch references out-of-date files compared to the main branch. It might be worth rebasing against the main branch, and respinning e2e once more to make sense we're not potentially silently regressing, but I'll leave that up to you as I just approved this.

Rebased and pushed. Looks like everything is good to go. Going to squash all the commits together. @exdx anything you want to touch on before we merge this?

Makefile
```diff
@@ -102,9 +102,12 @@ install-apis: cert-mgr generate kustomize ## Install the core rukpak CRDs
 install-plain: install-apis ## Install the rukpak CRDs and the plain provisioner
 	$(KUSTOMIZE) build manifests/provisioners/plain | kubectl apply -f -
 
+install-crdvalidator: cert-mgr kustomize ## Install the crdvalidator webhook manifests to the current cluster.
+	$(KUSTOMIZE) build manifests/crdvalidator | kubectl apply -f -
+
 install: install-plain ## Install all rukpak core CRDs and provisioners
 
-deploy: install-apis ## Deploy the operator to the current cluster
+deploy: install-apis install-crdvalidator ## Deploy the operator to the current cluster
```
exdx, line 110: nit: `install-apis` already calls `cert-mgr` so we end up installing cert-manager twice on deploy. I'm not sure a good way around this -- is there a way to only run a target if it hasn't already been run?

test/e2e/crdvalidator_test.go
```diff
@@ -0,0 +1,198 @@
...
+
+		// TODO (tylerslaton): Check CRDValidator safe storage logic
+		//
+		// This test is currently trying to simulate a situtation where an incoming
+		// CRD removes a stored version. However, it does not work as expected because
+		// something (potentially the apiserver) is intervening first and not allowing
+		// it through. This is fine and ultimately what the safe storage logic of the
+		// CRDValidator was designed to prevent but is unknown why it is occurring. We
+		// Should come back to this test case, figure out what is preventing it from
+		// hitting the webhook and decide if we want to keep that logic or not.
```
exdx, line 92: can we open an issue to track this and put the link to it here in the test comment?

test/e2e/crdvalidator_test.go
```diff
@@ -0,0 +1,198 @@
...
+		PWhen("an incoming crd event removes a stored version", func() {
```
exdx, line 93: I know we had some issues getting this test to work -- but we did confirm the test case where the schema changed in an incompatible way between upgrades. It's worth adding that test so we have at least one scenario where the admission was denied due to the webhook in our tests. This can be added in a follow-up though.

exdx (commented):
Looks great, just some small comments.

test/e2e/crdvalidator_test.go
timflannagan: Yep - good call out. I'm also cool with a follow-up here.

tylerslaton: Yep I'm planning on doing a follow-up PR like the recent git work after this lands.

Makefile
tylerslaton: I was also thinking on this. I haven't come up with a great solution other than having all the cert-mgr dependent installs in a single target.

exdx: That sounds reasonable, we can have something like `install-webhooks` which calls `cert-mgr` under the hood. I suppose we can address this is in a follow-up

tylerslaton: Totally, but I agree with you. Its an easy addition that I can just tack onto the follow-ups.

test/e2e/crdvalidator_test.go
tylerslaton: I believe that this action will create a issue for it:

https://github.com/operator-framework/rukpak/blob/main/.github/workflows/todo.yaml

Are we okay with that?

timflannagan:
LGTM again - do you want to push this through @tylerslaton.

PR #217, opened 2022-04-04
api union type validation

### LINKED ISSUE #216: Introduce CRD schema validation for the union-based Bundle API
Examine how to possibly incorporate CRD-schema level validation for the union type semantics exposed through the new source field. If not possible, move git-based bundle level validation into the bundle webhook.

akihikokuroda:

This is for latter of https://github.com/operator-framework/rukpak/issues/216 in case the former way doesn't work well.  When the former works, the e2e tests may still be useful.

Closes #216


exdx:
Thanks for this @akihikokuroda ... I think the schema validation is still worth looking into, but it is challenging and this seems like a fine intermediate solution IMO. Let's see where that ends up and then we can potentially merge this.

joelanford:
I agree a webhook implementation is better than no schema checks, so +1 on merging this.

However, I think we should try to make use of the documented CRD schema validations for union types because that will:
a) give us more confidence that we implemented the schema type correctly.
b) make our API and the responses sent by the API server in case in validation failure consistent for users who expect kube APIs to follow kube API standards.

akihikokuroda:
I believe that the CRD validation will be done by adding the `x-kubernetes-validations` in the `openAPIV3Schema.  This feature is still **alpha** and in `CustomResourceValidationExpressions` feature gate.
```
                        ref:
                          description: Ref configures the git source to clone a specific branch, tag, or commit from the specified repo. Ref is required, and exactly one field within Ref is required. Setting more than one field or zero fields will result in an error.
                          x-kubernetes-validations:
                            - rule: "(has(self.branch)? 1:0 + has(self.commit)? 1:0 + has(self.tag)? 1:0) == 1"
                          properties:
                            branch:
                              description: Branch refers to the branch to checkout from the repository. The Branch should contain the bundle manifests in the specified directory.
                              type: string
                            commit:
                              description: Commit refers to the commit to checkout from the repository. The Commit should contain the bundle manifests in the specified directory.
                              type: string
                            tag:
                              description: Tag refers to the tag to checkout from the repository. The Tag should contain the bundle manifests in the specified directory.
                              type: string 
```

joelanford:
We shouldn't depend on an alpha feature. For the `git.ref` field, is it possible to use `oneOf`?

For the `source` field, I'd expect to use what's specified here: https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/1027-api-unions/README.md

akihikokuroda:
It seems possible.  I'll work on it.

akihikokuroda:
I couldn't make the `oneOf` work.

akihikokuroda:
This PR has:

1. adding CRD validation in the generated bundles.yaml using Kustomize when it is deployed.
2. validating webhook for source type and property mismatch
3. removing the bundle name check from the validating webhook
4. e2e test for CRD validation

test/e2e/webhook_test.go
```diff
@@ -77,11 +76,123 @@ var _ = Describe("bundle api validating webhook", func() {
 		})
 		AfterEach(func() {
 			By("deleting the testing Bundle resource for failure case")
-			err := c.Delete(ctx, bundle)
-			Expect(err).To(WithTransform(apierrors.IsNotFound, BeTrue()))
+			err = c.Delete(ctx, bundle)
+			Expect(err).ToNot(BeNil())
```
joelanford, line 80: Can we assert on something specific in the error?

Generic `err != nil` assertions could keep passing if we make an unwanted change that results in a different error.

manifests/kustomization.yml
```diff
@@ -0,0 +1,10 @@
+resources:
+- core.rukpak.io_bundles.yaml
+- core.rukpak.io_bundleinstances.yaml
+patches:
+- path: crdvalidation.yaml
+  target:
+    group: apiextensions.k8s.io
+    version: v1
+    kind: CustomResourceDefinition
+    name: bundles.core.rukpak.io
```
exdx, line 10: nit: newline

manifests/crdvalidation.yaml
```diff
@@ -0,0 +1,28 @@
+# limit bundle name length up to 40
+- op: add
```
exdx, line 2: Very cool -- this is exactly what I was looking to implement -- looks great.

api/v1alpha1/bundle_webhook.go
```diff
@@ -62,9 +59,16 @@ func (r *Bundle) ValidateDelete() error {
 	return nil
 }
 
-func checkNameLength(r *Bundle) error {
-	if len(r.Name) > maxNameLength {
-		return fmt.Errorf("bundle name %s is too long: maximum allowed name length is %d", r.GetName(), maxNameLength)
+func checkBundleSource(r *Bundle) error {
```
joelanford, line 62: Were you able to try the union type validations described here?

Wondering if this PR is a stepping stone before that or if there's an issue using those in the CRD schema?

exdx (approved):

api/v1alpha1/bundle_webhook.go
akihikokuroda: Yes, the union type validation works.  e2e testing is verifying that the only one of the union items can be specified.

manifests/kustomization.yml
akihikokuroda: Thanks!  I fix it.

test/e2e/webhook_test.go
akihikokuroda: I'll put the `Expect(err).To(WithTransform(apierrors.IsNotFound, BeTrue()))` back.  Thanks!

api/v1alpha1/bundle_webhook.go
joelanford: Sorry, I meant to paste this link again: [https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/1027-api-unions/README.md](https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/1027-api-unions/README.md?rgh-link-date=2022-04-05T20%3A30%3A46Z)

I tried to track down the implementation of the union validations and it looks like that KEP's work fizzled out and didn't merge.

manifests/crdvalidation.yaml
joelanford: As a follow-up, I think it'd be nice to organize all the manifests a bit better (I'm partial to the way kubebuilder/operator-sdk has `config/<component>` directories and then a `config/default` that ties everything together.

api/v1alpha1/bundle_webhook.go
joelanford: Actually, I'm still unsure. It seems like the PR that didn't merge (https://github.com/kubernetes/kubernetes/pull/77370) maybe was just the implementation of that KEP against the the kubernetes core APIs, but perhaps we could still use the CRD extension schema called ([x-kubernetes-unions](https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/1027-api-unions/README.md#openapi))

manifests/crdvalidation.yaml
akihikokuroda: OK.  I'll open a follow up issue for it.

api/v1alpha1/bundle_webhook.go
akihikokuroda: I tried but the `x-kubernetes-unions` does not work.  

I got this message 
> unknown field "x-kubernetes-unions" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.JSONSchemaProps;

It is not documented in the kubernetes docs, either.

manifests/crdvalidation.yaml
akihikokuroda: https://github.com/operator-framework/rukpak/issues/228

api/v1alpha1/bundle_webhook.go
joelanford: Thanks for looking into it! That's a bummer.

joelanford (approved):
/lgtm

Thanks!

PR #218, opened 2022-04-04
e2e: fix duplicate BI e2e test flake

### LINKED ISSUE #200: [Flake] a BundleInstance targets a valid bundle but the bundle contains resources that already exist
Noticed a flake in CI recently:
```
• [FAILED] [60.108 seconds]
plain provisioner bundleinstance
/home/runner/work/rukpak/rukpak/test/e2e/plain_provisioner_test.go:464
  a BundleInstance targets a valid bundle but the bundle contains resources that already exist
  /home/runner/work/rukpak/rukpak/test/e2e/plain_provisioner_test.go:790
    [It] should succeed installation for the original but fail for the duplicate BundleInstance
    /home/runner/work/rukpak/rukpak/test/e2e/plain_provisioner_test.go:865
```

exdx:
This PR updates the duplicate BI e2e test to ensure that the original BI is always the owner of the provided Bundle, to prevent any flakes where the duplicate bundle ends up owning the Bundle before the original.

Closes #200

### DIFF
--- a/test/e2e/plain_provisioner_test.go
+++ b/test/e2e/plain_provisioner_test.go
@@ -882,9 +882,27 @@ var _ = Describe("plain provisioner bundleinstance", func() {
 
 			err = c.Create(ctx, biOriginal)
 			Expect(err).To(BeNil(), "failed to create original bundle instance")
-			Eventually(func() error { // Guarantee that the original bundle instance is create prior to the duplicate
-				return c.Get(ctx, client.ObjectKeyFromObject(biOriginal), &rukpakv1alpha1.BundleInstance{})
-			}).Should(Succeed(), "failed to validate original bundle instance was created in time")
+
+			// ensure the original BI owns the underlying bundle before creating the duplicate
+			By("projecting a successful installation status for the original BundleInstance")
+			Eventually(func() bool {
+				if err := c.Get(ctx, client.ObjectKeyFromObject(biOriginal), biOriginal); err != nil {
+					return false
+				}
+				if biOriginal.Status.InstalledBundleName != bundle.GetName() {
+					return false
+				}
+				existing := meta.FindStatusCondition(biOriginal.Status.Conditions, rukpakv1alpha1.TypeInstalled)
+				if existing == nil {
+					return false
+				}
+				expected := metav1.Condition{
+					Type:   rukpakv1alpha1.TypeInstalled,
+					Status: metav1.ConditionStatus(corev1.ConditionTrue),
+					Reason: rukpakv1alpha1.ReasonInstallationSucceeded,
+				}
+				return conditionsSemanticallyEqual(*existing, expected)
+			}).Should(BeTrue())
 
 			biDuplicate = &rukpakv1alpha1.BundleInstance{
 				ObjectMeta: metav1.ObjectMeta{
@@ -901,9 +919,9 @@ var _ = Describe("plain provisioner bundleinstance", func() {
 		})
 
 		AfterEach(func() {
-			By("deleting the testing Bundle resource")
+			By("deleting the testing duplicate BundleInstance resource")
 			Eventually(func() error {
-				return client.IgnoreNotFound(c.Delete(ctx, bundle))
+				return client.IgnoreNotFound(c.Delete(ctx, biDuplicate))
 			}).Should(Succeed())
 
 			By("deleting the testing original BundleInstance resource")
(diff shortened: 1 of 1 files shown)


test/e2e/plain_provisioner_test.go
```diff
@@ -882,9 +882,33 @@ var _ = Describe("plain provisioner bundleinstance", func() {
...
-				return c.Get(ctx, client.ObjectKeyFromObject(biOriginal), &rukpakv1alpha1.BundleInstance{})
-			}).Should(Succeed(), "failed to validate original bundle instance was created in time")
+
+			// ensure the original BI owns the underlying bundle before creating the duplicate
+			By("projecting a successful installation status for the original BundleInstance")
+			Eventually(func() bool {
+				if err := c.Get(ctx, client.ObjectKeyFromObject(biOriginal), biOriginal); err != nil {
+					return false
+				}
+				if biOriginal.Status.InstalledBundleName != "" {
```
akihikokuroda, line 892: Is this mistake?  Line 907 is also checking `biOriginal.Status.InstalledBundleName`.

exdx: Yup that's a good catch -- I consolidated it to check only once.

test/e2e/plain_provisioner_test.go
```diff
@@ -882,9 +882,31 @@ var _ = Describe("plain provisioner bundleinstance", func() {
...
+			// ensure the original BI owns the underlying bundle before creating the duplicate
+			By("projecting a successful installation status for the original BundleInstance")
+			Eventually(func() bool {
+				if err := c.Get(ctx, client.ObjectKeyFromObject(biOriginal), biOriginal); err != nil {
+					return false
+				}
+				if biOriginal.Status.InstalledBundleName != bundle.GetName() {
+					return false
+				}
+				existing := meta.FindStatusCondition(biOriginal.Status.Conditions, "Installed")
```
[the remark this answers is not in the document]
tylerslaton: +1

test/e2e/plain_provisioner_test.go
```diff
@@ -938,16 +949,16 @@ var _ = Describe("plain provisioner bundleinstance", func() {
 				if biDuplicate.Status.InstalledBundleName != "" {
 					return false
 				}
-				existing := meta.FindStatusCondition(biDuplicate.Status.Conditions, "Installed")
+				existing := meta.FindStatusCondition(biDuplicate.Status.Conditions, rukpakv1alpha1.TypeInstalled)
```
tylerslaton, line 952: Thanks for updating these 👍🏻

tylerslaton (approved):
This looks good to me from an implementation stand-point. 

Also this gave me a thought. What if we added a CI update that makes it so we run the `E2E` suite a few times before proceeding to confirm (or at least feel more confident) that we are not introducing any flakes?

test/e2e/plain_provisioner_test.go
```diff
@@ -882,9 +882,27 @@ var _ = Describe("plain provisioner bundleinstance", func() {
...
+				if existing == nil {
+					return false
+				}
+				expected := metav1.Condition{
+					Type:   rukpakv1alpha1.TypeInstalled,
+					Status: metav1.ConditionStatus(corev1.ConditionTrue),
+					Reason: rukpakv1alpha1.ReasonInstallationSucceeded,
+				}
+				return conditionsSemanticallyEqual(*existing, expected)
+			}).Should(BeTrue())
```
timflannagan, line 905: We've been building on top of this pattern that Joe had introduced recently: https://github.com/operator-framework/rukpak/blob/8f87c5cad26eb64d04516a5e65455c633f3e3aa7/test/e2e/plain_provisioner_test.go#L290-L297. I think my preference is using that pattern as it's slightly easier to ready, but this implementation is fine with me too.

exdx:
> What if we added a CI update that makes it so we run the `E2E` suite a few times before proceeding to confirm (or at least feel more confident) that we are not introducing any flakes?

Hmm we get more confidence about not introducing any potential flakes but at the cost of higher e2e runs for each PR. I feel like since we are at a fairly early stage in the project it may make more sense to have faster CI runs (since those really add up over time) then less/no flakes. Then, as the project matures, maybe it makes more sense to ensure flakes aren't causing issues at that point. As long as we stay vigilant and fix flakes when they occur then we shouldn't have too many issues -- occasionally they can get through but we can fix them right away. That's what I'm proposing, maybe @timflannagan can chime in?  

Also it's not entirely clear how many e2e runs it would take for the condition to manifest itself to cause the test to flake. For example, if we run the e2e suite 3 times, but the test flakes every 5 times, we would still introduce the flake into the tests.

timflannagan:
I think I'm still squarely in the camp that e2e tests can be treated as optional (but highly recommended). We should aim towards burning down flakes as we see them, but I think it's more important to stamp out functionality right now vs. being slowed down with additional testing processes.

test/e2e/plain_provisioner_test.go
exdx: Yes, I think it's a nicer pattern -- but let's leave the original implementation in place for now and have a refactor as a potential followup.

PR #225, opened 2022-04-05
.github/workflows: Consolidate the deploy and release workflows

### LINKED ISSUE #187: Consolidate deploy and release actions
We currently have two different actions in the repo that are responsible for pushing docker images:
1. `.github/workflows/deploy.yaml` builds and pushes `quay.io/operator-framework/plain-provisioner:main`
2. `.github/workflows/release.yaml` builds and pushes `quay.io/operator-framework/plain-provisioner:<tag>`

We should consolidate these so that there is only workflow responsible for pushing all image artifacts from this repo.

**Acceptance Criteria:**
 - only a single gha for building and pushing all image artefacts from this repo

timflannagan:
Goal: consolidate the deploy.yaml and release.yaml into a single workflow file. Remove the unused deploy.yaml file.

Closes #187


timflannagan:
Note: it wasn't clear to me what briefly poking at this locally whether there's anything else that was needed outside of copy-pasting the deploy job to the release workflow and deleting that now redundant deploy.yaml file. 

The release.yaml already had the main branch push trigger specified, so we're all set there, and I'm not aware of any easy way of consolidating the goreleaser job when we're interacting with non-image-tags (as it complains about a dirty staging environment, at least for me locally).

timflannagan:
What we could do in theory is have goreleaser build these binaries and manifestlist images but isn't responsible for any release automation, and then we have a docker job that's responsible for determining which container image tag is applicable based on the event trigger for that workflow, logging into the remote container registry, and somehow pushing those locally build image layers up to the configured container image that was prepared in a previous job.

.github/workflows/release.yaml
```diff
@@ -41,3 +41,28 @@ jobs:
...
+    env:
+      IMAGE: quay.io/operator-framework/plain-provisioner:main
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '~1.17'
+
+      - name: Build container
+        run: make build-container
```
timflannagan, line 56: I guess this step is technically redundant if goreleaser is also running on PRs and pushes to main. Maybe we could have this entire main job have a top-level if conditional that only runs on the main branch trigger. Thoughts?

joelanford: Long term, I imagined we'd have goreleaser handle main branch and tag builds with the same config.

joelanford: SDK manages to actually build and push from the same goreleaser config, so it's do-able. Just slightly more complex setup iirc.

joelanford: Oh I was thinking of operator-registry actually (but maybe we borrowed from SDK)

https://github.com/operator-framework/operator-registry/blob/de36104087733013025629d80bf8445045a8a6e7/.github/workflows/goreleaser.yaml#L145

timflannagan: Looking at https://github.com/operator-framework/operator-registry/blob/master/.github/workflows/goreleaser.yaml#L147-L151 again, my impression is that this kind of implementation is _really_ hacky and it would be nice to have a release and deploy that's feels like complex? That said, I'm a fan of the ingenuity and it certainly bridges the gap when goreleaser is the entrypoint for building binaries and container images.

tylerslaton: Would it make sense for us to just add a new tag to deploy for main into the `.goreleaser` file?

joelanford: @timflannagan That piece of operator-registry's build is _very_ quirky for two reasons:
1. `opm` is a CLI so we need to build binaries for a variety of OS/arch combinations and upload them to the release.
2. `opm` is built with CGO_ENABLED=1, which complicates the builds for windows and macOS (macOS is especially problematic because Apple disallows cross-compiling from Linux)

For this, we're just building linux binaries, we can cross-compile easily since we are using CGO_ENABLED=0, and we don't need to upload these binaries to our GH releases.

In short, GoReleaser can do everything we need it to do (I think).

timflannagan: @joelanford Just so I'm clear here: you're suggesting consolidating everything such that goreleaser handles the entire container image building process, and we have a separate job that's responsible for uploading the image layer archives and pushing that to the configured plain provisioner repository?

joelanford: > you're suggesting consolidating everything such that goreleaser handles the entire container image building process

yes, for both `main` branch and tags.

> have a separate job that's responsible for uploading the image layer archives and pushing that to the configured plain provisioner repository

no, goreleaser should be able to build/push the images as well.

the change i'm suggesting basically boils down to:
1. drop the `deploy` workflow
2. update the `goreleaser` workflow to trigger on pushes to `main`
3. update the `.goreleaser.yml` (similar to operator-registry's) to support passing image tags as environment variables, so that the same config works for both branches and tags
4. update the `goreleaser` workflow to correctly set the desired image environment variables, and set the appropriate goreleaser flags to allow image build/push even when the current build isn't a tag (i think that's `--skip-validate`)

timflannagan: Okay works for me - I can try and pick this back up next week assuming I'll have time to dedicate towards rukpak. Feel free to pick this up too in case you want some light work while on PTO :upside_down_face:.

timflannagan: I looked into this some more this afternoon and it looks like --skip-validate is a valid way to get around the need for running goreleaser against a non-tag build. The only thing I was running into locally was that there's no clear way to configure the publishing process, e.g. push those container and manfiestlist image, but don't publish a SCM release and run changelog generation which is unfortunate.

I think we can hack around this though by running `--skip-validate --skip-publish` and handling the publishing ourselves (hopefully `docker images` works) through explicitly docker commands.

timflannagan: I wonder whether it's possible to dynamically flip release.disable: e.g. have an environment variable `{{ env.RUN_RELEASE_PIPELINE }}` that defaults to false in the Makefile, and we explicitly override this when we're building up the GORELEASER_ARGS in the GHA workflow for tag event triggers?

timflannagan: @joelanford I took a different approach in https://github.com/operator-framework/rukpak/pull/250 which uses a goreleaser template file + envtest, and passes the release Makefile target a rendered .goreleaser.yml configuration file. The release workflow is responsible for enabling the release pipeline now by flipping a variable to true.

timflannagan:
Closing as #250 merged.

PR #227, opened 2022-04-06
fix: refactor git bundles e2e test

### LINKED ISSUE #214: Refactor git-based e2e tests to be more idiomatic

exdx:
Closes #214

### DIFF
--- a/test/e2e/plain_provisioner_test.go
+++ b/test/e2e/plain_provisioner_test.go
@@ -351,15 +351,19 @@ var _ = Describe("plain provisioner bundle", func() {
 
 	When("Bundles are backed by a git repository", func() {
 		var (
-			bundles []*rukpakv1alpha1.Bundle
-			ctx     context.Context
+			ctx context.Context
 		)
+
 		BeforeEach(func() {
 			ctx = context.Background()
+		})
 
-			By("creating the git based Bundles")
-			bundles = []*rukpakv1alpha1.Bundle{
-				{
+		When("the bundle is backed by a git commit", func() {
+			var (
+				bundle *rukpakv1alpha1.Bundle
+			)
+			BeforeEach(func() {
+				bundle = &rukpakv1alpha1.Bundle{
 					ObjectMeta: metav1.ObjectMeta{
 						GenerateName: "combo-git-commit",
 					},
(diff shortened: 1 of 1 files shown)


test/e2e/plain_provisioner_test.go
```diff
@@ -351,115 +351,100 @@ var _ = Describe("plain provisioner bundle", func() {
 
 	When("Bundles are backed by a git repository", func() {
```
exdx, line 352: For some reason, I can't target this test during a focus as its written now. I'm not sure whether its due to the way it's written but if it's not running during the overall e2e run then I will have to change it. It may not be finding the spec inside.

timflannagan: Yep I didn't see that When get run in the latest e2e-kind action output. Super weird.

exdx:
/hold 
seems to be getting skipped

test/e2e/plain_provisioner_test.go
```diff
@@ -351,115 +351,100 @@ var _ = Describe("plain provisioner bundle", func() {
...
+						└── combo.io_templates.yaml
+					*/
+					return len(bundle.Status.Info.Objects) == 7
+				}).Should(BeTrue())
+
+				// delete bundle
+				Eventually(func() bool {
+					err := c.Delete(ctx, bundle)
+					return apierrors.IsNotFound(err)
+				}).Should(BeTrue())
```
timflannagan, line 396: It seems like we risk this Bundle being stranded if the Eventually statement above fails.

test/e2e/plain_provisioner_test.go
```diff
@@ -351,115 +351,100 @@ var _ = Describe("plain provisioner bundle", func() {
...
+
+				Eventually(func() bool {
+					if err := c.Get(ctx, client.ObjectKeyFromObject(bundle), bundle); err != nil {
+						return false
+					}
+					return bundle.Status.Phase == rukpakv1alpha1.PhaseUnpacked
+				}).Should(BeTrue())
+
+				Eventually(func() bool {
+					if err := c.Get(ctx, client.ObjectKeyFromObject(bundle), bundle); err != nil {
```
tylerslaton, line 373: Is there a reason we're retrieving the `bundle` twice? We did it in the eventually above this ^ It could be possible to refactor this to only perform a `Get` once unless you've already thought of that.

tylerslaton (approved):
Nice, I think that the `DescribeTable` structure is quite clean in action.

test/e2e/plain_provisioner_test.go
exdx: That's a good point. I can explore trying to delete the bundle in an AfterEach, the trick is how to get the relevant bundle to delete. It's possible to just do a List and delete whatever comes back? Or label the bundle when creating it and then using a label selector to query for just those bundles (that sounds good since it's less destructive and doesn't impact bundles outside that particular e2e scope).

test/e2e/plain_provisioner_test.go
exdx: Yup, that makes sense. We can as well wrap everything in one Eventually().

test/e2e/plain_provisioner_test.go
timflannagan: Yeah I think my preference is towards labeling them and firing off a list query. I'm not sure whether this introduces non-determinism into our testing suite.

exdx: Yes, I'm not super thrilled about that, but I'm not sure there's another way given the limitations around `AfterEach` getting data from a `DescribeTable` 🤷

test/e2e/plain_provisioner_test.go
```diff
@@ -351,116 +351,107 @@ var _ = Describe("plain provisioner bundle", func() {
...
-								},
+		AfterEach(func() {
+			list := &rukpakv1alpha1.BundleList{}
+			err := c.List(ctx, list, &client.ListOptions{LabelSelector: labels.SelectorFromSet(gitLabel)})
+			Expect(err).To(BeNil())
+
+			for _, b := range list.Items {
+				err := c.Delete(ctx, &b)
+				Expect(err).To(BeNil())
+			}
```
timflannagan, line 369: Looking at this again: I think I'm torn between approving this PR so we can move onto other work, vs. second guessing this implementation. This still feels somewhat hacky, and I'd still think we'd need to move to a model that has the ability to enact a BeforeEach/AfterEach setup s.t. we don't need to perform list queries and $somewhat blindly delete a setup of Bundles that may, or may not, exist on cluster after these various tests.

Thoughts? Maybe this comes back how careful we need to be about the potential for non-deterministic tests, or maybe OLM's e2e testing suite has scarred me for life :shrug:.

exdx:
/hold
concerns about the AfterEach() implementation

akihikokuroda (approved):
This looks good to me.  Thanks!

timflannagan (approved):

PR #234, opened 2022-04-08
wip: major kustomize refactor

joelanford:
this commit has a whole bunch of related, but not-required-to-be-atomic
changes. just wanted to get this up for high level review of the
approach and to make sure i don't forget the components of this
changeset, which are:
- moving plain provisioner manifests into kustomize control
- removing rbac for webhook (it's unnecessary, at least right now)
- generating rbac and webhook configurations from kubebuilder markers
- group the api and provisioner installs into a single kustomize | kubectl invocation

PR #235, opened 2022-04-09
manifests/apis: more cleanup and automatic generation

joelanford:
This PR is the slice of #234 that involves a few more improvements in the `manifests/apis` directory. The two most substantive changes in this PR are the removal of unnecessary RBAC given to the webhook deployment's service account and the refactoring that enables the validating webhook configuration to be automatically generated from the [kubebuilder markers defined in the bundle webhook implementation](https://github.com/operator-framework/rukpak/blob/3ee8d201ec70315c2dd7080db43771c85884fdb8/api/v1alpha1/bundle_webhook.go#L37)

1. manifests/apis/webhooks: remove all RBAC
2. manifests/apis/webhooks: separate service and certificate out of webhook.yaml
3. manifests/apis/webhooks: generate validatingwebhookconfiguration from kubebuilder markers
4. manifests/apis/crds: remove 'base' directory layer
5. Makefile: remove unnecessary 'controller-gen schemapatch' call


akihikokuroda:
It’s a lot of changes. I’ll try to check everything Good.

manifests/apis/webhooks/resources/service.yaml
```diff
@@ -0,0 +1,13 @@
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: webhook-service
+  namespace: system
+spec:
```
akihikokuroda, line 7: This service name and namespace don't match to the client service is defined in the validatingwebhookconfig
```
webhooks:
- admissionReviewVersions:
  - v1
  clientConfig:
    caBundle: ........
    service:
      name: rukpak-webhook-service
      namespace: rukpak-system
      path: /validate-core-rukpak-io-v1alpha1-bundle
      port: 443
```
I got connection refused  message.

akihikokuroda (approved):
LGTM. Thanks!

manifests/apis/webhooks/resources/namespace.yaml
```diff
@@ -1,4 +1,4 @@
 apiVersion: v1
 kind: Namespace
 metadata:
-  name: rukpak-system
+  name: system
```
anik120, line 4: Why change this name to just `system`? `rukpak-system` was very intuitive to me when I bootstrapped rukpak for the first time. I'd imagine `system` wouldn't be as obvious that that's the namespace where the rukpak resources are deployed.

anik120 (commented):
Since we're all learning about kustomization for this repo, would you be open to, very quickly, leave comments explaining some of the features you used for writing the kustomize configs? Example, a short comment on the usage of `vars` and how that's helping us, etc.

Will probably help future devs too to have something like that on record via the PRs

manifests/apis/webhooks/resources/namespace.yaml
akihikokuroda: The `namePrefix` in `manifests/apis/webhooks/kustomization.yml` adds `rubpak-` to the names in the webhook/resources.

manifests/apis/webhooks/kustomization.yml
```diff
@@ -1,9 +1,44 @@
...
+
+vars:
+- name: CERTIFICATE_NAMESPACE # namespace of the certificate CR
+  objref:
+    kind: Certificate
+    group: cert-manager.io
+    version: v1
+    name: webhook-certificate # this name should match the one in certificate.yaml
+  fieldref:
+    fieldpath: metadata.namespace
```
joelanford, line 26: This defines the variable `CERTIFICATE_NAMESPACE`, which is derived from the `metadata.namespace` value of the `cert-manager.io/v1 Certificate` called `webhook-certificate`.

manifests/apis/webhooks/kustomization.yml
```diff
@@ -1,9 +1,44 @@
...
+- name: CERTIFICATE_NAME
+  objref:
+    kind: Certificate
+    group: cert-manager.io
+    version: v1
+    name: webhook-certificate # this name should match the one in certificate.yaml
```
joelanford, line 32: This defines the variable `CERTIFICATE_NAME`, which is derived from the name of the `cert-manager.io/v1 Certificate` called `webhook-certificate`. 

The object name is the default, used when `fieldref.fieldpath` is not set.

manifests/apis/webhooks/kustomization.yml
```diff
@@ -1,9 +1,44 @@
...
+- name: SERVICE_NAMESPACE # namespace of the service
+  objref:
+    kind: Service
+    version: v1
+    name: webhook-service
+  fieldref:
+    fieldpath: metadata.namespace
```
joelanford, line 39: This defines the variable `SERVICE_NAMESPACE`, which is derived from the `metadata.namespace` value of the `v1 Service` called `webhook-service`.

manifests/apis/webhooks/kustomization.yml
```diff
@@ -1,9 +1,44 @@
...
+- name: SERVICE_NAME
+  objref:
+    kind: Service
+    version: v1
+    name: webhook-service
```
joelanford, line 44: This defines the variable `SERVICE_NAME`, which is derived from the name of the `v1 Service` called `webhook-service`. 

The object name is the default, used when `fieldref.fieldpath` is not set.

manifests/apis/webhooks/kustomizeconfig.yaml
```diff
@@ -0,0 +1,28 @@
...
+  - kind: Certificate
+    group: cert-manager.io
+    path: spec/issuerRef/name
+
+- kind: Service
+  version: v1
+  fieldSpecs:
+  - kind: ValidatingWebhookConfiguration
+    group: admissionregistration.k8s.io
+    path: webhooks/clientConfig/service/name
```
joelanford, line 14: [Name references](https://github.com/kubernetes-sigs/kustomize/blob/master/examples/transformerconfigs/README.md#name-reference-transformer) tie a target resource's name to a list of other resources' referrers' names. Once tied, the referrers' names will change alongside the target name via transformers like namePrefix and nameSuffix.

So the first item in this section says "`spec.issuerRef.name` references the name of a `cert-manager.io Issuer`." This enables kustomize to apply the same transformation to both fields such that a change to an Issuer's name is _also_ applied to a Certificates `spec.issuerRef.name`.

manifests/apis/webhooks/kustomizeconfig.yaml
```diff
@@ -0,0 +1,28 @@
...
+  path: webhooks/clientConfig/service/namespace
+  create: true
+
+varReference:
+- kind: Certificate
+  group: cert-manager.io
+  path: spec/dnsNames
+- kind: ValidatingWebhookConfiguration
+  group: admissionregistration.k8s.io
+  path: metadata/annotations
```
joelanford, line 28: `varReference` paths are the paths in the specified group/version where variable substitution will occur.

joelanford:
For those reviewing:
1. I rebased this after merging #236 
2.  It'll definitely help to look at this PR commit-by-commit. It's all pretty straightforward except for the commit that switches over to autogenerating the webhook configs from kubebuilder markers.

api/v1alpha1/bundle_webhook.go
```diff
@@ -34,7 +34,7 @@ func (r *Bundle) SetupWebhookWithManager(mgr ctrl.Manager) error {
 		Complete()
 }
 
-//+kubebuilder:webhook:path=/validate-core-rukpak-io-v1alpha1-bundle,mutating=false,failurePolicy=fail,sideEffects=None,groups=core.rukpak.io,resources=bundles,verbs=create;update,versions=v1alpha1,name=core.rukpak.io,admissionReviewVersions=v1
+//+kubebuilder:webhook:path=/validate-core-rukpak-io-v1alpha1-bundle,mutating=false,failurePolicy=fail,sideEffects=None,groups=core.rukpak.io,resources=bundles,verbs=create;update,versions=v1alpha1,name=vbundles.core.rukpak.io,admissionReviewVersions=v1
```
timflannagan, line 37: I'm going commit-by-commit when reviewing this PR, so it's possible I'm looking at a backwards facing snapshot: is the `vbundles*` change a typing mistake? Should it be `name=bundles.core.rukpak.io` instead?

manifests/apis/webhooks/resources/webhook.yaml
```diff
@@ -1,21 +1,20 @@
...
   clientConfig:
     service:
-      name: rukpak-webhook
-      namespace: rukpak-system
+      name: webhook-service
+      namespace: system
       path: /validate-core-rukpak-io-v1alpha1-bundle
-      port: 443
   failurePolicy: Fail
+  name: vbundles.core.rukpak.io
```
timflannagan, line 17: Yep: same comment as above. Is this a spelling mistake?

timflannagan (approved):
This looks reasonable to me. Thanks for splitting up the commits to make this easier to review. I don't have a strong opinion on moving to a kustomize-focused model, despite kustomize feeling it has a decent barrier of entry at times.

I had one potential blocking comment around the `vbundles` change that I'd like to get clarification on, but I'll approve these changes ahead of time. Do we need a second approval here, or can we push these changes through?

api/v1alpha1/bundle_webhook.go
joelanford: No that was intentional. I took a hint from the way kubebuilder scaffolds webhooks. It uses `v` and `m` prefixes for validating and mutating webhooks respectively.

PR #237, opened 2022-04-11
(doc): git-based bundles with custom directory specified 

exdx:
This PR adds additional documentation for bundles with custom bundle directory roots.

### DIFF
--- a/docs/git-bundles.md
+++ b/docs/git-bundles.md
@@ -8,8 +8,12 @@ make it available within the cluster.
 
 The git repository backing the bundle, or "bundle git repo", must have a certain structure in order to produce a valid
 plain Bundle. It should have a directory where the Kubernetes manifests are stored -- this directory is used when
-copying manifests onto the cluster. By default, this directory is assumed to be `/manifests` at the top level. For an
-example of a bundle git repo that conforms to the specifications, see
+copying manifests onto the cluster. By default, this directory is assumed to be `/manifests` at the root level, but can
+be specified via `spec.source.git.directory`.
+
+> Note: There must be a `manifests` directory in the provided directory in order to have a valid bundle git repo.
+
+For an example of a bundle git repo that conforms to the specifications, see
 the [combo repository](https://github.com/operator-framework/combo/).
 
 When creating a Bundle from a git source, a reference to a particular commit, tag, or branch must be provided for the
--- a/docs/plain-bundle-spec.md
+++ b/docs/plain-bundle-spec.md
@@ -70,8 +70,10 @@ on bundles backed by git repositories, see the [git based bundles doc](git-bundl
 * The static manifests must be located in the root-level /manifests directory in a bundle image for the bundle to be a
   valid `plain+v0` bundle that the provisioner can unpack. A plain bundle image without a /manifests directory is
   invalid and will not be successfully unpacked onto the cluster.
-* For a bundle git repo, this limitation does not exist, and the manifests can be in any directory in the repository.
-  The manifest directory is assumed to be ./manifests in a bundle git repo but can be provided at runtime.
+* For a bundle git repo, the manifests directory can be anywhere in the repository, not just at the root-level. The
+  location can be specified via `spec.source.git.directory`. There must be a `manifests` directory at the provided
+  location in order to have a valid bundle git repo. If a specific directory is not provided, it is assumed to be
+  ./manifests in a bundle git repo.
 * The manifests directory should be flat: all manifests should be at the top-level with no subdirectories.
 * The plain bundle image can be built from any base image, but `scratch` is recommended as it keeps the resulting bundle
   image a minimal size.


docs/git-bundles.md
```diff
@@ -8,8 +8,9 @@ make it available within the cluster.
 
 The git repository backing the bundle, or "bundle git repo", must have a certain structure in order to produce a valid
 plain Bundle. It should have a directory where the Kubernetes manifests are stored -- this directory is used when
-copying manifests onto the cluster. By default, this directory is assumed to be `/manifests` at the top level. For an
-example of a bundle git repo that conforms to the specifications, see
+copying manifests onto the cluster. By default, this directory is assumed to be `/manifests` at the root level, but can
+be specified via `spec.source.git.directory`. There must be a `manifests` directory in the provided directory in order to
+have a valid bundle git repo. For an example of a bundle git repo that conforms to the specifications, see
 the [combo repository](https://github.com/operator-framework/combo/).
```
timflannagan, line 14: How do we feel about making this a note to make it more explicit:

---

The git repository backing the bundle, or "bundle git repo", must have a certain structure in order to produce a valid
plain Bundle. It should have a directory where the Kubernetes manifests are stored -- this directory is used when
copying manifests onto the cluster. By default, this directory is assumed to be `/manifests` at the root level, but can
be specified via `spec.source.git.directory`.

**Note**: There must be a `manifests` directory in the provided directory in order to
have a valid bundle git repo.

For an example of a bundle git repo that conforms to the specifications, see
the [combo repository](https://github.com/operator-framework/combo/).

timflannagan:
Should we separate out the doc and e2e changes in their own separate commits?

docs/git-bundles.md
exdx: That sounds good

exdx:
> Should we separate out the doc and e2e changes in their own separate commits?

We could yes, especially since the e2e is not passing currently (can't figure out why). This PR can just be for the doc then.

timflannagan (approved):

joelanford:
@exdx @timflannagan 

> The git repository backing the bundle, or "bundle git repo", must have a certain structure in order to produce a valid plain Bundle. It should have a directory where the Kubernetes manifests are stored -- this directory is used when copying manifests onto the cluster.

Should the "Git-based bundles" doc talk about plain bundles other than potentially using them as an example? It seems like we're conflating how a bundle is sourced (`bundle.spec.source`) and how a bundle is parsed/interpreted (`bundle.spec.provisionerClassName`)

> By default, this directory is assumed to be /manifests at the root level, but can be specified via spec.source.git.directory.

This ^ text isn't quite aligned to the implementation, still. I think this should be:

> By default, this directory is assumed to be git repository root, but can be specified via `spec.source.git.directory`. 
> 
> Take plain+v0 bundles for example. In plain+v0 bundles, we expect to find `./manifests` in the bundle root directory. When sourcing the plain+v0 bundle from a git repository, the bundle root is defined by `spec.source.git.directory` (which defaults to `<gitRoot>` when unset). Therefore, when `spec.source.git.directory` is unset, we expect to find `<gitRoot>/manifests`. When `spec.source.git.directory` is set to `path/to/bundle`, we expect to find `<gitRoot>/path/to/bundle/manifests`.

exdx:
> @exdx @timflannagan
> 
> > The git repository backing the bundle, or "bundle git repo", must have a certain structure in order to produce a valid plain Bundle. It should have a directory where the Kubernetes manifests are stored -- this directory is used when copying manifests onto the cluster.
> 
> Should the "Git-based bundles" doc talk about plain bundles other than potentially using them as an example? It seems like we're conflating how a bundle is sourced (`bundle.spec.source`) and how a bundle is parsed/interpreted (`bundle.spec.provisionerClassName`)
> 
> > By default, this directory is assumed to be /manifests at the root level, but can be specified via spec.source.git.directory.
> 
> This ^ text isn't quite aligned to the implementation, still. I think this should be:
> 
> > By default, this directory is assumed to be git repository root, but can be specified via `spec.source.git.directory`.
> > Take plain+v0 bundles for example. In plain+v0 bundles, we expect to find `./manifests` in the bundle root directory. When sourcing the plain+v0 bundle from a git repository, the bundle root is defined by `spec.source.git.directory` (which defaults to `<gitRoot>` when unset). Therefore, when `spec.source.git.directory` is unset, we expect to find `<gitRoot>/manifests`. When `spec.source.git.directory` is set to `path/to/bundle`, we expect to find `<gitRoot>/path/to/bundle/manifests`.

Based on the implementation we have currently, the `spec.source.git.directory` has to point to the manifests directly to have the unpack operation succeed. So if the manifests are in `./dev/deploy/manifests` then the directory variable has to point to there (see #238 for an example). It's almost as `spec.source.git.directory` should be renamed to `spec.source.git.manifestsDir` based on what our code is doing currently. I'm going to follow-up with some additional documentation improvements to make this more clear, now that the custom directory test is passing.

timflannagan:
@exdx @joelanford Should we open a ticket for making the documentation agnostic to an individual provisioner?

exdx:
We could -- I'm addressing this in a follow-up for this doc now

joelanford:
I _think_ #240  PR resolves the issue with how `spec.source.git.directory` is behaving.

PR #238, opened 2022-04-11
(e2e): verify a specified git bundle directory root works

### LINKED ISSUE #233: Git-based bundles do not support custom directory field
In the plain bundle spec, it's mentioned that the limitation of having the manifests stored under `./manifests` does not apply to git based bundles, since the `spec.source.git.directory` field exists and a user can point to an arbitrary directory in the repository where the manifests are stored.

However, applying a bundle with a custom directory set results in an error. 
```
k apply -f -<<EOF
kind: Bundle
metadata:
  name: combo-branch-ref-deploy
spec:
  source:
    type: git
    git:
      ref:
        branch: main
      directory: ./dev/deploy
      repository: https://github.com/exdx/combo-bundle
EOF
```

```
Status:
  Conditions:
    Last Transition Time:  2022-04-08T20:39:37Z
    Message:               get objects from bundle manifests: readdir manifests: file does not exist
    Reason:                UnpackFailed
    Status:                False
    Type:                  Unpacked
```

A/C:
* Determine whether git bundles should in fact have the same limitations as image based bundles
* If so, update the plain bundle spec
* If not, update code to handle arbitrary directories from git based bundles
* Add an e2e test for this scenario

exdx:
Adds an e2e test to ensure that git-based bundles with custom directories work as expected. Closes #233.

### DIFF
--- a/test/e2e/plain_provisioner_test.go
+++ b/test/e2e/plain_provisioner_test.go
@@ -532,9 +532,67 @@ var _ = Describe("plain provisioner bundle", func() {
 				}).Should(BeTrue())
 			})
 		})
+
+		When("the bundle has a custom manifests directory", func() {
+			var (
+				bundle *rukpakv1alpha1.Bundle
+			)
+			BeforeEach(func() {
+				bundle = &rukpakv1alpha1.Bundle{
+					ObjectMeta: metav1.ObjectMeta{
+						GenerateName: "combo-git-custom-dir",
+					},
+					Spec: rukpakv1alpha1.BundleSpec{
+						ProvisionerClassName: plainProvisionerID,
+						Source: rukpakv1alpha1.BundleSource{
+							Type: rukpakv1alpha1.SourceTypeGit,
+							Git: &rukpakv1alpha1.GitSource{
+								Repository: "https://github.com/exdx/combo-bundle",
+								Directory:  "./dev/deploy/manifests",
+								Ref: rukpakv1alpha1.GitRef{
+									Branch: "main",
+								},
+							},
+						},
+					},
+				}
+				err := c.Create(ctx, bundle)
+				Expect(err).To(BeNil())
+			})
+
+			AfterEach(func() {
+				err := c.Delete(ctx, bundle)
+				Expect(err).To(BeNil())
+			})
+
+			It("Can create and unpack the bundle successfully", func() {
+				Eventually(func() bool {
+					if err := c.Get(ctx, client.ObjectKeyFromObject(bundle), bundle); err != nil {
+						return false
+					}
+					if bundle.Status.Phase != rukpakv1alpha1.PhaseUnpacked {
+						return false
+					}
+					if bundle.Status.Info == nil {
+						return false
+					}
+					/*
+						manifests
+						├── 00_namespace.yaml
+						├── 01_cluster_role.yaml
+						├── 01_service_account.yaml
+						├── 02_deployment.yaml
+						├── 03_cluster_role_binding.yaml
+						├── combo.io_combinations.yaml
+						└── combo.io_templates.yaml
+					*/
+					return len(bundle.Status.Info.Objects) == 7
+				}).Should(BeTrue())
+			})
+		})
 	})
 
-	When("a bundle containing nested directry is created", func() {
+	When("a bundle containing nested directory is created", func() {
 		var (
 			bundle *rukpakv1alpha1.Bundle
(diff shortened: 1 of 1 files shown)


exdx:
This e2e test fails, even though everything seems to be in order. The custom `./dev/deploy` directory has a `manifests` directory inside it that contains all the expected manifests. The unpack pod runs to completion, and looking at the logs it's clear that it was able to unpack the manifests successfully. 
```
k -n rukpak-system logs plain-unpack-bundle-combo-branch-ref-deploy | jq -r '.content' | base64 -d | gunzip | tar tv
drwxrwxrwx  0 0      0           0 Apr 11 09:39 .
drwxr-xr-x  0 0      0           0 Apr 11 09:39 deploy
drwxr-xr-x  0 0      0           0 Apr 11 09:39 deploy/manifests
-rw-r--r--  0 0      0          55 Apr 11 09:39 deploy/manifests/00_namespace.yaml
-rw-r--r--  0 0      0         188 Apr 11 09:39 deploy/manifests/01_cluster_role.yaml
-rw-r--r--  0 0      0          88 Apr 11 09:39 deploy/manifests/01_service_account.yaml
-rw-r--r--  0 0      0         536 Apr 11 09:39 deploy/manifests/02_deployment.yaml
-rw-r--r--  0 0      0         264 Apr 11 09:39 deploy/manifests/03_cluster_role_binding.yaml
-rw-r--r--  0 0      0        7141 Apr 11 09:39 deploy/manifests/combo.io_combinations.yaml
-rw-r--r--  0 0      0        2121 Apr 11 09:39 deploy/manifests/combo.io_templates.yaml
```

yet the Bundle shows a failed status with the following message:
```   
Message:               get objects from bundle manifests: readdir manifests: file does not exist
```

test/e2e/plain_provisioner_test.go
```diff
@@ -532,9 +532,67 @@ var _ = Describe("plain provisioner bundle", func() {
...
+						└── combo.io_templates.yaml
+					*/
+					return len(bundle.Status.Info.Objects) == 7
+				}).Should(BeTrue())
+			})
+		})
 	})
 
-	When("a bundle containing nested directry is created", func() {
+	When("a bundle containing nested directory is created", func() {
```
timflannagan, line 595: :+1:

test/e2e/plain_provisioner_test.go
```diff
@@ -532,9 +532,67 @@ var _ = Describe("plain provisioner bundle", func() {
...
+						manifests
+						├── 00_namespace.yaml
+						├── 01_cluster_role.yaml
+						├── 01_service_account.yaml
+						├── 02_deployment.yaml
+						├── 03_cluster_role_binding.yaml
+						├── combo.io_combinations.yaml
+						└── combo.io_templates.yaml
+					*/
+					return len(bundle.Status.Info.Objects) == 7
```
timflannagan, line 589: I know someone had mentioned the want for a comment to be present but this decreases readability imo, and finding the test manifests is straight forward as we commit the bundle testdata manifests directly.

test/e2e/plain_provisioner_test.go
```diff
@@ -532,9 +532,67 @@ var _ = Describe("plain provisioner bundle", func() {
...
+				err := c.Create(ctx, bundle)
+				Expect(err).To(BeNil())
+			})
+
+			AfterEach(func() {
+				err := c.Delete(ctx, bundle)
+				Expect(err).To(BeNil())
+			})
+
+			It("Can create and unpack the bundle successfully", func() {
```
timflannagan, line 568: supersuper nit: the rest of the e2e package seems to use lowercase for the first word in the It spec.

timflannagan (approved):

test/e2e/plain_provisioner_test.go
exdx: We commit the test manifests for the image-based bundles, but for the git based bundles we don't have the content locally (we actually go to github for the source). This isn't ideal IMO, as we would like all testdata to live in the repository. Might be worth looking into a follow-up where we could spin up a git server and serve a repository we are interested in testing and reference that in the Bundle versus github directly.

exdx:
Note: this test works when updating the `spec.source.git.directory` to point to `./dev/deploy/manifests` ie the manifests directory directly, instead of the root that contains the directory. I will update documentation in a follow-up to reflect that the directory field should point to the location of the manifests directory directly.

PR #239, opened 2022-04-12
(doc): follow-up to git-based bundles doc

exdx:
Makes the git-based bundles document more generic with respect to plain+v0 bundles ~and fixes the provided example for a bundle with custom directories to match the current implementation.~


exdx:
Updated wrt #240

joelanford (approved):
/lgtm

PR #240, opened 2022-04-12
internal/git/git.go: fix copying of bundle root

joelanford:
This commit copies the _contents_ of the bundle directory into `/bundle` rather than the specified bundle directory itself.

This resolves an issue where if `spec.source.git.directory` specified `path/to/root` (where `path/to/root` contains a `manifests` directory at `path/to/root/manifests`, the unpacked `/bundle` directory would contain `/bundle/root/manifests`, when it should contain `/bundle/manifests`.

I introduced this bug in #211 as part of those fixes.

### DIFF
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -38,18 +38,18 @@ func (c *checkoutCmd) String() string {
 	}
 
 	if commit != "" {
-		checkoutCommand = fmt.Sprintf("git clone %s %s && cd %s && git checkout %s && rm -r .git && cp -r %s /bundle",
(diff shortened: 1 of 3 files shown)


exdx (approved):
LGTM -- if this is the behavior that we want to have (where we copy what's inside of the provided directory instead of the directory itself). 

Only comment I have is that the defaulDirectory is still `./` so now in our checkout command we actually use `.//.` as the default directory when none is provided -- I'm a little suprised all the e2e tests pass but I guess `cp` can parse that as `./` just the same?

joelanford:
> I'm a little suprised all the e2e tests pass but I guess cp can parse that as ./ just the same?

It does seem to work and pass all the test cases, so I think its safe. If we find any other corner cases, we can always add another test case and make any further fixes.

Issue #209, opened 2022-04-01
Remove the install and deploy Makefile targets


timflannagan:
This is a follow-up to the conversation in the rukpak-dev channel: <https://kubernetes.slack.com/archives/C038B7MF75M/p1648845370402289>

The Makefile `install` and `deploy` targets do roughly the same thing right now. In the [plain provisioner documentation](https://github.com/operator-framework/rukpak/blob/main/internal/provisioner/plain/README.md) we instruct users to use the `make run` command on a local Kubernetes cluster in order to install the core and plain provisioner rukpak stack, and that `make run` target is sufficient for local and CI installation workflows.

Goal: remove the install/deploy Makefile targets.
Stretch: update the plain provisioner README and add a supported way of installing the plain provisioner stack.

**Acceptance Criteria:**
 - `install` and `deploy` make targets removed
 - README updated

PR #246, opened 2022-04-14
Remove the duplicate cert-manager commands; remove install/deploy directives

exdx:
The install section of the Makefile is currently pretty confusing and this PR attemps to streamline it, as well as removing the duplicate calls to `cert-manager` that are currently occurring. That being said, there are some other issues in the 0.3 milestone (#187 and #209) that address the Makefile as well so this PR could potentially not be necessary after those go through.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,7 +83,7 @@ test-e2e: ginkgo ## Run the e2e tests
 	$(GINKGO) -trace -progress $(FOCUS) test/e2e
 
 e2e: KIND_CLUSTER_NAME=rukpak-e2e
-e2e: build-container kind-cluster kind-load cert-mgr kind-load-bundles deploy test-e2e ## Run e2e tests against a kind cluster
+e2e: build-container kind-cluster kind-load kind-load-bundles run test-e2e ## Run e2e tests against a kind cluster
 
 kind-cluster: ## Standup a kind cluster for e2e testing usage
 	${KIND} delete cluster --name ${KIND_CLUSTER_NAME}
@@ -92,25 +92,25 @@ kind-cluster: ## Standup a kind cluster for e2e testing usage
 ###################
 # Install and Run #
 ###################
-.PHONY: install-apis install-plain install deploy run
+.PHONY: install-apis install-plain install-crdvalidator-webhook install-core-webhook install-webhooks run cert-mgr
 
 ##@ install/run:
 
-install-apis: cert-mgr generate kustomize ## Install the core rukpak CRDs
-	$(KUSTOMIZE) build manifests/apis | kubectl apply -f -
+install-apis: generate kustomize ## Install the core rukpak CRDs
+	$(KUSTOMIZE) build manifests/apis/crds | kubectl apply -f -
 
-install-plain: install-apis ## Install the rukpak CRDs and the plain provisioner
+install-plain: kustomize ## Install the plain provisioner
 	$(KUSTOMIZE) build manifests/provisioners/plain | kubectl apply -f -
 
-install-crdvalidator: cert-mgr kustomize ## Install the crdvalidator webhook manifests to the current cluster.
+install-crdvalidator-webhook: kustomize ## Install the crdvalidator webhook to the current cluster.
 	$(KUSTOMIZE) build manifests/crdvalidator | kubectl apply -f -
 
-install: install-plain ## Install all rukpak core CRDs and provisioners
+install-core-webhook: kustomize ## Install the core bundle webhook to the current cluster.
+	$(KUSTOMIZE) build manifests/apis/webhooks | kubectl apply -f -
 
-deploy: install-apis install-crdvalidator ## Deploy the operator to the current cluster
-	$(KUSTOMIZE) build manifests/provisioners/plain | kubectl apply -f -
+install-webhooks: cert-mgr install-crdvalidator-webhook install-core-webhook ## Install all webhooks to the current cluster
 
-run: build-container kind-load cert-mgr deploy ## Build image and run operator in-cluster
+run: build-container kind-load install-apis install-webhooks install-plain ## Build image and run operator in-cluster
 
 cert-mgr: ## Install the certification manager
 	kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/$(CERT_MGR_VERSION)/cert-manager.yaml


tylerslaton (approved):
I like this a lot more than how it was before. The only real question I have is if there is any need for us to cut down on the number of installs that we have here. I ask mainly as it may become tricky if we choose to add uninstall targets for each of them.

Makefile
```diff
@@ -96,21 +96,21 @@ kind-cluster: ## Standup a kind cluster for e2e testing usage
 
 ##@ install/run:
 
-install-apis: cert-mgr generate kustomize ## Install the core rukpak CRDs
-	$(KUSTOMIZE) build manifests/apis | kubectl apply -f -
+install-apis: generate ## Install the core rukpak CRDs
+	$(KUSTOMIZE) build manifests/apis/crds | kubectl apply -f -
```
akihikokuroda, line 100: I think that the core webhook is a part of the api.

Makefile
```diff
@@ -96,21 +96,21 @@ kind-cluster: ## Standup a kind cluster for e2e testing usage
...
+install-crdvalidator-webhook: ## Install the crdvalidator webhook manifests to the current cluster.
 	$(KUSTOMIZE) build manifests/crdvalidator | kubectl apply -f -
 
-install: install-plain ## Install all rukpak core CRDs and provisioners
+install-core-webhook:
+	$(KUSTOMIZE) build manifests/apis/webhooks | kubectl apply -f -
 
-deploy: install-apis install-crdvalidator ## Deploy the operator to the current cluster
-	$(KUSTOMIZE) build manifests/provisioners/plain | kubectl apply -f -
+install-webhooks: cert-mgr install-crdvalidator-webhook install-core-webhook
```
akihikokuroda, line 111: I’m not sure if putting the 2 different webhooks together make sense.

Makefile
exdx: I agree the core webhook can be part of the default installation but since it's a webhook and not a CRD should we treat it differently?

Makefile
exdx: The motivation is to only run `cert-mgr` once when installing webhooks. Currently we install cert-manager twice when installing the CRDvalidator webhook. I suppose we could tackle that by checking to see if cert manager is installed before installing cert-manager

Makefile
akihikokuroda: I think that the CRDs and it's validation should installed together.  WDYT?

exdx: I think that's fair -- I'm looking for a way to do that and not duplicate the cert-manager installation though. Maybe the way to do this is to check whether cert-manager is installed before installing it?

timflannagan: > I think that the CRDs and it's validation should installed together. WDYT?

I'm sure where I land with this tbh. I think my initial impression is that a user running `make install-apis` would result in only the CRDs being stamped out on a cluster, but at the same time, I can understand the perspective that we should also be installed the validation webhook. I don't have a strong opinion though: my current preference would be to separate them out into individual targets, and it looks like Dan's already doing this throughout these changes.

Makefile
```diff
@@ -96,21 +96,21 @@ kind-cluster: ## Standup a kind cluster for e2e testing usage
...
 
-install-crdvalidator: cert-mgr kustomize ## Install the crdvalidator webhook manifests to the current cluster.
+install-crdvalidator-webhook: ## Install the crdvalidator webhook manifests to the current cluster.
 	$(KUSTOMIZE) build manifests/crdvalidator | kubectl apply -f -
 
-install: install-plain ## Install all rukpak core CRDs and provisioners
+install-core-webhook:
+	$(KUSTOMIZE) build manifests/apis/webhooks | kubectl apply -f -
 
-deploy: install-apis install-crdvalidator ## Deploy the operator to the current cluster
```
timflannagan, line 110: Looking at the overall diff, it looks like we're removing the deploy Makefile target, but there's still a .PHONY deploy target?

Makefile
```diff
@@ -96,21 +96,21 @@ kind-cluster: ## Standup a kind cluster for e2e testing usage
...
-install-plain: install-apis ## Install the rukpak CRDs and the plain provisioner
+install-plain: ## Install the rukpak CRDs and the plain provisioner
 	$(KUSTOMIZE) build manifests/provisioners/plain | kubectl apply -f -
 
-install-crdvalidator: cert-mgr kustomize ## Install the crdvalidator webhook manifests to the current cluster.
+install-crdvalidator-webhook: ## Install the crdvalidator webhook manifests to the current cluster.
 	$(KUSTOMIZE) build manifests/crdvalidator | kubectl apply -f -
 
-install: install-plain ## Install all rukpak core CRDs and provisioners
+install-core-webhook:
```
timflannagan, line 108: Missing the comment that drives the `make help` output?

timflannagan: Also: do these new targets that rely on Kustomize need to have the kustomize target as a pre-requisite?

exdx: Yes, I tried to deduplicate the kustomize command as well since I wasn't sure whether each invocation rebuilds the kustomize binary from the tools directory? 

I suppose, if the targets are meant to be run individually, then we should rely on the kustomize target for each one? I was looking to avoid a scenario where running `make run` builds kustomize for each of the install targets.

timflannagan: I _think_ we're in the clear here as the tools.go approach that Joe had added should detect whether that binary is already present, but I'm mainly interested in adding the help output and removing the .PHONY target before approving this PR so I'll let you determine whether you think that kustomize prerequisite target is needed.

exdx: Yes, I was playing around with it locally, and it looks like it does detect the kustomize binary is present and doesn't build it each time, which is great. In that case I'll add the kustomize targets back in.

timflannagan (approved):

Makefile
joelanford: I had originally combined them because there's really no case where we want the CRDs to exist without the webhooks. If they could, that would allow custom objects to be created and stored in etcd that violate the checks that the webhook performs and thus potentially violate assumptions we make in the controllers.

In fact, I was on the path of combining _all_ of the rukpak repo's manifests into a single `kustomize build ./manifests | kubectl apply -f` invocation.

PR #247, opened 2022-04-14
Add uninstall command to Makefile 

exdx:
Adds an uninstall command which removes all namespaces and CRDs associated with the rukpak installation, restoring the cluster to the previous state before rukpak was installed. This results in a better local-dev experience of `kind create -> make run -> make uninstall` instead of `kind create -> make run -> kind delete`.

There is no way to "unload" the provisioner image loaded in via `kind load...` but on a subsequent invocation of `make run` the image should be loaded again to reflect any local changes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -116,6 +116,33 @@ cert-mgr: ## Install the certification manager
 	kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/$(CERT_MGR_VERSION)/cert-manager.yaml
 	kubectl wait --for=condition=Available --namespace=cert-manager deployment/cert-manager-webhook --timeout=60s
 
+uninstall: ## Uninstall all rukpak-related content from the cluster
+	kubectl delete ns rukpak-system
+	kubectl delete ns crdvalidator-system
+	kubectl delete ns cert-manager
+
+	kubectl delete crd bundles.core.rukpak.io
+	kubectl delete crd bundleinstances.core.rukpak.io
+	kubectl delete crd certificaterequests.cert-manager.io
+	kubectl delete crd certificates.cert-manager.io
+	kubectl delete crd challenges.acme.cert-manager.io
+	kubectl delete crd clusterissuers.cert-manager.io
+	kubectl delete crd issuers.cert-manager.io
+	kubectl delete crd orders.acme.cert-manager.io
+
+	kubectl delete validatingwebhookconfiguration cert-manager-webhook
+	kubectl delete validatingwebhookconfiguration crd-validation-webhook
+	kubectl delete validatingwebhookconfiguration rukpak-webhook
+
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -116,6 +116,19 @@ cert-mgr: ## Install the certification manager
 	kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/$(CERT_MGR_VERSION)/cert-manager.yaml
 	kubectl wait --for=condition=Available --namespace=cert-manager deployment/cert-manager-webhook --timeout=60s
 
+uninstall: ## Uninstall all rukpak-related content from the cluster
```
tylerslaton, line 119: Should this also delete the webhooks we install in the current `deploy`/`run` target?

Makefile
```diff
@@ -116,6 +116,19 @@ cert-mgr: ## Install the certification manager
...
+	kubectl delete ns crdvalidator-system
+	kubectl delete ns cert-manager
+	kubectl delete crd bundles.core.rukpak.io
+	kubectl delete crd bundleinstances.core.rukpak.io
+	kubectl delete crd certificaterequests.cert-manager.io
+	kubectl delete crd certificates.cert-manager.io
+	kubectl delete crd challenges.acme.cert-manager.io
+	kubectl delete crd clusterissuers.cert-manager.io
+	kubectl delete crd issuers.cert-manager.io
+	kubectl delete crd orders.acme.cert-manager.io
```
akihikokuroda, line 130: The following ValidatingWebhookConfigurations remain after this.
```
cert-manager-webhook 
crd-validation-webhook 
rukpak-webhook
```
They need to be cleaned, too.  The `crd-validation-webhook` causes an install error when the cert-manager is reinstalled.

exdx: Sounds good, will update

exdx:
This cleans up all the cert-manager related objects as well. There is a risk of this command getting stale over time, if its not consistently updated, but if it's useful enough then it can be kept up to date.

timflannagan (commented):
I'll defer to the masses but I think this kind of target may introduce a source of extra maintenance burden over time, and given the current state of the project, it seems sufficient to simply recycle the cluster (or manually clean up these resources using kustomize)?

exdx:
That's fair, I got a little tired of recycling the cluster all the time but it's probably the safest bet. This command would get outdated very fast, which is true, so for now we can not check it in. Using kustomize to delete resources would be nicer, I'll look into it.

joelanford:
I think we can revisit this once we have the kustomize stuff consolidated down to the point that we can run `kustomize build ./manifests`

With that, we can run:
- `kustomize build ./manifests | kubectl apply -f -` to apply rukpak to the cluster
- `kustomize build ./manifests | kubectl apply -f -` to delete rukpak from the cluster

PR #248, opened 2022-04-14
test(crdvalidator): add e2e test case for a crd update that breaks an existing cr

tylerslaton:
Adding a new test case to the `crdvalidator` e2e suite that prevents a schema update from breaking an existing CR.


exdx (approved):

exdx:
/hold
looks like lint is failing

PR #250, opened 2022-04-14
ci: Consolidate the deploy and release actions

### LINKED ISSUE #187: Consolidate deploy and release actions

timflannagan:
This is a slightly different take compared to the existing #225, and introduces a .goreleaser-template.yml file which includes dynamically enables the release pipeline when the release workflow is handling a tag event trigger.

For main branch and pull request event triggers, the release pipeline is disabled. The root Makefile has been updated to run envsubst and spitting out a rendered .goreleaser.yml, which is the default location that goreleaser expects, and the release target has been updated to add that target as a prerequisite. The root .goreleaser.yml file is now ignored in the .gitignore.

Closes #187.


.github/workflows/release.yaml
```diff
@@ -27,17 +28,30 @@ jobs:
 
     - name: Docker Login
       uses: docker/login-action@v1
-      if: startsWith(github.ref, 'refs/tags')
```
timflannagan, line 30: Looks like the release action is failing when attempting to log into the docker using the secret credentials. It seems like this is the correct behavior: https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow. I might be worth adding this conditional check back in place.

timflannagan:
I'm testing this out in my fork - already discovered a couple of issues.

/hold

timflannagan:
Ran this in my fork on the main branch trigger: https://quay.io/repository/tflannag/rukpak?tab=tags&tag=latest.

tylerslaton (approved):
I'm a fan of this approach to consolidating the two workflows. I do think that the if/elif chain can be slightly confusing to understand but it works as you demonstrated.

PR #253, opened 2022-04-18
test(crdvalidator): add unit testing for underlying crd validation package

tylerslaton:
In addition I am moving the `NewTestingCRD` function into a new `testutil` package in `test` along side a new function called `NewTestingCR` to streamline the creation of these resources.

### DIFF
--- a/internal/crd/crd_test.go
+++ b/internal/crd/crd_test.go
@@ -151,14 +151,17 @@ func TestValidate(t *testing.T) {
 			existingCr := testutil.NewTestingCR(testutil.DefaultCrName, testutil.DefaultGroup, tt.existingCrVersion, uniqueName)
 			newCrd := testutil.NewTestingCRD(uniqueName, testutil.DefaultGroup, tt.newCrdVersions)
 
-			// Create existing CRD
+			// Create existing CRD and wait for it to be ready
 			err := kubeclient.Create(ctx, existingCrd)
 			require.NoError(t, err, "failed to create initial crd for testing")
 			require.Eventuallyf(t, func() bool {
-				return kubeclient.Get(ctx, client.ObjectKeyFromObject(existingCrd), existingCrd) == nil
+				if err := kubeclient.Get(ctx, client.ObjectKeyFromObject(existingCrd), existingCrd); err != nil {
+					return false
+				}
+				return testutil.CrdReady(&existingCrd.Status)
 			}, defaultWaitPeriod, defaultTick, "failed to get initial crd for testing: %v", err)
 
-			// Creating existing CR
+			// Creating existing CR and wait for it to be created
 			err = kubeclient.Create(ctx, existingCr)
 			require.NoError(t, err, "failed to create initial cr for testing")
 			require.Eventuallyf(t, func() bool {
--- a/internal/crd/crd_test.go
+++ b/internal/crd/crd_test.go
@@ -6,7 +6,7 @@ import (
 	"time"
 
 	"github.com/operator-framework/rukpak/internal/unit"
-	"github.com/operator-framework/rukpak/internal/util"
+	"github.com/operator-framework/rukpak/test/testutil"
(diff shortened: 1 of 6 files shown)


internal/util/kubestructs.go
```diff
@@ -0,0 +1,39 @@
+package util
```
exdx, line 1: I'm not sure how I feel about introducing test utility functions inside the main codebase but since these are used by both unit and e2e tests I guess it makes sense. Can we have the file name include `test` in it so that its clear these are for test purposes? Maybe `testcrds` makes sense?

test/e2e/crdvalidator_test.go
```diff
@@ -30,19 +27,20 @@ var _ = Describe("crd validation webhook", func() {
...
-					util.GenName(defaultTestingCrdName),
-					[]apiextensionsv1.CustomResourceDefinitionVersion{{
-						Name:    "v1alpha1",
-						Served:  true,
-						Storage: true,
-						Schema: &apiextensionsv1.CustomResourceValidation{
-							OpenAPIV3Schema: &apiextensionsv1.JSONSchemaProps{
-								Type:        "object",
-								Description: "my crd schema",
+				crd = util.NewTestingCRD(defaultTestingCrdName, defaultTestingCrdGroup, true,
```
exdx, line 30: does it make sense to move the `defaultTestingCrdName` and other associated constants to the util package as well?

internal/util/kubestructs.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"fmt"
+
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+func NewTestingCRD(name, group string, unique bool, versions []apiextensionsv1.CustomResourceDefinitionVersion) *apiextensionsv1.CustomResourceDefinition {
```
exdx, line 11: This function signature is a little funny. WDYT about something like
```
if name == "" {
  name = GenName(defaultName)
} 
```
and get rid of the unique bool altogether?

internal/crd/crd_test.go
```diff
@@ -0,0 +1,181 @@
...
+
+			// Create existing CRD
+			err := kubeclient.Create(ctx, existingCrd)
+			require.NoError(t, err, "failed to create initial crd for testing")
+			require.Eventuallyf(t, func() bool {
+				return kubeclient.Get(ctx, client.ObjectKeyFromObject(existingCrd), existingCrd) == nil
+			}, defaultWaitPeriod, defaultTick, "failed to get initial crd for testing: %v", err)
+
+			// Creating existing CR
+			err = kubeclient.Create(ctx, existingCr)
```
exdx, line 162: Since `envtest` actually uses an api-server under the hood to create resources, this test may be racy where the CRD is not accepted and established on the cluster (but is GET-able) by the time the CR is created. See https://github.com/operator-framework/operator-lifecycle-manager/blob/c36183ef17f757dbf1f96a2619a2a91184afc579/test/e2e/bundle_e2e_test.go#L179 for an example of a function that ensures the CRD is available on the cluster and ready to accept CRs. We might want something like this in our codebase too.

internal/util/kubestructs.go
tylerslaton: What if we add a "util" package into the `test` directory along side `e2e`? That way these testing utility functions can have their own home away from the internal codebase. I tend to agree with you on this that it feels weird.

internal/crd/crd_test.go
tylerslaton: Great point. I think we could benefit from a function like this in the `testutil` package that this PR adds too.

exdx (approved):

PR #260, opened 2022-04-26
manifests: Add a base kustomization.yaml for the set of provisioner manifests

timflannagan:
Update the manifests/provisioner directory structure and add a base kustomization.yaml that only points to the plain provisioner kustomization right now. This allows people to run `kustomize build manifests/provisioners` locally without having to specify the manifests/provisioners/plain directory, which makes it easier to script around when generating release artifacts.

### DIFF
--- a/manifests/provisioners/kustomization.yaml
+++ b/manifests/provisioners/kustomization.yaml
@@ -1,0 +1,3 @@
+bases:
+- ./plain
+
--- a/Makefile
+++ b/Makefile
@@ -92,15 +92,15 @@ kind-cluster: ## Standup a kind cluster for e2e testing usage
 ###################
 # Install and Run #
 ###################
-.PHONY: install-apis install-plain install-crdvalidator-webhook install-core-webhook install-webhooks run cert-mgr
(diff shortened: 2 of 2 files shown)


manifests/provisioners/kustomization.yaml
```diff
@@ -0,0 +1,3 @@
+bases:
+- ./plain
+
```
timflannagan, line 3: Note: we could also add a kustomize-related Makefile target for installing the rukpak provisioners, e.g. `make install-provisioners`, given we already have a dedicated plain provisioner Makefile target. I know @joelanford is a big fan of introducing new targets too.

joelanford: Lol.  I'm a +1 to this PR if we remove the existing `install-plain-provisioner` and replace it with `install-provisioners`

timflannagan: Done - just ran the `install-provisioners` target locally and ran into no errors. I didn't see anything in our documentation that indicates we use any Makefile target outside of `make run` for quickstart purposes.

joelanford (approved):

tylerslaton (approved):
Nice, I like how we are using `kustomize` with this PR.

PR #264, opened 2022-04-26
Introduce framework to emit prometheus metrics from plain provisioner

### LINKED ISSUE #266: Introduce framework to allow emitting promethues metrics from the plain provisioner

anik120:
Note: This PR only introduces a gauge vector metric emitted from the bundle controller. 
Holding off on introducing metrics from the bundleInstance controller until [this is resovled](https://github.com/operator-framework/rukpak/blob/main/internal/provisioner/plain/controllers/bundleinstance_controller.go#L79-L82):
```
// TODO(user): Modify the Reconcile function to compare the state specified by
// the BundleInstance object against the actual cluster state, and then
// perform operations to make the cluster state reflect the state specified by
// the user.
```

Closes #266

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,6 @@ vendor/
 bin/
 dist/
 .goreleaser.yml
+
+# Editor specific directories
+.vscode/
--- a/api/v1alpha1/bundle_types.go
+++ b/api/v1alpha1/bundle_types.go
@@ -22,6 +22,8 @@ import (
 
 type BundleConditionType string
 
+type BundlePhase string
+
 const (
 	SourceTypeImage = "image"
 	SourceTypeGit   = "git"
@@ -33,10 +35,11 @@ const (
 	ReasonUnpackSuccessful = "UnpackSuccessful"
 	ReasonUnpackFailed     = "UnpackFailed"
 
-	PhasePending   = "Pending"
-	PhaseUnpacking = "Unpacking"
-	PhaseFailing   = "Failing"
-	PhaseUnpacked  = "Unpacked"
+	PhasePending   = BundlePhase("Pending")
+	PhaseUnpacking = BundlePhase("Unpacking")
+	PhaseFailing   = BundlePhase("Failing")
+	PhaseUnpacked  = BundlePhase("Unpacked")
+	PhaseUnknown   = BundlePhase("Unknown")
 )
 
 // BundleSpec defines the desired state of Bundle
--- a/internal/provisioner/plain/controllers/bundle_controller.go
+++ b/internal/provisioner/plain/controllers/bundle_controller.go
@@ -66,6 +66,15 @@ type BundleReconciler struct {
 	GitClientImage  string
 }
 
+type BundlePhaseTransitioner struct {
+	ctx         context.Context
+	bundle      *rukpakv1alpha1.Bundle
+	pod         *corev1.Pod
+	podOpResult controllerutil.OperationResult
+	u           *updater.Updater
+	err         error
+}
+
 //+kubebuilder:rbac:groups=core.rukpak.io,resources=bundles,verbs=list;watch
 //+kubebuilder:rbac:groups=core.rukpak.io,resources=bundles/status,verbs=update;patch
 //+kubebuilder:rbac:groups=core.rukpak.io,resources=bundles/finalizers,verbs=update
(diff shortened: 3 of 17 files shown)


timflannagan (commented):
Can we create an issue for introducing metrics to the plain provisioner?

anik120:
@timflannagan I opened #266 and edited the description to include the issue number too, not sure if github picked up on that though.

internal/provisioner/plain/metrics/metrics.go
```diff
@@ -0,0 +1,100 @@
...
+		prometheus.GaugeOpts{
+			Name: "bundle",
+			Help: "Bundle Infomation. A positive value represents successful bundle unpack. A zero value represents bundle unpacking is pending. A negative value represents bundle unpacking has failed.",
+		},
+		[]string{nameLabel, namespaceLabel},
+	)
+)
+
+// ServeMetrics enables the provisioner to serve prometheus metrics.
+func ServeMetrics() error {
```
joelanford, line 39: Rather than running a separate metrics server, we should register these metrics with the existing controller-runtime metrics server.

https://book.kubebuilder.io/reference/metrics.html#publishing-additional-metrics

api/v1alpha1/bundle_types.go
```diff
@@ -22,6 +22,8 @@ import (
 
 type BundleConditionType string
 
+type BundlePhase string
```
exdx, line 25: I see the reasoning behind this but it looks like we end having to coerce everything to `string` in the codebase and it doesn't look that great. Is using a custom phase type beneficial for metrics somehow? 

Happy to let others weigh in.

exdx: If we do want this, I think breaking it out into its own separate PR makes sense as well.

internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -95,27 +116,35 @@ func (r *BundleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ c
...
-		return ctrl.Result{}, updateStatusUnpackFailing(&u, fmt.Errorf("ensure unpack pod: %w", err))
-	} else if op == controllerutil.OperationResultCreated || op == controllerutil.OperationResultUpdated || pod.DeletionTimestamp != nil {
-		updateStatusUnpackPending(&u)
+	op, err := r.ensureUnpackPod(ctx, bundle, pod)
+	return r.transitionBundlePhase(BundlePhaseTransitioner{ctx, bundle, pod, op, &u, metricsRecorder, err})
+}
+
+func (r *BundleReconciler) transitionBundlePhase(pt BundlePhaseTransitioner) (_ ctrl.Result, reconcileErr error) {
+	// first record metric for phase bundle currently is in
+	pt.metricsRecorder.SetBundleMetric()
```
joelanford, line 125: Maybe I'm put off by the size of this PR at first glance and that's tainting my perception, but this seems like a somewhat heavyweight approach to simply incrementing/decrementing a single metric value.

There are two alternatives I have in mind, that I think would be much simpler
1. A totally separate controller whose only responsibility is to set metrics
2. A custom controller runtime handler that set metrics as events stream through the queue. We have something like that in `operator-lib`: https://github.com/operator-framework/operator-lib/blob/main/handler/instrumented_enqueue_object.go

joelanford:
This PR does more than what's described in #266 (it also actually adds a metric). Should we have an issue that discussed the actual metric(s) we want to add?

.gitignore
```diff
@@ -16,3 +16,6 @@ vendor/
 bin/
 dist/
 .goreleaser.yml
+
+# Editor specific directories
+.vscode/
```
timflannagan, line 21: Interesting: vscode is my main IDE and I haven't had to add the vscode hidden directory to any gitignore when working on git-enabled repositories. I poked at my global .gitignore as well and didn't see any .vscode items. Any idea on why this is necessarily? Mainly curious - this isn't blocking feedback.

internal/provisioner/plain/metrics/metrics.go
timflannagan: Yup +1 on using anything we can use that we get for free.

api/v1alpha1/bundle_types.go
anik120: So my instinct to turn this into a typed struct (BundlePhase) was to try and highlight the Phases of the Bundle resource more. The way the Reconcile is coded right now, it seems like phase transition is almost an after thought of reconciliation, whereas my argument is that phase transition is one of the primary goal of reconciliation. 
However, as I kept plumbing this I realized that just turning `Phase` from `string` to `BundlePhase` isn't enough, as highlighted by comment [here](https://github.com/operator-framework/rukpak/pull/264/files#diff-c484260008b9e8123e074e620f312d60c48375aea24e977e1dbf08c1fe13b4f6R98-R101). Reading your comment it's becoming more clear to me that it'll probably be cleaner to do that 👈🏽  first, before something like metrics emitting functionality is added, where you have to "do something before reconciliation"/"do something during reconciliation"/"do something after reconciliation" etc

.gitignore
anik120: My .vscode directory was actually being tracked until I added this 🤷🏽‍♂️

internal/provisioner/plain/metrics/metrics.go
anik120: Reasoning for a custom server explained here: https://github.com/operator-framework/rukpak/issues/266#issuecomment-1111598474

internal/provisioner/plain/controllers/bundle_controller.go
anik120: I have to look more closely at `operator-lib` and that particular handler, but before I spend time doing that, I'm curious to read the responses in the #266

tylerslaton:
@anik120 @operator-framework/rukpak-maintainers How do we want to proceed with this PR? We could put it in `lifecycle/frozen` if we want to keep it open for reference but we should figure out a longterm plan for this PR and its issue
https://github.com/operator-framework/rukpak/issues/266.

timflannagan:
We've already linked the PR back to the original issue so I think we're fine to close this until we re-prioritize that work.

PR #268, opened 2022-04-26
Optional CRD Validation

### LINKED ISSUE #245: Allow CRDs to opt-out of the CRD Validation admission webhook
The CRD Validation webhook provides a way to ensure that CRD upgrades are safe and do not cause any potential data loss. However, some users may want to opt-out of this behavior, since the CRD webhook is bundled with rukpak by default. Allowing users potentially destructive updates may be desirable in certain cases (test environments, the user is trying to fix a bad installation, etc) and the user should be able to do so in a straightforward way. 

Some ideas:
* Provide an annotation that, if set on the incoming CRD, skips the CRD validation entirely. The CRD webhook can be updated to filter only for CRDs that do not have this annotation set. For example `core.rukpak.io/crd-validate: false`.
* Provide a convenient way to turn off the CRD validation at the cluster scope, via some kind of CRDValidatorConfig object. The webhook looks at this object and, if it sees that the setting is set to disabled, stops performing validation. 
* Some property on the Bundle itself that gets propagated to the webhook (unclear)

### LINKED ISSUE #259: Only validate CRD updates for CRDs that are installed by RukPak
Currently the CRD Validation admission webhook that was implemented in #213 intercepts update events for all CRDs on the cluster, not just those that are related to a rukpak installation. There should be a way to limit the scope of the webhook down to validate just the CRDs that are installed as part of a Bundle and activated via a BI. This avoids side effects where the webhook intercepts CRD update events that are unrelated to rukpak installs and should not be validated. 

A/C:
* Implement some kind of metadata (for example, an annotation) that should be present on all CRDs that are being installed by RukPak provisioners
* Ensure that the CRD validation webhook only responds to CRD update events that have the metadata set

tylerslaton:
The intent of this PR is to give an opt-out approach for the `crdvalidator` which only runs against RukPak provisioned CRDs. `crdvalidator` will now check the `Labels` of an incoming CRD event for `core.rukpak.io/owner-kind` to determine if it is being created by RukPak and thus run the validation logic. To turn this off for a specific CRD made by RukPak, the following annotation needs to be set for the CRD within the `Bundle`:

```yaml
core.rukpak.io/safe-upgrade-validation: disabled
```

Closes #245 
Closes #259

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,3 +100,11 @@ a `spec.provisionerClassName` that matches that particular ID.
 For example, in this repository the [plain](internal/provisioner/plain/README.md) provisioner is implemented.
 The `plain` provisioner is able to unpack a given `plain+v0` bundle onto a cluster and then instantiate it, making
 the content of the bundle available in the cluster.
+
+### CustomResourceDefinition (CRD) Validator
+
+RukPak comes with a webhook for validating the upgrade of CRDs from `Bundle`s. If a CRD does potentially destructive
+actions to the cluster, it will not allow it go be applied. In the context of RukPak, this will result in a failed 
+`BundleInstance` resolution. 
+
+To read more about this webhook, and learn how to disable this default behavior, view the `crdvalidator` [documentation](cmd/crdvalidator/README.md).
--- a/cmd/crdvalidator/README.md
+++ b/cmd/crdvalidator/README.md
@@ -1,0 +1,106 @@
+# crdvalidator
+
+## Summary
+A part of the core value proposition for RukPak is the safe upgrade of a [CustomResourceDefinition](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/) (`CRD`). As a result, this repository comes equipped with a "CRD Validator Webhook" (`crdvalidator`), that will validate all `CRD` upgrades created by RukPak. This protects your `BundleInstance` pivots from having potentially dangerous effects such as data loss. 
+
+### How to use
+
+`crdvalidator` will ensure that any `CRD` upgrade is safe as long as the CRD is associated with RukPak. If you would like to disable validation for a `CRD` included in a specific `Bundle`, you will need to explicitly set the `core.rukpak.io/safe-crd-upgrade-validation` annotation to `false` within that specific CRD's manifest.
+
+```yaml
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    core.rukpak.io/safe-crd-upgrade-validation: false # <--
+```
+
+### What is protected
+There are three main cases `crdvalidator` is checking against with every specified `CRD` upgrade. These are where a new `CRD`:
+1. Removes a stored version and ensures that removing the stored version does not result in data loss.
+2. Changes a version that old `CRD` had and it must validate existing CRs against the new schema.
+3. Adds a version that old `CRD` does not have. In this case `crdvalidator` checks if the conversion strategy is:
+    - None, then ensures that existing CRs validate with new schema.
+    - Webhook, then allow update (assuming webhook handles conversion correctly)
+
+#### Update the schema that invalidates existing CRs
+Say that you have installed a `CRD` onto the cluster with a Kind of `Sample` that has a single property of `examplearray`. 
+
+```yaml
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.7.0
+    core.rukpak.io/safe-upgrade-validation: enabled
+  creationTimestamp: null
+  name: sample.example.io
+#
+# ...
+#
+spec:
+    versions:
+    - name: v1alpha1
+        schema:
+        openAPIV3Schema:
+            properties:
+                exampleProperty:
+                    description: exampleProperty is a string property to demonstrate crd validation logic.
+                    type: string
+```
+
+With this installed onto the cluster, you go ahead and create a `CR` of a `Sample` onto the cluster. 
+
+```yaml
+apiVersion: sample.example.io
+kind: Sample
+metadata:
+    name: sampleResource
+```
+
+If you then update that `CRD`'s schema in such a way that the `CR` you just created was invalidated, like so:
+
+```yaml
+#
+# ...
+#
+spec:
+    versions:
+    - name: v1alpha1
+        schema:
+        openAPIV3Schema:
+            type: string
+            required: # Updating the exampleProperty to be required
+                - exampleProperty
+            properties:
+                exampleProperty:
+                    description: exampleProperty is an array of strings to demonstrate crd validation logic.
+                    type: string
+```
+
+
+`crdvalidator` will prevent the update from occurring with the following error. This is because the existing `CR` does not have a value set for the required field of `exampleProperty`.
+
+```
+admission webhook "webhook.crdvalidator.io" denied the request: failed to validate safety of UPDATE for CRD "sample.example.io" (NOTE: to disable this validation, set the "core.rukpak.io/safe-crd-upgrade-validation" annotation to "false"): error validating existing CRs against new CRD's schema for "sample.example.io": existing custom object /sampleResource failed validation for new schema version v1alpha1: [].exampleProperty: Required value
+```
+
+## Running locally
+
+`crdvalidator` is included with every default installation of RukPak. To run this locally, simply run the make target with a kind cluster started.
+
+```console
+make run
+```
+
+By default, `crdvalidator`'s test suite runs alongside RukPak's, but you can also just run `crdvalidator`'s specific tests.
+
+```console
+make e2e TEST="crdvalidator"
+```
+
+If you would like to install the `crdvalidator` onto your current cluster without RukPak, you'll need to install `cert-mgr` and then `crdvalidator`
+
+```console
+make cert-mgr install-crdvalidator-webhook
+```
--- a/cmd/crdvalidator/annotation/annotation.go
+++ b/cmd/crdvalidator/annotation/annotation.go
@@ -1,0 +1,6 @@
+package annotation
+
+const (
+	ValidationKey = "core.rukpak.io/safe-crd-upgrade-validation"
+	Disabled      = "false"
+)
--- a/cmd/crdvalidator/handlers/crd.go
+++ b/cmd/crdvalidator/handlers/crd.go
@@ -22,7 +22,9 @@ import (
 	"net/http"
 
 	"github.com/go-logr/logr"
+	"github.com/operator-framework/rukpak/cmd/crdvalidator/annotation"
 	"github.com/operator-framework/rukpak/internal/crd"
+
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
@@ -49,17 +51,21 @@ func NewCrdValidator(log logr.Logger, client client.Client) CrdValidator {
 // a safe upgrade based on the crd.Validate() function call
 func (cv *CrdValidator) Handle(ctx context.Context, req admission.Request) admission.Response {
 	incomingCrd := &apiextensionsv1.CustomResourceDefinition{}
-
-	err := cv.decoder.Decode(req, incomingCrd)
-	if err != nil {
+	if err := cv.decoder.Decode(req, incomingCrd); err != nil {
 		message := fmt.Sprintf("failed to decode CRD %q", req.Name)
 		cv.log.V(0).Error(err, message)
 		return admission.Errored(http.StatusBadRequest, fmt.Errorf("%s: %w", message, err))
 	}
 
-	err = crd.Validate(ctx, cv.client, incomingCrd)
-	if err != nil {
-		message := fmt.Sprintf("failed to validate safety of %s for CRD %q: %s", req.Operation, req.Name, err)
+	// Check if the request should get validated
+	if disabled(incomingCrd) {
+		return admission.Allowed("")
+	}
+
+	if err := crd.Validate(ctx, cv.client, incomingCrd); err != nil {
+		message := fmt.Sprintf(
+			"failed to validate safety of %s for CRD %q (NOTE: to disable this validation, set the %q annotation to %q): %s",
+			req.Operation, req.Name, annotation.ValidationKey, annotation.Disabled, err)
 		cv.log.V(0).Info(message)
 		return admission.Denied(message)
 	}
@@ -72,4 +78,11 @@ func (cv *CrdValidator) Handle(ctx context.Context, req admission.Request) admis
 func (cv *CrdValidator) InjectDecoder(d *admission.Decoder) error {
 	cv.decoder = d
 	return nil
+}
+
+// disabled takes a CRD and checks its content to see crdvalidator
+// is disabled explicitly
+func disabled(crd *apiextensionsv1.CustomResourceDefinition) bool {
+	// Check if the annotation to disable validation is set
+	return crd.GetAnnotations()[annotation.ValidationKey] == annotation.Disabled
 }
--- a/manifests/crdvalidator/04_webhook.yaml
+++ b/manifests/crdvalidator/04_webhook.yaml
@@ -18,6 +18,9 @@ webhooks:
         name: crd-validation-webhook
         path: /validate-crd
         port: 9443
+    objectSelector:
+      matchLabels:
+        core.rukpak.io/owner-kind: BundleInstance
     admissionReviewVersions: ["v1"]
     sideEffects: None
 ---
--- a/test/e2e/crdvalidator_test.go
+++ b/test/e2e/crdvalidator_test.go
@@ -2,15 +2,20 @@ package e2e
 
 import (
 	"context"
+	"fmt"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"github.com/operator-framework/rukpak/cmd/crdvalidator/annotation"
+	"github.com/operator-framework/rukpak/internal/util"
 	"github.com/operator-framework/rukpak/test/testutil"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	rukpakv1alpha1 "github.com/operator-framework/rukpak/api/v1alpha1"
 )
 
-var _ = Describe("crd validation webhook", func() {
+var _ = Describe("crdvalidator", func() {
 	When("a crd event is emitted", func() {
 		var ctx context.Context
 
@@ -47,7 +52,7 @@ var _ = Describe("crd validation webhook", func() {
 				Expect(c.Delete(ctx, crd)).To(BeNil())
 			})
 
-			It("should allow the crd update event to occur", func() {
+			It("should allow the crd update event to occur without being owned by RukPak", func() {
 				Eventually(func() error {
 					if err := c.Get(ctx, client.ObjectKeyFromObject(crd), crd); err != nil {
 						return err
@@ -66,6 +71,125 @@ var _ = Describe("crd validation webhook", func() {
 							},
 						},
 					})
+
+					return c.Update(ctx, crd)
+				}).Should(Succeed())
+			})
+
+			It("should allow the crd update event to occur being owned by RukPak", func() {
+				Eventually(func() error {
+					if err := c.Get(ctx, client.ObjectKeyFromObject(crd), crd); err != nil {
+						return err
+					}
+
+					crd.Spec.Versions[0].Storage = false
+
+					crd.Spec.Versions = append(crd.Spec.Versions, apiextensionsv1.CustomResourceDefinitionVersion{
+						Name:    "v1alpha2",
+						Served:  true,
+						Storage: true,
+						Schema: &apiextensionsv1.CustomResourceValidation{
+							OpenAPIV3Schema: &apiextensionsv1.JSONSchemaProps{
+								Type:        "object",
+								Description: "my crd schema",
+							},
+						},
+					})
+
+					crd.Labels = map[string]string{util.CoreOwnerKindKey: rukpakv1alpha1.BundleInstanceKind}
+
+					return c.Update(ctx, crd)
+				}).Should(Succeed())
+			})
+		})
+
+		When("an incoming crd event modifies the schema in a way that breaks an existing cr", func() {
+			var crd *apiextensionsv1.CustomResourceDefinition
+
+			BeforeEach(func() {
+				crd = testutil.NewTestingCRD("", testutil.DefaultGroup,
+					[]apiextensionsv1.CustomResourceDefinitionVersion{{
+						Name:    "v1alpha1",
+						Served:  true,
+						Storage: true,
+						Schema: &apiextensionsv1.CustomResourceValidation{
+							OpenAPIV3Schema: &apiextensionsv1.JSONSchemaProps{
+								Type:        "object",
+								Description: "my crd schema",
+								Properties: map[string]apiextensionsv1.JSONSchemaProps{
+									"sampleProperty": {Type: "string"},
+								},
+							},
+						},
+					}},
+				)
+
+				crd.Labels = map[string]string{util.CoreOwnerKindKey: rukpakv1alpha1.BundleInstanceKind}
+
+				Eventually(func() error {
+					return c.Create(ctx, crd)
+				}).Should(Succeed(), "should be able to create a safe crd but was not")
+
+				// Build up a CR to create out of unstructured.Unstructured
+				sampleCR := testutil.NewTestingCR(testutil.DefaultCrName, testutil.DefaultGroup, "v1alpha1", crd.Spec.Names.Singular)
+				Eventually(func() error {
+					return c.Create(ctx, sampleCR)
+				}).Should(Succeed(), "should be able to create a cr for the sample crd but was not")
+
+			})
+
+			AfterEach(func() {
+				By("deleting the testing crd")
+				Expect(c.Delete(ctx, crd)).To(BeNil())
+			})
+
+			It("should be invalidated if owned by RukPak", func() {
+				Eventually(func() string {
+					if err := c.Get(ctx, client.ObjectKeyFromObject(crd), crd); err != nil {
+						return err.Error()
+					}
+
+					// Update the v1alpha1 schema to invalidate existing CR created in BeforeEach()
+					crd.Spec.Versions[0].Schema.OpenAPIV3Schema.Required = []string{"sampleProperty"}
+
+					err := c.Update(ctx, crd)
+					fmt.Println(err)
+					if err != nil {
+						return err.Error()
+					}
+					return ""
+				}).Should(ContainSubstring("error validating existing CRs against new CRD's schema"))
+			})
+
+			It("should be admitted if not owned by RukPak", func() {
+				Eventually(func() error {
+					if err := c.Get(ctx, client.ObjectKeyFromObject(crd), crd); err != nil {
+						return err
+					}
+
+					// Remove the "core.rukpak.io/owner-kind" label
+					crd.Labels = map[string]string{}
+					Expect(c.Update(ctx, crd)).To(BeNil())
+
+					// Update the v1alpha1 schema to invalidate existing CR created in BeforeEach()
+					crd.Spec.Versions[0].Schema.OpenAPIV3Schema.Required = []string{"sampleProperty"}
+
+					return c.Update(ctx, crd)
+				}).Should(Succeed())
+			})
+
+			It("should be admitted if owned by RukPak but disabled explicitly by annotation", func() {
+				Eventually(func() error {
+					if err := c.Get(ctx, client.ObjectKeyFromObject(crd), crd); err != nil {
+						return err
+					}
+
+					// Add the "core.rukpak.io/safe-upgrade-validation" label set to "disabled"
+					crd.Annotations = map[string]string{annotation.ValidationKey: annotation.Disabled}
+					Expect(c.Update(ctx, crd)).To(BeNil())
+
+					// Update the v1alpha1 schema to invalidate existing CR created in BeforeEach()
+					crd.Spec.Versions[0].Schema.OpenAPIV3Schema.Required = []string{"sampleProperty"}
 
 					return c.Update(ctx, crd)
 				}).Should(Succeed())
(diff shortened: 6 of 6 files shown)


cmd/crdvalidator/handlers/crd.go
```diff
@@ -28,6 +28,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 )
 
+const ValidateAnnotation = "validatecrd"
```
exdx, line 31: could we name this something more kube-like? I know we want to keep the crd as sort of it's own component, but does something generic like `crdvalidation.webhook.io` make sense?

cmd/crdvalidator/annotation/annotation.go
```diff
@@ -0,0 +1,16 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func Add(obj client.Object) {
+	// Add the annotation that crdvalidator is looking for
+	annotations := obj.GetAnnotations()
+	if annotations == nil {
+		annotations = map[string]string{}
+	}
+	annotations[handlers.ValidateAnnotation] = "true"
```
exdx, line 14: nit: I think `"enabled"` makes a bit more sense than `"true"` in this context? Since we are specifying that the webhook should be active for this particular CRD. WDYT?

cmd/crdvalidator/handlers/crd.go
```diff
@@ -57,6 +59,11 @@ func (cv *CrdValidator) Handle(ctx context.Context, req admission.Request) admis
 		return admission.Errored(http.StatusBadRequest, fmt.Errorf("%s: %w", message, err))
 	}
 
+	// Only run validation on CRDs with the ValidateAnnotation set to true
+	if val, ok := incomingCrd.GetAnnotations()[ValidateAnnotation]; !ok || val != "true" {
```
exdx, line 63: I was thinking, instead of watching all CRD events, and then filtering out the ones for CRDs that don't have the annotation set, we could build a predicate into the webhook's cache such that it only receives events for CRDs that have this annotation/label set. That approach is more scalable (since we don't see events we don't care about) and intuitive, since there is no filtering in the webhook code itself.

internal/provisioner/plain/controllers/bundleinstance_controller.go
```diff
@@ -45,9 +45,11 @@ import (
 	"sigs.k8s.io/yaml"
 
 	rukpakv1alpha1 "github.com/operator-framework/rukpak/api/v1alpha1"
+	crdValidatorAnnotation "github.com/operator-framework/rukpak/cmd/crdvalidator/annotation"
```
exdx, line 48: nit: this import could be lowercase and more condensed, at the expense of clarity, to be more conformant to go import conventions. Something like `crdanno`?

exdx (commented):
Nice! Good to get this in so soon.

cmd/crdvalidator/handlers/crd.go
tylerslaton: Good call, that makes absolute sense. Let me iterate on this and see what I can do to that end.

cmd/crdvalidator/handlers/crd.go
tylerslaton: Yep, I agree. I think this annotation could also be used for the different levels we are talking about once-upon-a-time so this more generic approach sounds better.

timflannagan (commented):
Just a quick review since this is still marked as WIP:

- I _think_ I'm okay with an annotation driving this functionality but I'd like to see if we can find a cleaner solution over time
- I'd expect that this functionality would be opt-out vs. opt-in like the current implementation reads
- Can we add documentation around optional CRD validation? This can be a follow-up candidate too as long as we track that effort.

Thoughts?

tylerslaton:
> Just a quick review since this is still marked as WIP:
> 
> * I _think_ I'm okay with an annotation driving this functionality but I'd like to see if we can find a cleaner solution over time
> * I'd expect that this functionality would be opt-out vs. opt-in like the current implementation reads
> * Can we add documentation around optional CRD validation? This can be a follow-up candidate too as long as we track that effort.
> 
> Thoughts?

1. What would you see as a potentially cleaner solution to an annotation here? I'd rather hash that conversation out now rather than later before we start establishing a UX.
2. What would your thoughts be around this being opt-out only for Rukpak `Bundle`s? In the current implementation we are making it so every `Bundle` with a `CRD` in it creates that `CRD` with the `crdvalidator` annotation. As a result, all `CRD`s created by Rukpak are validated through the webhook. I've labelled this as WIP as there is not currently a way to turn this functionality off within Rukpak and thus wanted to add an opt-out knob. The alternative would be to have the `crdvalidator` be opt-out for the entire cluster.
3. Absolutely, I would say that probably makes sense as a follow-up PR and might even make sense for there to be a separate README inside of the `crdvalidator` cmd directory.

exdx:
I think it has to be opt-in to ensure that the validator webhook doesn't accidentally intercept updates to CRDs that are not associated with a particular BI right? We have to keep the webhook scope limited to just the CRDs associated with rukpak, per the feedback from the control plane call.

timflannagan:
@exdx I was under the impression that we attach metadata to CRDs managed by a BundleInstance resource? Can we filter out CRDs by something like a label selector, or check whether there's an owner reference present pointing to a rukpak-related resource as an initial pass in the validation chain? We could then build on top of that logic and check whether that opt-out annotation is present.

exdx:
>  Can we filter out CRDs by something like a label selector, or check whether there's an owner reference present pointing to a rukpak-related resource as an initial pass in the validation chain? We could then build on top of that logic and check whether that opt-out annotation is present.

I see, that makes sense 👍

cmd/crdvalidator/annotation/annotation.go
```diff
@@ -0,0 +1,23 @@
+package annotation
+
+import (
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	ValidationKey = "core.rukpak.io/enable-safe-upgrade-validation"
```
exdx, line 8: nit: now that the flag is `enabled` I don't think we should have `enable` in the key as well. I also think we should include the word `crd` in here just so it's more clear what is being validated. Also, in the control plane call the preferred term was "update" instead of "upgrade" since there is no upgrade verb and it's more clear. So I think we can tweak this a bit more.

How does something like `"core.rukpak.io/safe-crd-updates"` sound?

exdx (approved):
LGTM. I would still prefer if this webhook were to only watch the CRDs associated with a rukpak install, versus all CRDs on-cluster, but it doesn't have to stop this going through. We do end up with an opt-out mechanism for free, which is nice.

cmd/crdvalidator/annotation/annotation.go
tylerslaton: I agree so much that I did this on accident before reading this comment 😆 Just pushed it in the most recent change 👍🏻

internal/provisioner/plain/controllers/bundleinstance_controller.go
```diff
@@ -45,9 +45,11 @@ import (
 	"sigs.k8s.io/yaml"
 
 	rukpakv1alpha1 "github.com/operator-framework/rukpak/api/v1alpha1"
+	crdanno "github.com/operator-framework/rukpak/cmd/crdvalidator/annotation"
 	helmpredicate "github.com/operator-framework/rukpak/internal/helm-operator-plugins/predicate"
 	"github.com/operator-framework/rukpak/internal/storage"
 	"github.com/operator-framework/rukpak/internal/util"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
```
timflannagan, line 52: nit: this import belongs with the 2nd block of package imports (e.g. stdlib, third party, internal)

internal/provisioner/plain/controllers/bundleinstance_controller.go
```diff
@@ -143,6 +145,11 @@ func (r *BundleInstanceReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		Metadata: &chart.Metadata{},
 	}
 	for _, obj := range desiredObjects {
+		// Add required crdvalidator annotation if it is a CRD
+		if obj.GetObjectKind().GroupVersionKind().GroupVersion() == apiextensionsv1.SchemeGroupVersion {
+			crdanno.Add(obj)
+		}
```
timflannagan, line 151: I think I'd be slightly surprised as a user with this UX where we inject runtime annotations to objects but I won't block a PR over this kind of behavior.

tylerslaton: I agree with you. Let me try an implementation where we look at the owner reference for incoming CRDs and only run the `crdvalidator` against that.

tylerslaton: Okay, the most recent PR added this functionality. Now the provisioner is not adding anything and instead the `crdvalidator` is watching the `OwnerReference` of incoming CRDs. If it is owned by Rukpak or has the annotation set  to `enabled` it will run the validation. If the CRD has the annotation set as `disabled`, it will not run the validation. This keeps it so we're not editing annotations and have a valve for opting out.

cmd/crdvalidator/README.md
```diff
@@ -0,0 +1,116 @@
+# crdvalidator
+
+## Summary
+A part of the core value proposition for RukPak is the safe upgrade of `CRD`s. As a result, this repository comes equipped with a "CRD Validator Webhook", `crdvalidator`, that will validate specified `CRD` upgrade events.
```
exdx, line 4: nit: could we use the full name of the resource and then use the abbreviation from then on? So CustomResourceDefinition (CRD). Also maybe a link to the upstream k8s docs on CustomResourceDefinitions could be helpful here as well?

cmd/crdvalidator/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+
+### How to use
+
+The CRD Validator webhook will ensure that any CRD upgrade is safe as long as the incoming CRD has the `core.rukpak.io/enable-safe-upgrade-validation` annotation set to `enabled` or if it is owned by RukPak via an owner reference.
```
exdx, line 9: nit: `owned by a rukpak provisioner` is more clear.

I guess we make an implicit assumption that other provisioners will implement the same ownerref pattern for objects that they manage. I think it's safe to make that assumption now but might be worth calling out in some kind of "provisioner suggested patterns" doc somewhere down the line.

cmd/crdvalidator/README.md
```diff
@@ -0,0 +1,116 @@
...
+    core.rukpak.io/safe-upgrade-validation: disabled # <--
+# 
+# ...
+#
+```
+
+
+### What is protected
+There are three main cases `crdvalidator` is checking against with every specified `CRD` upgrade. These are where a new `CRD`:
+1. Removes a version that the old `CRD` had and it must ensured that nothing is stored at the removed version.
```
exdx, line 40: nit: `Removes a stored version and ensures that removing the stored version does not result in data loss` for example?

cmd/crdvalidator/README.md
```diff
@@ -0,0 +1,116 @@
...
+`crdvalidator` is included with every default installation of RukPak. To run this locally, simply run the make target with a kind cluster started.
+
+```console
+$ make run
+```
+
+You can additionally run the `crdvalidator`'s e2e suite along side of RukPak's.
+
+```console
+$ make e2e
```
exdx, line 109: Should it be `make e2e TEST='crd validation webhook'`  if they want to run those tests specifically?

cmd/crdvalidator/README.md
```diff
@@ -0,0 +1,116 @@
...
+```
+
+If you would like to install the `crdvalidator` onto your current cluster without RukPak, you can easily do this with the appropriate make target.
```
exdx, line 112: Should we mention the cert-manager dependency? Right now we refactored the Makefile to pull the cert-manager dependency out of the install path of the individual webhook. Might be helpful to mention `make cert-manager` to install cert-manager as a prereq.

test/testutil/structs.go
```diff
@@ -62,3 +64,15 @@ func CrdReady(status *apiextensionsv1.CustomResourceDefinitionStatus) bool {
 	}
 	return established && namesAccepted
 }
+
+func NewRukpakOwnerReference(name string, uid types.UID) []metav1.OwnerReference {
+	trueRef := true
+	return []metav1.OwnerReference{{
+		APIVersion:         "core.rukpak.io/v1alpha1",
```
exdx, line 71: nit: is it possible to get this from the api definition? For example I think `rukpakv1alpha1.GroupVersion.String()` should equal this.

test/testutil/structs.go
```diff
@@ -62,3 +64,15 @@ func CrdReady(status *apiextensionsv1.CustomResourceDefinitionStatus) bool {
...
+		BlockOwnerDeletion: &trueRef,
+		Controller:         &trueRef,
+		Kind:               "BundleInstance",
```
exdx, line 74: nit: same here, we can use the new `rukpakv1alpha1.BundleInstanceKind` here

cmd/crdvalidator/README.md
exdx, line 4: I think it would also be helpful to give a background on why CRD updates are potentially risky and some of the nuances involved. Just so someone reading this immediately understands why the webhook is useful, especially if they are not already well-versed in the problem. Could be done as a follow-up though.

exdx (commented):
Nice, just some additional thoughts.

cmd/crdvalidator/README.md
tylerslaton: Yeah, we should probably create a provisioner onboarding document at some point. Probably not a huge issue now as we move closer to GA but will become important.

cmd/crdvalidator/README.md
tylerslaton: Hm, yeah it is probably a good idea to bring that up.

README.md
```diff
@@ -100,3 +100,8 @@ a `spec.provisionerClassName` that matches that particular ID.
 For example, in this repository the [plain](internal/provisioner/plain/README.md) provisioner is implemented.
 The `plain` provisioner is able to unpack a given `plain+v0` bundle onto a cluster and then instantiate it, making
 the content of the bundle available in the cluster.
+
+# Utilized Webhooks
+- [crdvalidator](cmd/crdvalidator/README.md)
+- [cert-manager](https://cert-manager.io/docs/)
+- [core](cmd/core/main.go)
```
timflannagan, line 107: nit: I think this would read weird to me if I was checking out the rukpak repository for the first time. The README.md is still a WIP so I won't block anything, but it might be nice to start working towards outlining the CRD validation component and the direct dependency on cert manager in the future.

cmd/crdvalidator/handlers/crd.go
```diff
@@ -73,3 +78,20 @@ func (cv *CrdValidator) InjectDecoder(d *admission.Decoder) error {
...
+	if annotation, ok := crd.GetAnnotations()[crdanno.ValidationKey]; ok {
+		return annotation == crdanno.Enabled
+	}
+
+	// Read the owner reference and check to see if it is owned by RukPak
+	for _, ref := range crd.OwnerReferences {
+		if ref.APIVersion == rukpakv1alpha1.GroupVersion.String() {
+			return true
+		}
+	}
```
timflannagan, line 95: I think I'd prefer if we check for the presence of known labels used by the controllers instead of owner references.

Testing this locally with a Bundle/BI that references the main branch of the combo repository, we should be in the clear to use label selectors:

```console
$ kubectl get crds --show-labels
combinations.combo.io                 2022-04-28T22:06:22Z   app.kubernetes.io/managed-by=Helm,core.rukpak.io/owner-kind=BundleInstance,core.rukpak.io/owner-name=combo
```

Where the relevant labels are `core.rukpak.io/owner-kind=BundleInstance,core.rukpak.io/owner-name=combo`.

test/e2e/crdvalidator_test.go
```diff
@@ -221,5 +224,141 @@ var _ = Describe("crd validation webhook", func() {
...
+									"sampleProperty": {Type: "string"},
+								},
+							},
+						},
+					}},
+				)
+				crd.OwnerReferences = testutil.NewRukpakOwnerReference(bi.GetName(), bi.Kind, bi.GetUID())
+				Eventually(func() error {
+					return c.Create(ctx, crd)
+				}).Should(Succeed(), "should be able to create a safe crd but was not")
```
timflannagan, line 287: This shouldn't be necessary as we should be placing owner references on all underlying Bundle source contents.

test/e2e/crdvalidator_test.go
```diff
@@ -221,5 +224,141 @@ var _ = Describe("crd validation webhook", func() {
...
+
+			AfterEach(func() {
+				By("deleting the testing bundle")
+				Expect(c.Delete(ctx, bundle)).To(BeNil())
+
+				By("deleting the testing bundleinstance")
+				Expect(c.Delete(ctx, bi)).To(BeNil())
+
+				By("deleting the testing crd")
+				Expect(c.Delete(ctx, crd)).To(BeNil())
```
timflannagan, line 304: Same comment as above: an explicit delete shouldn't be necessary as we place an owner reference on a CRD for activated Bundles.

test/e2e/crdvalidator_test.go
```diff
@@ -221,5 +224,141 @@ var _ = Describe("crd validation webhook", func() {
...
+							OpenAPIV3Schema: &apiextensionsv1.JSONSchemaProps{
+								Type:        "object",
+								Description: "my crd schema",
+								Properties: map[string]apiextensionsv1.JSONSchemaProps{
+									"sampleProperty": {Type: "string"},
+								},
+							},
+						},
+					}},
+				)
```
timflannagan, line 283: Is this kind of logic needed, or can we pivot to a model where we have a CRD bundle manifest that contains this annotation, and let the Bundle/BI controllers stamp out those resources for us?

README.md
timflannagan: That's a good question: maybe we provide a brief overview of the component in the root README.md, and link out to a more detailed README.md that's nested in the crdvalidators codebase? This can all be done in a follow-up though.

test/e2e/crdvalidator_test.go
tylerslaton: Good point 👍🏻

test/e2e/crdvalidator_test.go
tylerslaton: So basically just have it to where we are making a Bundle and BundleInstance where the Bundle contains these resources?

joelanford:
Something else (non-blocking) to think about: Right now I think we're still building all of the binaries (this one included) and putting them in a single plain provisioner image. Should we rename that image `rukpak` or `rukpak-core` or start producing separate images for the separate concerns?

I think for me, the former option would be a pretty good first step.

timflannagan:
@joelanford Is it worth creating an issue for this? I don't think we're at the stage yet where having multiple images is needed in terms of performance as the overall container image size isn't crazy large, at least when you compare it to something like OLM's singleton image.

cmd/crdvalidator/README.md
```diff
@@ -0,0 +1,116 @@
+# crdvalidator
+
+## Summary
+A part of the core value proposition for RukPak is the safe upgrade of a [CustomResourceDefinition](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/) (`CRD`). As a result, this repository comes equipped with a "CRD Validator Webhook", `crdvalidator`, that will validate specified `CRD` upgrade events. This protects your CRD upgrades from having potentially dangerous effects such as data loss. 
+
+
+### How to use
+
+The CRD Validator webhook will ensure that any CRD upgrade is safe as long as the incoming CRD has the `core.rukpak.io/enable-safe-upgrade-validation` annotation set to `enabled` or if it is owned by a RukPak provisioner.
```
joelanford, line 9: I wonder if `safe-upgrade-validation` is misleading. I'm sure there are upgrades this webhook would deem safe that in fact are not safe (e.g. a conversion webhook in the CRD does something wrong -- it would be impossible for us to tell).

Should we have explicit annotations for each of the checks that we _are_ doing? For example: 
- `core.rukpak.io/block-removal-of-inuse-stored-version: true`
- `core.rukpak.io/block-schema-tightening: true`

joelanford:
@timflannagan maybe an issue to rename the image to `rukpak`? A singleton image is nice from the perspective of CI and release.  I'd potentially even be onboard with a single binary with subcommands for the various components.

I'll file a new issue.

cmd/crdvalidator/README.md
joelanford: Seems reasonable, since we're still on the "breaking changes are what we do" train :)

timflannagan: Created #289, and made a note about the uncertainly around whether it makes sense to expose this kind of configuration behavior. These kinds of configuration knobs appear to be appropriate for upstream environments, but potentially problematic in downstream environments.

cmd/crdvalidator/handlers/crd.go
tylerslaton: I went ahead and updated the manifests for the webhook to only receive CRDs that are owned by RukPak. If that CRD has the `core.rukpak.io/safe-upgrade-validation` annotation set to `disabled` it will not validate it. Otherwise, the validation logic will run.

test/e2e/crdvalidator_test.go
tylerslaton: Updated the most recent tests to a place where this is no longer an issue.

test/e2e/crdvalidator_test.go
tylerslaton: Updated the tests to where this is no longer an issue.

cmd/crdvalidator/handlers/crd.go
tylerslaton: Implemented 🔥 Great suggestion.

cmd/crdvalidator/README.md
```diff
@@ -0,0 +1,96 @@
+# crdvalidator
+
+## Summary
+A part of the core value proposition for RukPak is the safe upgrade of a [CustomResourceDefinition](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/) (`CRD`). As a result, this repository comes equipped with a "CRD Validator Webhook" (`crdvalidator`), that will validate all `CRD` upgrades created by RukPak. This protects your `BundleInstance` pivots from having potentially dangerous effects such as data loss. 
+
+### How to use
+
+`crdvalidator` will ensure that any `CRD` upgrade is safe as long as it is owned by RukPak. If you would like to disable validation for a `CRD` included in a specific `Bundle`, you will need to explicitly set the `core.rukpak.io/safe-upgrade-validation` annotation to `disabled` within that specific `Bundle`.
```
exdx, line 8:
```suggestion
`crdvalidator` will ensure that any `CRD` upgrade is safe as long as the CRD is associated with RukPak. If you would like to disable validation for a `CRD` included in a specific `Bundle`, you will need to explicitly set the `core.rukpak.io/safe-upgrade-validation` annotation to `disabled` within that specific CRD's manifest.
```

I also _really_ think we should include `crd` in the annotation key so that it's more clear what this annotation applies to. Do we feel strongly it should not be included?

cmd/crdvalidator/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+```yaml
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.7.0
+    core.rukpak.io/safe-upgrade-validation: enabled
+  creationTimestamp: null
+  name: templates.combo.io
```
exdx, line 37: nit: since in the description on L27 you reference "a CRD with a Kind of `Sample`" should the name of this CRD be something like `sample.example.io` instead?

cmd/crdvalidator/README.md
```diff
@@ -0,0 +1,96 @@
...
+            properties:
+                examplearray:
+                    description: Examplearray is an array of strings to demonstrate crd validation logic.
+                    type: array
+                    minItems: 1
+                    items:
+                    type: string
+```
+
+With this installed onto the cluster, you go ahead and create a `CR` of a `Sample` onto the cluster. If you were to then update that `CRD`'s schema in such a way that that `CR` you created was invalidated, like so:
```
exdx, line 55: should we actually show a `Sample` manifest here, or is this just for the sake of example? Is there any intent that the user would actually follow this example?

cmd/crdvalidator/README.md
```diff
@@ -0,0 +1,96 @@
...
+                examplearray:
+                    description: Examplearray is an array of strings to demonstrate crd validation logic.
+                    type: array
+                    minItems: 100
+                    items:
+                    type: string
+```
+
+
+`crdvalidator` will prevent the update from occurring if the existing `CR` does not have at least 100 items in the `examplearray` field.
```
exdx, line 76: Might be helpful to include the specific error that the webhook returns so users can be familiar with this type of error up front.

cmd/crdvalidator/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Running locally
+
+`crdvalidator` is included with every default installation of RukPak. To run this locally, simply run the make target with a kind cluster started.
+
+```console
+$ make run
```
exdx, line 83: nit: I think we decided to pull the `$` out of our bash documentation based on previous PRs

cmd/crdvalidator/README.md
```diff
@@ -0,0 +1,96 @@
...
+```
+
+By default, `crdvalidator`'s test suite runs alongside RukPak's, but you can also just run `crdvalidator`'s specific tests.
+
+```console
+$ make e2e TEST="crdvalidator"
```
exdx, line 89: nit: same here wrt removing the `$`

cmd/crdvalidator/annotation/annotation.go
```diff
@@ -0,0 +1,7 @@
+package annotation
+
+const (
+	ValidationKey = "core.rukpak.io/safe-upgrade-validation"
+	Enabled       = "enabled"
```
joelanford, line 5: Apologies if this question has already been asked, but was there discussion about `"enabled"`/`"disabled"` vs `"true"`/`"false"` values for this annotation. I think `"true"`/`"false"` is what users would generally expect, but if there's a plan to have more than just `"enabled"`/`"disabled"` values, I could see an argument for non-boolean enums.

cmd/crdvalidator/handlers/crd.go
```diff
@@ -73,3 +77,11 @@ func (cv *CrdValidator) InjectDecoder(d *admission.Decoder) error {
 	cv.decoder = d
 	return nil
 }
+
+// enabled takes a CRD and checks its content to see crdvalidator
+// is disabled explicitly
+func enabled(crd *apiextensionsv1.CustomResourceDefinition) bool {
```
exdx, line 83: based on the description of the function, I would expect this code to look for
```
crd.GetAnnotations()[annotation.ValidationKey] == annotation.Disabled 
```


I suggest a change to the comment or the implementation. I think changing this function to be called `disabled()` and checking for the specific disabled condition makes the most sense, if we want to be strict about in what cases CRD validation is disabled.

cmd/crdvalidator/handlers/crd.go
```diff
@@ -73,3 +77,11 @@ func (cv *CrdValidator) InjectDecoder(d *admission.Decoder) error {
...
+	// Check if the annotation to disable validation is set
+	return crd.GetAnnotations()[annotation.ValidationKey] == annotation.Enabled ||
+		crd.GetAnnotations()[annotation.ValidationKey] == ""
```
joelanford, line 86: I think if we're thinking of this CRD validator as a standalone thing, it may actually make more sense to require opting into it.

If we think opt-in is "bad" for a specific rukpak scenario, we can always have a provisioner implementation inject this annotation automatically or based on a `BundleInstance` annotation.

cmd/crdvalidator/handlers/crd.go
```diff
@@ -49,16 +51,18 @@ func NewCrdValidator(log logr.Logger, client client.Client) CrdValidator {
...
-	if err != nil {
+	// Check if the request should get validated
+	if !enabled(incomingCrd) {
+		return admission.Allowed("")
+	}
+
+	if err := crd.Validate(ctx, cv.client, incomingCrd); err != nil {
 		message := fmt.Sprintf("failed to validate safety of %s for CRD %q: %s", req.Operation, req.Name, err)
 		cv.log.V(0).Info(message)
 		return admission.Denied(message)
```
joelanford, line 68: Since this webhook is operating on a built-in type (where no upstream documentation will exist about this webhook), should we include something in the deny message that mentions the annotation that's used? That way a user can know what to change to get a different answer from this webhook.

cmd/crdvalidator/annotation/annotation.go
tylerslaton: https://github.com/operator-framework/rukpak/pull/268#discussion_r859147932 

There was some discussion about it here ^ That being said, I'm not sure if this is the field we would use to have different levels of validation like it was originally intended.

cmd/crdvalidator/handlers/crd.go
tylerslaton: So we had an implementation that did just that but it got some push back to be opt-out, specifically for RukPak. If we feel it is a good idea I don't mind going back to the opt-in mechanic for the entire cluster but having RukPak inject the annotation. I think there was also some concern around the fact that we're adding annotations to CRD's that did not previously exist in the bundle prior to resolving.

cmd/crdvalidator/handlers/crd.go
tylerslaton: Sure, that sounds fair.

manifests/crdvalidator/04_webhook.yaml
```diff
@@ -18,6 +18,9 @@ webhooks:
         name: crd-validation-webhook
         path: /validate-crd
         port: 9443
+    objectSelector:
```
exdx, line 21: 👍

exdx (commented):
Looks good, nothing blocking

cmd/crdvalidator/handlers/crd.go
joelanford: Yeah it seems like the feedback was that it's less useful in non-rukpak contexts. And that's actually why I changed my mind and think it needs to be opt-in. Imagine if a cluster admin installed this webhook because they wanted rukpak. That would mean that all existing and new CRDs would need to consider this annotation, regardless of whether that CRD was/is managed by rukpak.

tylerslaton: Ah, so actually there is a new change that this webhook _only_ receives CRDs owned by a BundleInstance. Accordingly, regular CRD's on the cluster will not be validated against this webhook. The only time you need to worry about this annotation is if you want to disable CRD validation for a CRD within a RukPak Bundle. 

[See this change for reference](https://github.com/operator-framework/rukpak/blob/e38a5a36ff9f3d0ecc637fa1c9d433b8d4989f93/manifests/crdvalidator/04_webhook.yaml#L21-L23).

cmd/crdvalidator/README.md
tylerslaton: Sure, that's an easy fix. I'll also add Joe's "true"/"false" suggestion while I'm at it.

cmd/crdvalidator/README.md
tylerslaton: Good catch 👍🏻

cmd/crdvalidator/README.md
tylerslaton: That would probably make this a bit clearer and flow more naturally. I'll add that.

cmd/crdvalidator/annotation/annotation.go
tylerslaton: I agree with you that it should be "true"/"false" as it more closely aligns with common standards. Added the change.

cmd/crdvalidator/handlers/crd.go
joelanford: Ok, that solves that concern then. In that case, I think I'm good with this merging.

I think I would still like to revisit the opt-in/opt-out question (even regardless of it _just_ applying to rukpak) as its own thing, mainly just to make sure we have a focused discussion about _just_ that without also thinking about the validation itself, which I think we all agree is pretty useful.

timflannagan: Maybe this is something worth bubbling up as a topic in the upstream OLM WG meeting later this week?

exdx (approved):

manifests/crdvalidator/04_webhook.yaml
```diff
@@ -18,6 +18,9 @@ webhooks:
...
+      matchLabels:
+        core.rukpak.io/owner-kind: BundleInstance
```
timflannagan, line 23: This seems like it would also close out #259 too? Should we update the PR description and also add a "Closes ..." entry for that issue?

test/e2e/crdvalidator_test.go
```diff
@@ -2,15 +2,19 @@ package e2e
 
 import (
 	"context"
+	"fmt"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	rukpakv1alpha1 "github.com/operator-framework/rukpak/api/v1alpha1"
```
timflannagan, line 9: supernit: any of the rukpak related packages should live in the last section of the package imports.

test/testutil/constants.go
```diff
@@ -1,7 +1,14 @@
...
+
 const (
 	DefaultCrdName = "samplecrd"
 	DefaultCrName  = "samplecr"
 	DefaultGroup   = "e2e.io"
 )
+
+var (
+	CrdValidationEnabled  = map[string]string{crdanno.ValidationKey: crdanno.Enabled}
+	CrdValidationDisabled = map[string]string{crdanno.ValidationKey: crdanno.Disabled}
```
timflannagan, line 13: Do we use either of these variables anywhere? I wasn't able to find their usage throughout this diff.

cmd/crdvalidator/annotation/annotation.go
```diff
@@ -0,0 +1,7 @@
+package annotation
+
+const (
+	ValidationKey = "core.rukpak.io/safe-crd-upgrade-validation"
+	Enabled       = "true"
```
timflannagan, line 5: Do we use/set this variable anywhere? It looks like we just use the Disabled key throughout this diff?

manifests/crdvalidator/04_webhook.yaml
tylerslaton: It sure does, good catch.

test/testutil/constants.go
tylerslaton: We do not, these are left over from previous iterations. Will remove.

cmd/crdvalidator/annotation/annotation.go
tylerslaton: We don't, but wanted to keep it set incase we want to add `opt-in` functionality. To your point, we can add that when we need it.

tylerslaton:
> I had a couple more questions but nothing blocking. I'll let you decide whether you want to address any of them.

Addressed the changes. Once the CI passes i'll go ahead and merge.

PR #269, opened 2022-04-27
*: Avoid hardcoding the Bundle and BI metadata.Kind(s)

timflannagan:
Goal: quick cleanup PR that adds a global variable for the Bundle/BI GVK/Kind values so the repository can reference those values, and avoid hardcoding the string metadata.Kind values.

### DIFF
--- a/internal/provisioner/plain/main.go
+++ b/internal/provisioner/plain/main.go
@@ -90,7 +90,7 @@ func main() {
 		setupLog.Error(err, "unable to create kubernetes client")
 		os.Exit(1)
 	}
-	dependentRequirement, err := labels.NewRequirement("core.rukpak.io/owner-kind", selection.In, []string{"Bundle", "BundleInstance"})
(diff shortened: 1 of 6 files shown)


internal/provisioner/plain/main.go
```diff
@@ -90,7 +90,7 @@ func main() {
 		setupLog.Error(err, "unable to create kubernetes client")
 		os.Exit(1)
 	}
-	dependentRequirement, err := labels.NewRequirement("core.rukpak.io/owner-kind", selection.In, []string{"Bundle", "BundleInstance"})
+	dependentRequirement, err := labels.NewRequirement("core.rukpak.io/owner-kind", selection.In, []string{rukpakv1alpha1.BundleKind, rukpakv1alpha1.BundleInstanceKind})
```
timflannagan, line 93: Note: I wanted to keep those changes fairly small, but it looks like we have some potential cleanup candidates outside of these changes:

- Make the commonly used core.rukpak.io/owner-* labels constant variables, so we're not hardcoding these variables throughout the repository.
- Update this dependentRequirement variable used to build up the bundle/bI controller caches to be a interna/util function.

exdx (approved):
Nice!

PR #276, opened 2022-04-27
Generalize bundle image printcolumn

### LINKED ISSUE #163: Generalize Bundle Image printcolumn for non-image based content sources

exdx:
This PR is a bandaid solution to #163 which renames the displayed field from `Image` to `Reference` and moves it behind the `-o wide` output. It does not successfully display an arbitrary reference to an image or git repository, as JSONPath does not support conditional statements and there is no singular field to reference. Potentially once #164 merges there can be a follow-up that links `Reference` to a well-defined field on the Bundle status. 

Closes #163 
There will be a follow-up issue to move the displayed field back to the default output, instead of under `-o wide`.

### DIFF
--- a/api/v1alpha1/bundle_types.go
+++ b/api/v1alpha1/bundle_types.go
@@ -121,7 +121,7 @@ type BundleObject struct {
 //+kubebuilder:object:root=true
 //+kubebuilder:resource:scope=Cluster
 //+kubebuilder:subresource:status
-//+kubebuilder:printcolumn:name=Image,type=string,JSONPath=`.spec.source.image.ref`
+//+kubebuilder:printcolumn:name=Reference,type=string,JSONPath=`.spec.source.image.ref`,priority=1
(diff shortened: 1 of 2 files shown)


tylerslaton (approved):

timflannagan (approved):
Nice - I'm partial to just Ref but this works for me as well.

PR #278, opened 2022-04-27
tidy hack/tools module in "make tidy" target

joelanford:


timflannagan (approved):
Nice catch!

Issue #81, opened 2022-03-01
Support private bundle


exdx:
Currently there is no first-class way of supporting bundle images which are private, or underlying operator/operand images which are private as well. The initial prototype did include a `spec.Secrets` field which referenced secrets by name and copied them to the relevant namespace, but this was removed due to security concerns. 

Users of the system may store their bundle or underlying operator images in private registries. There should be an intuitive, documented way to use rukpak and the set of provisioners on-cluster with a Bundle image that is private, or if the underlying related images for a given BundleInstance are also private (for example, if the Deployment spec references a private image).

joelanford:
For sourcing bundle images, it seems like we could:
- add a `spec.source.image.imagePullSecretName` field to the Bundle API and document that the secret must be present in the namespace that the provisioner is deployed to. That way we don't have to copy secrets anywhere.

For sourcing operator images, things are more difficult because we would need some way to:
- know what secrets to use and where to inject them
- actually inject the secret into the pod spec(s) that need them as the BI is being installed. (this sounds a lot like what the cert-manager injection annotations do)

Perhaps we should separate these problems into separate issues?

exdx:
0.3 candidate

exdx:
Potentially moving to 0.4

nsapse:
Question - while this is a work in progress would it make sense from a UX perspective to have an error letting users know the reason their deployment is failing is the private source?

timflannagan:
I'm going to move this out of the 0.4 milestone, as that milestone already has a good chunk of tickets already assigned. This seems like a solid 0.5 milestone candidate.

akihikokuroda:
> Perhaps we should separate these problems into separate issues?

I agree.  The first issue is just for the image bundle and the second one applies for both github and image bundles.  I'll create a separate issue for the second one.

PR #279, opened 2022-04-27
add image pull secret suport for image bundle

akihikokuroda:

An optional `ImagePullSecretName` is added in the `ImageSource`.

Closes: #81

### DIFF
--- a/api/v1alpha1/bundle_types.go
+++ b/api/v1alpha1/bundle_types.go
@@ -64,6 +64,8 @@ type BundleSource struct {
 type ImageSource struct {
 	// Ref contains the reference to a container image containing Bundle contents.
 	Ref string `json:"ref"`
+	// ImagePullSecretName contains the name of the image pull secret in the namespace that the provisioner is deployed.
+	ImagePullSecretName string `json:"pullSecret,omitempty"`
 }
 
 type GitSource struct {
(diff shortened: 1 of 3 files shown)


akihikokuroda:
Are there any private image registries for e2e testing?

joelanford:
This looks good to me. We don't actually use an image registry in any of the tests yet. We just build the bundle images and load them into the `kind` cluster running the tests. I think this feature calls for some actual tests involving a registry that the cluster can pull from, but that seems like it can be captured as follow-on work.

Short-term, I suppose we could put some images in a "private" quay repo (private in quotes because I assume we'd put the credentials in plain view in the test). Longer-term, it'd be good to run an image registry as part of the e2e.

akihikokuroda:
> would an imagepullsecret in the provisioner namespace still make the most sense as the API to use?

I believe so.  The provisioner pod needs to access to the secret (by mounting or getting) to use it with the in-process client.  It makes sense to have it in the same namespace.

akihikokuroda:
> Short-term, I suppose we could put some images in a "private" quay repo (private in quotes because I assume we'd put the credentials in plain view in the test).

I could test this change using my personal private quay.io registry. 

I'll create a separate issue for creating e2e test with real registries.

exdx (approved):

tylerslaton (approved):

PR #286, opened 2022-04-29
bundle webhook: enforce spec immutability

joelanford:

### DIFF
--- a/api/v1alpha1/bundle_webhook.go
+++ b/api/v1alpha1/bundle_webhook.go
@@ -17,8 +17,10 @@ limitations under the License.
 package v1alpha1
 
 import (
+	"errors"
 	"fmt"
 
+	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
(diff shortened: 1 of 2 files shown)


joelanford:
With just this change, it's possible to cause the Bundle controller to permanently fail for a particular Bundle by deleting the underlying unpack pod. When the pod is deleted, the Bundle controller will notice, create a new unpack pod, and then attempt to set the `Bundle` phase back to `Pending`. This will fail because the `Bundle` is in the `Unpacked` phase.

timflannagan (approved):
Implementation looks good to me. I think we're covered on the following bases too:

- The API server would reject any updates to central metadata fields, like metadata.Name that would mess with higher level controllers
- We're fine with users/controllers updating any labels/annotations
- We're fine with users/controllers updating the status sub-resource

timflannagan:
> We're fine with users/controllers updating any labels/annotations

Actually I wonder whether this is behavior we'd allow if we're moving towards an embedded Bundle model?

joelanford:
Alright I updated this PR to only enforce spec immutability. That means there is still a possibility of a BI pivoting without a change to the Bundle. That could be accomplished by:
1. Using an image tag, a git branch, or a git tag in the Bundle source
2. Moving the image tag to a new digest, pushing changes to a git branch, or deleting and re-pushing a git tag on a different commit
3. Doing something to cause the unpack pod to be re-created (e.g. deleting the unpack pod)

If a user performs these steps, the new content from step 2 will be unpacked as a result of step 3, and the BI will notice, causing its pivot logic to execute.

IMO, this isn't a whole lot different than the pod example where one of the pod's container images uses and tag, the tag is moved to a different digest, and then at some unknown point in the future the existing pod gets rescheduled on a different node, and that node pulls the new image at the new digest and runs something different without you explicitly asking for it.

joelanford:
> Actually I wonder whether this is behavior we'd allow if we're moving towards an embedded Bundle model?

Something to think about here, but in general I don't think those should be immutable. The BI controller may want to reconcile the labels/annotations for the Bundles _it_ controls though. But I think that's separate from general immutability.

PR #287, opened 2022-04-29
add private git source support

### LINKED ISSUE #285: Support private git source bundle

akihikokuroda:

Add `Secret` at the top-level in the Bundle.Spec and removed `ImagePullSecretName` from `ImageSource`
```
        // Secret contains the secret name that has authorization information for the source and is	in the namespace that the provisioner is deployed.
	// for image source, the secret contains the image pull secret.
	// for git source, the secret contains the user ID (key: user) and the personal access token to the repository (key: token)
	Secret string `json:"secret,omitempty"`
```

Closes #285

### DIFF
--- a/api/v1alpha1/bundle_types.go
+++ b/api/v1alpha1/bundle_types.go
@@ -81,6 +81,11 @@ type GitSource struct {
 	// is required. Setting more than one field or zero fields will result in an
 	// error.
 	Ref GitRef `json:"ref"`
+	// SecretName contains the secret name that has authorization information for HTTPS protocol and is in the namespace that the provisioner is deployed.
+	// The secret is expected to contain `data.username` and `data.accesstoken` for the username and personal access token, respectively.
+	SecretName string `json:"secretName,omitempty"`
+	// If SslNoVerify is set true, the server certificate is ignored.
+	SslNoVerify bool `json:"sslNoVerify,omitempty"`
 }
 
 type GitRef struct {
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -3,6 +3,7 @@ package git
 import (
 	"errors"
 	"fmt"
+	"net/url"
 
 	rukpakv1alpha1 "github.com/operator-framework/rukpak/api/v1alpha1"
 )
@@ -10,6 +11,7 @@ import (
 const (
 	defaultDirectory = "./"
 	repositoryName   = "repo"
+	sslNoVerify      = "GIT_SSL_NO_VERIFY=true "
 )
 
 type checkoutCmd struct {
@@ -22,36 +24,49 @@ func CloneCommandFor(s rukpakv1alpha1.GitSource) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	return cmd.String(), nil
+	return cmd.String()
 }
 
-func (c *checkoutCmd) String() string {
+func (c *checkoutCmd) String() (string, error) {
 	var checkoutCommand string
 	var repository = c.Repository
 	var directory = c.Directory
 	var branch = c.Ref.Branch
 	var commit = c.Ref.Commit
 	var tag = c.Ref.Tag
+	var preset string
+
+	if c.SecretName != "" {
+		repositoryURL, err := url.Parse(repository)
+		if err != nil {
+			return "", fmt.Errorf("authorization is supported for https: %w", err)
+		}
+		repositoryURL.User = url.UserPassword("$USER", "$TOKEN")
+		repository = repositoryURL.String()
+	}
+	if c.SslNoVerify {
+		preset = sslNoVerify
+	}
 
 	if directory == "" {
 		directory = defaultDirectory
 	}
 
 	if commit != "" {
-		checkoutCommand = fmt.Sprintf("git clone %s %s && cd %s && git checkout %s && rm -r .git && cp -r %s/. /bundle",
-			repository, repositoryName, repositoryName, commit, directory)
-		return checkoutCommand
+		checkoutCommand = fmt.Sprintf("%sgit clone %s %s && cd %s && %sgit checkout %s && rm -r .git && cp -r %s/. /bundle",
+			preset, repository, repositoryName, repositoryName, preset, commit, directory)
+		return checkoutCommand, nil
 	}
 
 	if tag != "" {
-		checkoutCommand = fmt.Sprintf("git clone --depth 1 --branch %s %s %s && cd %s && git checkout tags/%s && rm -r .git && cp -r %s/. /bundle",
-			tag, repository, repositoryName, repositoryName, tag, directory)
-		return checkoutCommand
+		checkoutCommand = fmt.Sprintf("%sgit clone --depth 1 --branch %s %s %s && cd %s && %sgit checkout tags/%s && rm -r .git && cp -r %s/. /bundle",
+			preset, tag, repository, repositoryName, repositoryName, preset, tag, directory)
+		return checkoutCommand, nil
 	}
 
-	checkoutCommand = fmt.Sprintf("git clone --depth 1 --branch %s %s %s && cd %s && git checkout %s && rm -r .git && cp -r %s/. /bundle",
-		branch, repository, repositoryName, repositoryName, branch, directory)
-	return checkoutCommand
+	checkoutCommand = fmt.Sprintf("%sgit clone --depth 1 --branch %s %s %s && cd %s && %sgit checkout %s && rm -r .git && cp -r %s/. /bundle",
+		preset, branch, repository, repositoryName, repositoryName, preset, branch, directory)
+	return checkoutCommand, nil
 }
 
 func (c *checkoutCmd) Validate() error {
--- a/internal/provisioner/plain/controllers/bundle_controller.go
+++ b/internal/provisioner/plain/controllers/bundle_controller.go
@@ -484,6 +484,12 @@ func bundleGitRepoPod(pod *corev1.Pod, source rukpakv1alpha1.GitSource, unpackIm
 	}
 	pod.Spec.InitContainers[1].Command = []string{"/bin/sh", "-c", cmd}
 	pod.Spec.InitContainers[1].VolumeMounts = []corev1.VolumeMount{{Name: "bundle", MountPath: "/bundle"}}
+	if source.SecretName != "" {
+		pod.Spec.InitContainers[1].Env = []corev1.EnvVar{
+			{Name: "USER", ValueFrom: &corev1.EnvVarSource{SecretKeyRef: &corev1.SecretKeySelector{LocalObjectReference: corev1.LocalObjectReference{Name: source.SecretName}, Key: "username"}}},
+			{Name: "TOKEN", ValueFrom: &corev1.EnvVarSource{SecretKeyRef: &corev1.SecretKeySelector{LocalObjectReference: corev1.LocalObjectReference{Name: source.SecretName}, Key: "accesstoken"}}},
+		}
+	}
 
 	if len(pod.Spec.Containers) != 1 {
 		pod.Spec.Containers = make([]corev1.Container, 1)
--- a/internal/git/git_test.go
+++ b/internal/git/git_test.go
@@ -86,6 +86,31 @@ func TestCheckoutCommand(t *testing.T) {
 			expected: "",
 			err:      errors.New("cannot specify both branch and commit: only one is allowed"),
 		},
+		{
+			source: rukpakv1alpha1.GitSource{
+				Repository: "https://github.com/operator-framework/combo",
+				Ref: rukpakv1alpha1.GitRef{
+					Commit: "4567031e158b42263e70a7c63e29f8981a4a6135",
+				},
+				SecretName: "secretname",
+			},
+			expected: fmt.Sprintf("git clone %s %s && cd %s && git checkout %s && rm -r .git && cp -r %s/. /bundle",
+				"https://$USER:$TOKEN@github.com/operator-framework/combo", repositoryName, repositoryName, "4567031e158b42263e70a7c63e29f8981a4a6135",
+				"./"),
+		},
+		{
+			source: rukpakv1alpha1.GitSource{
+				Repository: "https://github.com/operator-framework/combo",
+				Ref: rukpakv1alpha1.GitRef{
+					Commit: "4567031e158b42263e70a7c63e29f8981a4a6135",
+				},
+				SecretName:  "secretname",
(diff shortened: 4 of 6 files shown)


internal/git/git.go
```diff
@@ -14,44 +15,55 @@ const (
...
 	var commit = c.Ref.Commit
 	var tag = c.Ref.Tag
 
+	if c.User != "" || c.Token != "" {
+		repositoryURL, err := url.Parse(repository)
+		if err != nil {
+			return "", err
+		}
+		repositoryURL.User = url.UserPassword(c.User, c.Token)
+		repository = repositoryURL.String()
```
joelanford, line 45: This will result in the plaintext credentials being in the pod spec of the git unpack pod, which IMO is a non-starter.

Is there a way to mount the credentials as a volume or set environment variables with `valueFrom` and have git read from there?

internal/git/git.go
```diff
@@ -14,44 +15,55 @@ const (
...
 	var directory = c.Directory
 	var branch = c.Ref.Branch
 	var commit = c.Ref.Commit
 	var tag = c.Ref.Tag
 
+	if c.User != "" || c.Token != "" {
+		repositoryURL, err := url.Parse(repository)
+		if err != nil {
+			return "", err
+		}
```
joelanford, line 43: Git supports several [protocols](https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols), (not just http) so it seems like we need to support them or at least have our API make it obvious which of them we support?

I wonder if it would be easier to refactor the git source using a git library before we try to tackle git auth stuff?

internal/git/git.go
```diff
@@ -14,44 +15,55 @@ const (
 
 type checkoutCmd struct {
 	rukpakv1alpha1.GitSource
+	User  string
+	Token string
```
joelanford, line 19: We probably need a way for a user to provide a CA cert in case the git remote isn't trusted by the root certificate store of the unpack pod.

Does git permit authorization via client certs? If so, we might need to think about handling that?

internal/git/git.go
exdx: I kind of agree, the original implementation got us far enough but we probably need to overhaul it to be more robust. I wonder if it makes more sense to have #191 land first so the git codepath can be a lot cleaner. I can open a tracking issue for this and we can see how we prioritize it.

internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -455,6 +456,14 @@ func bundleGitRepoPod(pod *corev1.Pod, source rukpakv1alpha1.GitSource, unpackIm
 	return pod, nil
 }
 
+func (r *BundleReconciler) getCredential(ctx context.Context, secretname string) (string, string, error) {
+	secret, err := r.KubeClient.CoreV1().Secrets(r.PodNamespace).Get(ctx, secretname, metav1.GetOptions{})
```
exdx, line 460: would the namespace be the one the Bundle is in versus the one the BundleReconciler (plain provisioner) is in?

internal/git/git.go
timflannagan: What's our story around rate limiting when using git clients under-the-hood then? Instructing users to configure a Secret reference so we can have proper authentication? That seems potentially reasonable, but I'm not sure if that's surprising UX behavior, or whether this introduces a more leaky abstraction.

api/v1alpha1/bundle_types.go
```diff
@@ -50,6 +50,10 @@ type BundleSpec struct {
 	ProvisionerClassName string `json:"provisionerClassName"`
 	// Source defines the configuration for the underlying Bundle content.
 	Source BundleSource `json:"source"`
+	// Secret contains the secret name that has authorization information for the source and is in the namespace that the provisioner is deployed.
+	// for image source, the secret contains the image pull secret.
+	// for git source, the secret contains the user ID (key: user) and the personal access token to the repository (key: token)
```
exdx, line 55: I wonder if we want to support auth via ssh as well? I think we can take inspiration from source-controller's implementation https://github.com/fluxcd/source-controller/blob/065a7607523dde326c3d513461befb047bd77eff/api/v1beta2/gitrepository_types.go#L62

internal/provisioner/plain/controllers/bundle_controller.go
timflannagan, line 460: Should we use c-r dynamic client here? The only reason we have that kubernetes.Interface client present is because c-r doesn't have support for reading a Pod's log sub-resource.

internal/git/git.go
akihikokuroda: >Does git permit authorization via client certs? If so, we might need to think about handling that?

Yes, it does.  But I think that the personal access token is the most popular way for authorization.  Can the client certs support be a separate issue?  

>We probably need a way for a user to provide a CA cert in case the git remote isn't trusted by the root certificate store of the unpack pod.

We probably start with the option to skip the server certificate.

internal/git/git.go
akihikokuroda: Yes, I'll try the environment variables.

internal/git/git.go
exdx: See #291 for tracking issue

internal/provisioner/plain/controllers/bundle_controller.go
akihikokuroda: It is the namespace where the plain provisioner running.  Bundle is cluster scope, isn't it?

api/v1alpha1/bundle_types.go
akihikokuroda: Yes, we do.  Can it be a separate issue?

internal/provisioner/plain/controllers/bundle_controller.go
akihikokuroda: I see.  I probably remove this function.  I'll mount the secret instead of reading it.  Thanks!

exdx: yup, sorry blanked on that haha, they are cluster scoped. 

so we would expect a user to create a secret in the `rukpak-system` namespace for example for the provisioner to be able to source a bundle that references some particular private git content. I suppose that UX makes sense.

timflannagan: I'm cool with requiring that Secrets must live in the rukpak system namespace in order to use this functionality. I'm mainly interested in avoiding an implementation where we copy Secrets between namespaces.

internal/git/git.go
akihikokuroda: I would like to implement the authentication with the current design for now.  I'll work on the #291 after that.

internal/git/git.go
akihikokuroda: Issue for the client certificate authorization: #294

akihikokuroda: > We probably start with the option to skip the server certificate.
I added `SslNoVerify` option in the GitSource in the last commit.

internal/git/git.go
joelanford: I'm still a little concerned about the impact of this change on SSH remotes. To me, it seems like this will make the API more brittle (i.e. "you can use a secret with http remotes, but if you try to use it with SSH remotes, things will break in potentially weird ways")

akihikokuroda: I see.  The `url.Parse` should be used only for https protocol.  This PR only works with the https protocol.  May I make the SSH remote case in the follow up issue #294?

akihikokuroda: I'll update to return the `repository` as-is when the url.Parse so that the error is the same one without the secret provided.

joelanford: Thinking about this more, I suppose that ssh remotes have never worked because we don't have anything that supports providing SSH credentials, which are required to even make the SSH connection. So I suppose we're okay on that front.

Should we perhaps improve docs/validation to make it clear that only SSH is supported?

akihikokuroda: I updated the description of the `Secret.Secret`.  Thanks!

internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -436,6 +436,12 @@ func bundleGitRepoPod(pod *corev1.Pod, source rukpakv1alpha1.GitSource, unpackIm
 	}
 	pod.Spec.InitContainers[1].Command = []string{"/bin/sh", "-c", cmd}
 	pod.Spec.InitContainers[1].VolumeMounts = []corev1.VolumeMount{{Name: "bundle", MountPath: "/bundle"}}
+	if source.Secret != "" {
+		pod.Spec.InitContainers[1].Env = []corev1.EnvVar{
+			{Name: "USER", ValueFrom: &corev1.EnvVarSource{SecretKeyRef: &corev1.SecretKeySelector{LocalObjectReference: corev1.LocalObjectReference{Name: source.Secret}, Key: "user"}}},
+			{Name: "TOKEN", ValueFrom: &corev1.EnvVarSource{SecretKeyRef: &corev1.SecretKeySelector{LocalObjectReference: corev1.LocalObjectReference{Name: source.Secret}, Key: "token"}}},
```
joelanford, line 442: Does git use "token" in its terminology for auth? I wonder if we should call this "password" instead?

akihikokuroda: It is called `Personal Access Token`.  It can be generated in Settings->Developer settings->Personal access tokens.

api/v1alpha1/bundle_types.go
```diff
@@ -80,6 +80,11 @@ type GitSource struct {
 	// is required. Setting more than one field or zero fields will result in an
 	// error.
 	Ref GitRef `json:"ref"`
+	// Secret contains the secret name that has authorization information for HTTPS protocol and is in the namespace that the provisioner is deployed.
+	// The secret contains the user ID (key: user) and the personal access token to the repository (key: token)
```
joelanford, line 84: Maybe make this slightly more concise and direct?
```suggestion
	// Secret contains the secret name that has authorization information for HTTPS protocol and is in the namespace that the provisioner is deployed.
	// The secret is expected to contain `data.username` and `data.password` for the username and password, respectively.
```

api/v1alpha1/bundle_types.go
```diff
@@ -80,6 +80,11 @@ type GitSource struct {
...
+	Secret string `json:"secret,omitempty"`
```
joelanford, line 85: Rename to `SecretName`?

I noticed that the `pod.spec.serviceAccount` field is deprecated in favor of `pod.spec.serviceAccountName`. Seems like the reasoning for that applies here as well.

akihikokuroda: Yes, I'll update.  Thanks!

api/v1alpha1/bundle_types.go
akihikokuroda: Thanks!  I'll update `token` and `personal access token`.  It's a good practice to use the personal access token instead of the password.

internal/git/git.go
```diff
@@ -22,36 +24,48 @@ func CloneCommandFor(s rukpakv1alpha1.GitSource) (string, error) {
...
 
-func (c *checkoutCmd) String() string {
+func (c *checkoutCmd) String() (string, error) {
 	var checkoutCommand string
 	var repository = c.Repository
 	var directory = c.Directory
 	var branch = c.Ref.Branch
 	var commit = c.Ref.Commit
 	var tag = c.Ref.Tag
+	var preset = ""
```
exdx, line 37:
```suggestion
	var preset string
```

internal/git/git.go
```diff
@@ -22,36 +24,48 @@ func CloneCommandFor(s rukpakv1alpha1.GitSource) (string, error) {
...
+
+	if c.Secret != "" {
+		repositoryURL, err := url.Parse(repository)
+		if err == nil {
+			repositoryURL.User = url.UserPassword("$USER", "$TOKEN")
+			repository = repositoryURL.String()
+		}
+	}
+	if c.SslNoVerify {
+		preset = sslNoVerify
```
exdx, line 47: nit: if this variable is being assigned only here once, maybe we can remove it entirely and inject `sslNoVerify` into the checkoutCommand directly?

exdx (commented):
Looks good, ready to approve after the latest round of feedback 

Only thing I think missing is some documentation around private git bundles being added to the existing `docs/git-bundles` doc. That can be done as follow-up though. Also maybe for the image based bundles as well.

internal/git/git.go
akihikokuroda: Thanks!  I'll update.

internal/git/git.go
akihikokuroda: Does the above comment make sense?  Thanks!

akihikokuroda:
I created the doc update issue #324.

internal/git/git.go
exdx: I see, that makes sense, you only want to conditionally set it 👍

internal/git/git.go
exdx: That makes sense -- Go assigns a default zero value to all variables when they're declared (and not initialized to something) so something like
```
var preset string
```
results in `preset` equaling the default string `""` value automatically. Declaring it this way here is good because it's more clear IMO. Checkout https://go.dev/tour/basics/12 for a demo. Using the default zero types is generally very helpful and makes code clearer.

exdx (approved):

internal/git/git.go
timflannagan: Yeah +1 to using the default value here as well. We can initialize the `preset` variable to be an empty string by default, and if the c.SSLNoVerify option is true, override the default empty string value with the sslNoVerify value.

internal/git/git.go
```diff
@@ -22,36 +24,48 @@ func CloneCommandFor(s rukpakv1alpha1.GitSource) (string, error) {
...
 	var repository = c.Repository
 	var directory = c.Directory
 	var branch = c.Ref.Branch
 	var commit = c.Ref.Commit
 	var tag = c.Ref.Tag
+	var preset = ""
+
+	if c.SecretName != "" {
+		repositoryURL, err := url.Parse(repository)
+		if err == nil {
```
timflannagan, line 41: What happens if we encounter a parsing error here? Does it make sense to silently ignore it?

akihikokuroda:
@timflannagan Is it OK to merge this?  Are there more review necessary?

akihikokuroda:
The git source refactor #366 invalidated this PR.  #407 is the new implementation after the refactor.

PR #292, opened 2022-04-29
Use emptyDir to store bundle content rather than configmaps

joelanford:

Closes https://github.com/operator-framework/rukpak/issues/296


internal/storage/localdir.go
```diff
@@ -0,0 +1,117 @@
...
+		h.Uid = 0
+		h.Gid = 0
+		h.Uname = ""
+		h.Gname = ""
+		h.Name = path
+
+		if err := tw.WriteHeader(h); err != nil {
+			return fmt.Errorf("write tar header for %q: %w", path, err)
+		}
+		if d.IsDir() {
```
akihikokuroda, line 73: I believe that this never happen because the `getObjects` fails if the manifest has a directory.  So it doesn't matter but I'm wondering if the header should be written with out the writing the data for the directory.

joelanford: This is meant to be generic and usable by another provisioner type that uses a different bundle format.

This code is effectively "just tar.gz the bundle and save it." It's up to the provisioner to read the `fs.FS` and decide what's valid or not.

In the case of the plain provisioner, you're right. We use `getObjects` in the bundle controller just to check that what we're about to save is actually a valid plain provisioner. Other provisioners should probably follow that same pattern.

internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -94,6 +97,29 @@ func (r *BundleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ c
 	}()
 	u.UpdateStatus(updater.EnsureObservedGeneration(bundle.Generation))
 
+	res, err := r.Finalizers.Finalize(ctx, bundle)
+	if err != nil {
+		return ctrl.Result{}, err
```
timflannagan, line 102: Do we need to surface this error on the Bundle's status or is persisting that error state a leaky abstraction?

timflannagan (approved):
This looks reasonable to me at a high level: did you put any thought into an implementation that avoids the usage of finalizers, and instead has logic that invalidates a store key when we intercept the deletion timestamp, or when we attempt to get an individual Bundle's contents from the store and run into an error during that process?

Are there any obvious follow-ups here outside of adding support for configuring this volume mount?

internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -33,12 +33,14 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
+	apimacherrors "k8s.io/apimachinery/pkg/util/errors"
```
timflannagan, line 36: Interesting: I thought our goimportas linter configuration would've had a package import enforcement for `utilerrors` but I guess I never added that. I'm fine with the import name as long as we're consistent throughout the codebase, and it doesn't look like we use this package anywhere else right now.

internal/provisioner/plain/controllers/bundle_controller.go
joelanford: Oh good catch. Yeah I think we need to surface this error. My only hesitation is that I have in my head that you can't update an object in certain ways once it has a deletion timestamp. However, I'm guessing that status updates must be allowed. Otherwise how would you ever know why a finalizer was failing to be removed?

joelanford:
> invalidates a store key when we intercept the deletion timestamp

I'd agree that a finalizer-free implementation would be nicer.

I figure there's a decent possibility we'd miss deletion events and we'd slowly accrue zombie files. I suppose we could solve that by running a goroutine that periodically re-lists from the cache and deletes bundle files for anything not found in the cache.

Let me give that a quick spin.

timflannagan:
> I figure there's a decent possibility we'd miss deletion events and we'd slowly accrue zombie files. I suppose we could solve that by running a goroutine that periodically re-lists from the cache and deletes bundle files for anything not found in the cache.

Something else we may need to consider is a partial cache write where we encounter an error while storing objects, but I think this is only a problem with the bundle controller right now, and it's not entirely clear to me how you'd be able to do this without introducing extra complexity (maybe through channels/signals/etc.?)

internal/storage/localdir_test.go
```diff
@@ -0,0 +1,148 @@
...
+				ModTime: info.ModTime(),
+				Sys:     info.Sys(),
+			}
+			return nil
+		}
+	}
+	if err := fs.WalkDir(a, ".", walkFunc(a, aMap)); err != nil {
+		return false, err
+	}
+	if err := fs.WalkDir(a, ".", walkFunc(b, bMap)); err != nil {
```
akihikokuroda, line 144: Is `fs.WalkDir(a,` correct?

internal/storage/storage.go
```diff
@@ -1,312 +1,14 @@
...
-
-var _ Storage = &ConfigMaps{}
-
-type ConfigMaps struct {
-	Client     client.Client
-	Namespace  string
-	NamePrefix string
+	Load(ctx context.Context, owner client.Object) (fs.FS, error)
+	Store(ctx context.Context, owner client.Object, bundle fs.FS) error
+	Delete(ctx context.Context, owner client.Object) error
```
akihikokuroda, line 13: This looks good.  Can we support multiple kind of storage and make them selectable?

timflannagan: How would a provisioner expose configuring storage? I assume it would be through the ProvisionerClass API, but we'd need to be able to query for that resource in a main.go before the manager instance has started.

internal/storage/localdir_test.go
joelanford: Nope. Copy/paste error. Thanks for catching that!

internal/provisioner/plain/controllers/bundle_controller.go
timflannagan: > In short, what’s happened is that the object was updated, not deleted. That’s because Kubernetes saw that the object contained finalizers and put it into a read-only state. The deletion timestamp signals that the object can only be read, with the exception of removing the finalizer key updates. In other words, the deletion will not be complete until we edit the object and remove the finalizer.

I'm not entirely reading through [Using Finalizers to Control Deletion](https://kubernetes.io/blog/2021/05/14/using-finalizers-to-control-deletion/) whether updating a sub-resource once the finalizer is in place would be a valid workflow? Maybe this is something we can spec out locally using a simple main.go?

internal/storage/localdir.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+	defer bundleFile.Close()
+	tarReader, err := gzip.NewReader(bundleFile)
+	if err != nil {
+		return nil, err
+	}
+	return tarfs.New(tarReader)
+}
+
+func (s *LocalDirectory) Store(_ context.Context, owner client.Object, bundle fs.FS) error {
```
exdx, line 43: should we check to see if the bundle is already stored before attempting to store it again? Or is it because the bundle path is `fmt.Sprintf("%s.tgz", owner.GetUID())` we are always going to load the bundle, regardless if another copy of it is already on-disk? 

I suppose this could be a premature optimization that we could address later.

exdx (approved):
This looks good. I like the finalizer implementation. 

I think we should probably update the docs too to reference the fact that the default storage mechanism is now a PV mounted in the provisioner deployment, versus a series of configmaps that users of OLM might be expecting.

joelanford:
> I suppose we could solve that by running a goroutine that periodically re-lists from the cache and deletes bundle files for anything not found in the cache.

I looked into this a bit and realized that we'd have to expand the `storage.Storage` interface to include a `List` or `Reap` or `Prune` method. And that implies that a storage implementation would have to be able to list all of the files. For a mounted filesystem implementation, that's fairly cheap, but for something like S3, that would require at least 1 extra API call (maybe more depending on pagination requirements) per whatever the polling interval is.

I'm tempted to leave the finalizer in place for now to keep the storage interface minimal, and then revisit this question separately. Thoughts on that?

internal/storage/localdir.go
joelanford: It's actually important that `os.Create` works the way it does (if the file exists, it truncates the file and allows re-writing it).

That way, our reconcile logic is still always just reading the pod logs and writing (or re-writing) what we read from the pod. If the pod logs change, we overwrite what we had stored before.

That gets us back into the territory of what's immutable and what's not, but the behavior in this PR at least matches the existing behavior, which is that if the pod logs change, then the stored bundle changes with it.

joelanford:
> Does it make sense to track whether we need to re-evaluate our storage interface down the line?

Yeah I'll make some new tracking issues that cover:
1. Are we okay with using finalizers; what are the alternatives and pros/cons?
2. Are we okay with the storage interface? Should we keep it minimal? If it expands what are the implications?

internal/provisioner/plain/finalizer/deletecachedbundle.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"context"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/finalizer"
+
+	"github.com/operator-framework/rukpak/internal/storage"
+)
+
+var _ finalizer.Finalizer = &DeleteCachedBundle{}
```
tylerslaton, line 12: `non-blocking question`: What exactly is this line doing? I've seen this pattern before but don't quite understand it.

internal/provisioner/plain/finalizer/deletecachedbundle.go
```diff
@@ -0,0 +1,25 @@
...
+
+type DeleteCachedBundle struct {
+	storage.Storage
+}
+
+func (f DeleteCachedBundle) Finalize(ctx context.Context, obj client.Object) (finalizer.Result, error) {
+	if err := f.Delete(ctx, obj); err != nil {
+		return finalizer.Result{}, err
+	}
+	return finalizer.Result{}, nil
```
tylerslaton, line 24: This logic is essentially equivalent to this:
```suggestion
	return finalizer.Result{}, f.Delete(ctx, obj)
```

tylerslaton (approved):
Couple of questions about the code but nothing blocking!

internal/provisioner/plain/finalizer/deletecachedbundle.go
joelanford: It's a compile-time check that ensures `DeleteCachedBundle` conforms to the `finalizer.Finalizer` interface.

akihikokuroda (approved):
LGTM.  Please fix the one copy/paste error in the unit test.

internal/provisioner/plain/finalizer/deletecachedbundle.go
tylerslaton: Understood, thank you for the answer.

internal/provisioner/plain/controllers/bundle_controller.go
joelanford: I just did a manual test using the kubectl edit-status plugin. Status updates are allowed.

I'll update this to set a condition relating to this error.

internal/provisioner/plain/README.md
```diff
@@ -85,7 +85,8 @@ my-bundle-instance   my-bundle        my-bundle          InstallationSucceeded
 There is a natural separation between sourcing of the content and application of that content via two separate RukPak
 APIs, `Bundle` and `BundleInstance`. A user can specify a particular `Bundle` to be available in the cluster for
 inspection before any application of the resources. Given a `Bundle` resource named `my-bundle`, the plain provisioner
-will pull down and unpack the bundle to a set of ConfigMaps.
+will pull down and unpack the bundle to a tar.gz file that is saved into a bundle cache directory mounted in the
```
timflannagan, line 88: Any idea whether we should expose/document that directory location, or do we provide enough discoverability through our set of kustomize manifests?

timflannagan (approved):
LGTM - I only had nits that didn't feel worth surfacing as comments.

internal/provisioner/plain/README.md
joelanford: I thought about that. This implementation _does_ make the bundle content a bit less observable. You can no longer `kubectl get configmap -o jsonpath={.data.bundle} | base64 -d | tar zf` or whatever to see individual manifests.

Now its something closer to `kubectl exec <provisionerPod> -- cat /var/cache/bundles/<bundleUID>.tgz | tar tzv`

I think we should handwave over that for now and come up with a better interface.

My initial thought is to define an http api for retrieving bundles and then adding a `ServeHTTP` method to the storage interface such that a provisioner can attach the storage implementation to an http handler, and then put the address of the bundle contents in the bundle status. I'll write another follow-up issue for this idea.

joelanford: > It might be time to also start re-evaluating the objects array on the Bundle's status?

It's like you're in my head. Already working that idea. I should have a PR up today (I think) that removes the entire BundleInfo struct.

PR #293, opened 2022-04-29
*: Add support for dynamically generating Bundle resources

### LINKED ISSUE #73: Add support for embedding Bundles in the BundleInstance spec
Implement support in the BundleInstance API and the plain manifest bundle provisioner for embedding a Bundle spec as outlined in the [e2e strawman](https://hackmd.io/hppR60SiRGKPcqKQuByvVg?view#Instance-API).

timflannagan:
# Summary

Introduce support for embedding Bundles within the BundleInstance API, and dynamically generating Bundle resources within that API.

## Prior work

- [Design document](https://hackmd.io/HNWfNUbqTUGAW5VRMcc40w?both)
- [Immutable Bundles](#286)

## API Changes

The BundleInstance (BI) API has been updated to add support for configuring Bundle resources through a BundleTemplate `spec.template` field, which replaces the existing `spec.BundleName` require field. This API now closely resembles patterns seen throughout core Kubernetes, like the Deployment/ReplicaSet/StatefulSet/etc. APIs that contain an embedded PodSpec resource.

## Behavior Changes

The plain provisioner now supports dynamically generated Bundle resources based on the desired `spec.Template` template configuration. That template configuration field is a required field in the BI CRD. Any Bundle that was generated by the plain provisioner's BI controller will contain the `core.rukpak.io/owner-*` labels, and an owner reference marking the individual BI resource as the parent resource. When Bundles are manually deleted (e.g. through a user/controller), the BI controller will re-create those Bundles to match the desired specification outlined in the `spec.Template` Bundle template field.

It's important to note that existing behavior around statically "activating' Bundle resources through the `spec.BundleName` field for an individual BundleInstance resource is no longer a possible workflow. In the future, both the static and dynamic "activation' of Bundle resources within an individual BundleInstance resource may be possible as it's a pattern seen throughout core Kubernetes (e.g. PersistentVolumes can be [statically or dynamically provisioned](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#provisioning).

> Note: You can still manually create Bundle resources to help enable auditing workflows.

When the plain provisioner's BI controller sees a BundleInstance resource that contains a desired Bundle template that doesn't match any Bundles that exist on the cluster, a new Bundle will be generated to match that desired configuration. Due  [immutability of a Bundle's spec](#286) field, any underlying Bundle that belong to an individual BI resource will need to be re-created vs. updated to match that desired specification. The plain provisioner only respects the metadata.Labels and metadata.Annotations field when comparing the existing and desired state of a Bundle resource.

In the case that multiple Bundles match the desired configuration, the BI controller is responsible for selecting the most recent Bundle based on the `metadata.CreationTimestamp` field, and ensuring that selected Bundle's unpacked contents are present on the cluster.

During updates to an individual BundleInstance resource, the plain provisioner is responsible for checking whether the newly generated Bundle resource is able to be successfully unpacked and installed, before removing the previously installed Bundle's contents from the cluster. This behavior helps prevent the scenario where newly generated Bundle resources fail to unpack, and the existing installation is deleted due to poor pivoting behavior.

## Misc.

Closes #73


api/v1alpha1/bundleinstance_types.go
```diff
@@ -46,25 +47,37 @@ const (
...
+}
 
-	// BundleName is the name of the bundle that this instance is managing on the cluster.
-	BundleName string `json:"bundleName"`
+// BundleTemplate defines the desired state of a Bundle resource
+type BundleTemplate struct {
+	// Standard object's metadata.
+	// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata
+	// +optional
+	// +kubebuilder:pruning:PreserveUnknownFields
```
timflannagan, line 61: TODO: move this PreserveUnknownFields kubebuilder marker to the parent BundleTemplate field.

api/v1alpha1/bundleinstance_types.go
```diff
@@ -46,25 +47,37 @@ const (
...
+	InstalledBundleRef *corev1.ObjectReference `json:"installedBundleRef,omitempty"`
+	DesiredBundleRef   *corev1.ObjectReference `json:"desiredBundleRef,omitempty"`
 }
 
 //+kubebuilder:object:root=true
 //+kubebuilder:resource:scope=Cluster
 //+kubebuilder:subresource:status
-//+kubebuilder:printcolumn:name="Desired Bundle",type=string,JSONPath=`.spec.bundleName`
-//+kubebuilder:printcolumn:name="Installed Bundle",type=string,JSONPath=`.status.installedBundleName`
+//+kubebuilder:printcolumn:name="Desired Bundle",type=string,JSONPath=`.status.desiredBundleRef.name`
```
timflannagan, line 79: TODO: fix the "Desired Bundle" printer column.

api/v1alpha1/bundleinstance_types.go
```diff
@@ -46,25 +47,37 @@ const (
...
 // BundleInstanceStatus defines the observed state of BundleInstance
 type BundleInstanceStatus struct {
-	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
-	// Important: Run "make" to regenerate code after modifying this file
-	Conditions []metav1.Condition `json:"conditions,omitempty"`
-
-	InstalledBundleName string `json:"installedBundleName,omitempty"`
+	Conditions         []metav1.Condition      `json:"conditions,omitempty"`
+	InstalledBundleRef *corev1.ObjectReference `json:"installedBundleRef,omitempty"`
+	DesiredBundleRef   *corev1.ObjectReference `json:"desiredBundleRef,omitempty"`
```
timflannagan, line 73: TODO: figure out whether this is the correct implementation.

internal/util/util.go
```diff
@@ -49,7 +49,7 @@ func MapBundleToBundleInstanceHandler(cl client.Client, log logr.Logger) handler
 		}
 		for _, bi := range bundleInstances.Items {
 			bi := bi
-			if bi.Spec.BundleName == b.Name {
+			if bi.Spec.Template.Name == b.Name {
```
timflannagan, line 52: TODO: this seems like the wrong comparison in the context of the current implementation. We'd need to either parse the current Bundle's owner references, or compare the Bundle's label selectors with the current BI's spec.Selector values.

internal/provisioner/plain/controllers/bundleinstance_controller.go
```diff
@@ -267,10 +262,106 @@ func (r *BundleInstanceReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+				Name:            bi.GetName() + "-" + rand.String(5),
+				OwnerReferences: []metav1.OwnerReference{*controllerRef},
+				Labels:          bi.Spec.Template.Labels,
+			},
+			Spec: bi.Spec.Template.Spec,
+		}
+		if err := r.Create(ctx, b); err != nil {
+			// TODO: handle this case better as an IsAlreadyExists error could be due to a slow cache?
+			if apierrors.IsAlreadyExists(err) {
+				return nil, nil, err
```
exdx, line 349: maybe we just return a nil error in this case?

exdx (commented):
I like the overall implementation, LGTM

api/v1alpha1/bundleinstance_types.go
timflannagan: Nope - this was more-or-less a stale artifact when I was throwing code against the wall last week. I'm still trying to work out in my head whether this kind of field is needed, or whether we still need to surface the desired bundle printer column in the immediate term.

README.md
```diff
@@ -122,8 +134,8 @@ the content of the bundle available in the cluster.
 ### CustomResourceDefinition (CRD) Validator
 
 RukPak comes with a webhook for validating the upgrade of CRDs from `Bundle`s. If a CRD does potentially destructive
-actions to the cluster, it will not allow it go be applied. In the context of RukPak, this will result in a failed 
-`BundleInstance` resolution. 
+actions to the cluster, it will not allow it go be applied. In the context of RukPak, this will result in a failed
```
tylerslaton, line 137: May as well fix it while we're here 🤷🏻 
```suggestion
actions to the cluster, it will not allow it to be applied. In the context of RukPak, this will result in a failed
```

README.md
```diff
@@ -105,8 +105,20 @@ kind: BundleInstance
 metadata:
   name: my-bundle-instance
 spec:
-  bundleName: my-bundle
   provisionerClassName: core.rukpak.io/plain
+  selector:
```
tylerslaton, line 109: Is it worth adding a block describing what exactly this selector is selecting? Essentially, I think we should add some documentation around what the embedded `Bundle` is and how a user uses it since it isn't initially intuitive.

api/v1alpha1/bundleinstance_types.go
```diff
@@ -29,41 +29,55 @@ const (
...
 type BundleInstanceSpec struct {
 	// ProvisionerClassName sets the name of the provisioner that should reconcile this BundleInstance.
 	ProvisionerClassName string `json:"provisionerClassName"`
+	// Label selector for Bundles resources.
+	Selector *metav1.LabelSelector `json:"selector"`
+	// Template describes the generated Bundle that this instance will manage.
+	// +kubebuilder:pruning:PreserveUnknownFields
+	// +kubebuilder:validation:Schemaless
+	Template *BundleTemplate `json:"template"`
+}
```
tylerslaton, line 58: Can we make these properties be a bit more explicit? By giving them `Bundle` prefixes it makes this a bit more obvious what these fields are used for. I'm not entirely devoted to this exact fix but maybe we can be clearer here. Plus, it matches the struct names.
```suggestion
	// Label selector for Bundles resources.
	BundleSelector *metav1.LabelSelector `json:"selector"`
	// BundleTemplate describes the generated Bundle that this instance will manage.
	// +kubebuilder:pruning:PreserveUnknownFields
	// +kubebuilder:validation:Schemaless
	BundleTemplate *BundleTemplate `json:"template"`
}
```

internal/provisioner/plain/README.md
```diff
@@ -125,52 +131,44 @@ Create the bundle:
...
 ```
 
 Check the BundleInstance status to ensure that the installation was successful.
 
 ```console
 $ kubectl get bundleinstance combo
-NAME    DESIRED BUNDLE   INSTALLED BUNDLE   INSTALL STATE           AGE
-combo   combo-v0.0.1     combo-v0.0.1       InstallationSucceeded   10s
+NAME    INSTALLED BUNDLE   INSTALL STATE           AGE
+combo   combo-9njsj        InstallationSucceeded   10s
```
tylerslaton, line 171: Should we add a printerColumn that indicates what ref this `BundleInstance` is referencing? It isn't immediately obvious what we're referencing with this `BundleInstance` without going to the created Bundle itself. Not important but I could see a situation like this:

```console
NAME    INSTALLED BUNDLE   INSTALL STATE           REF      AGE
combo   combo-9njsj        InstallationSucceeded   v0.0.1   10s
```

In essence it is very similar to https://github.com/operator-framework/rukpak/pull/276

internal/provisioner/plain/controllers/bundleinstance_controller.go
```diff
@@ -51,6 +51,17 @@ import (
 	"github.com/operator-framework/rukpak/internal/util"
 )
 
+const (
+	maxGeneratedBundleLimit = 4
```
tylerslaton, line 55: Should this be configurable with a default? How did we settle on 4?

internal/provisioner/plain/controllers/bundleinstance_controller.go
```diff
@@ -262,12 +283,84 @@ func (r *BundleInstanceReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		if b.GetName() == currBundle.GetName() {
+			continue
+		}
+		if err := r.Delete(ctx, b); err != nil {
+			errors = append(errors, err)
+			continue
+		}
+		deletedBundles++
+	}
+	return deletedBundles, utilerrors.NewAggregate(errors)
```
tylerslaton, line 361: Is there a point to return the `deletedBundles` count if we're not using it when we call this function above?

tylerslaton (commented):
Initial round of review for this PR. I like the implementation and it matches the [design](https://hackmd.io/HNWfNUbqTUGAW5VRMcc40w?both) well.

internal/provisioner/plain/controllers/bundleinstance_controller.go
timflannagan: This is just a short term bandaid to prevent hotlooping behavior before we start implementing revision history limit behavior, which the deployment/replicaset controllers use.

internal/provisioner/plain/controllers/bundleinstance_controller.go
```diff
@@ -96,36 +103,50 @@ func (r *BundleInstanceReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
-			meta.SetStatusCondition(&bi.Status.Conditions, metav1.Condition{
-				Type:   rukpakv1alpha1.TypeInstalled,
-				Status: metav1.ConditionFalse,
-				Reason: reason,
-			})
-			return ctrl.Result{}, nil
+	if bundle.Status.Phase != rukpakv1alpha1.PhaseUnpacked {
+		reason := rukpakv1alpha1.ReasonUnpackPending
+		status := metav1.ConditionTrue
+		message := fmt.Sprintf("Waiting for the %s Bundle to be unpacked", bundle.GetName())
```
joelanford, line 128: For this section where the bundle is not unpacked, would it make sense to grab most of this directly from the Bundle `Unpacked` condition and just bundle it up?

Would it make sense to get rid of the BI `HasValidBundle` condition and just take the latest bundle's `Unpacked` condition directly?

internal/provisioner/plain/controllers/bundleinstance_controller.go
```diff
@@ -262,12 +283,84 @@ func (r *BundleInstanceReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+	if _, err := r.reconcileOldBundles(ctx, oldBundles, bundle); err != nil {
+		meta.SetStatusCondition(&bi.Status.Conditions, metav1.Condition{
+			Type:    rukpakv1alpha1.TypeProgressing,
+			Status:  metav1.ConditionFalse,
+			Reason:  rukpakv1alpha1.ReasonCleanupFailed,
+			Message: err.Error(),
+		})
+		return ctrl.Result{}, err
+	}
+	meta.RemoveStatusCondition(&bi.Status.Conditions, rukpakv1alpha1.TypeProgressing)
```
joelanford, line 299: I'm not sure if guidance has changed, but I read that its best to keep _all_ of the possible condition types populated at all times. Essentially, a missing type is interpretted as "Unknown", so if we know that we're not progressing, we should explicitly set it to "False".

internal/provisioner/plain/controllers/bundleinstance_controller.go
```diff
@@ -262,12 +283,84 @@ func (r *BundleInstanceReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
-		Message: fmt.Sprintf("instantiated bundle %s successfully", bi.Spec.BundleName),
+		Message: fmt.Sprintf("instantiated bundle %s successfully", bundle.GetName()),
 	})
-	bi.Status.InstalledBundleName = bi.Spec.BundleName
+	bi.Status.InstalledBundleName = bundle.GetName()
+
+	if _, err := r.reconcileOldBundles(ctx, oldBundles, bundle); err != nil {
+		meta.SetStatusCondition(&bi.Status.Conditions, metav1.Condition{
+			Type:    rukpakv1alpha1.TypeProgressing,
+			Status:  metav1.ConditionFalse,
```
joelanford, line 293: What's your working definition of "Progressing". My initial thought for this would have been Progressing=true, since we're going to try to re-reconcile and ostensibly attempt cleanup again.

internal/provisioner/plain/controllers/bundleinstance_controller.go
```diff
@@ -262,12 +283,84 @@ func (r *BundleInstanceReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+	if err != nil {
+		return nil, nil, err
+	}
+	sort.Sort(util.BundlesByCreationTimestamp(existingBundles))
+
+	// check whether the BI controller has already reached the maximum
+	// generated Bundle limit to avoid hotlooping scenarios.
+	if len(existingBundles) > maxGeneratedBundleLimit {
+		return nil, nil, ErrMaxGeneratedLimit
+	}
```
joelanford, line 322: How does deployment handle this with replica sets? Can we just start deleting the oldest bundles we have if we reach the limit? That way, we can always try to create the desired bundle as specified by the BI spec.

internal/provisioner/plain/controllers/bundleinstance_controller.go
```diff
@@ -262,12 +283,76 @@ func (r *BundleInstanceReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+
+	// check whether there's an existing Bundle that matches the desired Bundle template
+	// specified in the BI resource, and if not, generate a new matches the template.
+	b := util.CheckExistingBundlesMatchesTemplate(existingBundles, bi.Spec.Template)
```
joelanford, line 320: What if we generated a hash of `bi.spec.template`, and used that hash in the name of the bundle we create. Then all you have to do is check if a bundle with that name already exists. Would that sort of comparison work?

internal/provisioner/plain/controllers/bundleinstance_controller.go
```diff
@@ -262,12 +283,76 @@ func (r *BundleInstanceReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		if err := c.Create(ctx, b); err != nil {
+			return nil, nil, err
+		}
+	}
+	return b, existingBundles, err
+}
+
+// reconcileOldBundles is responsible for garbage collecting any Bundles
+// that no longer match the desired Bundle template.
+func (r *BundleInstanceReconciler) reconcileOldBundles(ctx context.Context, oldBundles []*rukpakv1alpha1.Bundle, currBundle *rukpakv1alpha1.Bundle) error {
```
joelanford, line 340: Nit/question:

```suggestion
func (r *BundleInstanceReconciler) reconcileOldBundles(ctx context.Context, allBundles []*rukpakv1alpha1.Bundle, currBundle *rukpakv1alpha1.Bundle) error {
```

internal/util/sort.go
```diff
@@ -0,0 +1,15 @@
...
+type BundlesByCreationTimestamp []*rukpakv1alpha1.Bundle
+
+func (o BundlesByCreationTimestamp) Len() int      { return len(o) }
+func (o BundlesByCreationTimestamp) Swap(i, j int) { o[i], o[j] = o[j], o[i] }
+func (o BundlesByCreationTimestamp) Less(i, j int) bool {
+	if o[i].CreationTimestamp.Equal(&o[j].CreationTimestamp) {
+		return o[i].Name < o[j].Name
+	}
+	return o[i].CreationTimestamp.Before(&o[j].CreationTimestamp)
+}
```
joelanford, line 15: Any consideration for using `sort.Slice` with this `Less` function re-implemented as an anonymous function?

internal/util/util.go
```diff
@@ -37,23 +39,76 @@ func BundleInstanceProvisionerFilter(provisionerClassName string) predicate.Pred
 
 func MapBundleToBundleInstanceHandler(cl client.Client, log logr.Logger) handler.MapFunc {
 	return func(object client.Object) []reconcile.Request {
-		b := object.(*rukpakv1alpha1.Bundle)
+		b, ok := object.(*rukpakv1alpha1.Bundle)
+		if !ok {
+			return nil
+		}
```
joelanford, line 45: Did you run into a problem here? I wonder if we should let this panic to warn us that we're using this in an unintended way.

internal/util/util.go
```diff
@@ -37,23 +39,76 @@ func BundleInstanceProvisionerFilter(provisionerClassName string) predicate.Pred
...
 		}
+		var requests []reconcile.Request
 		for _, bi := range bundleInstances.Items {
 			bi := bi
-			if bi.Spec.BundleName == b.Name {
-				requests = append(requests, reconcile.Request{NamespacedName: client.ObjectKeyFromObject(&bi)})
+			for _, ref := range b.GetOwnerReferences() {
+				if ref.Name == bi.GetName() {
+					requests = append(requests, reconcile.Request{NamespacedName: client.ObjectKeyFromObject(&bi)})
+				}
```
joelanford, line 57: Should we go by label selector instead of owner reference? Or maybe both?

internal/util/util_test.go
```diff
@@ -0,0 +1,143 @@
...
+	"testing"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/operator-framework/rukpak/api/v1alpha1"
+	rukpakv1alpha1 "github.com/operator-framework/rukpak/api/v1alpha1"
+	plain "github.com/operator-framework/rukpak/internal/provisioner/plain/types"
+)
+
+func TestCheckDesiredBundleTemplate(t *testing.T) {
```
joelanford, line 13: Do we have a style guide for our repo's tests? I recently re-implemented the storage tests to use ginkgo/gomega thinking that's what we had more recently gravitated to. But maybe we haven't?

timflannagan: Nope - I can refactor to use ginkgo as a follow-up? I use vscode locally which has some nice test autogeneration features, vs. having to bootstrap ginkgo style tests myself.

internal/util/util.go
timflannagan: Nope - I can revert this change.

internal/util/util.go
timflannagan: I can update to use a label selector implementation.

internal/provisioner/plain/controllers/bundleinstance_controller.go
```diff
@@ -262,12 +283,76 @@ func (r *BundleInstanceReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+				OwnerReferences: []metav1.OwnerReference{*controllerRef},
+				Labels:          bi.Spec.Template.Labels,
+			},
+			Spec: bi.Spec.Template.Spec,
+		}
+		if err := c.Create(ctx, b); err != nil {
+			return nil, nil, err
+		}
+	}
+	return b, existingBundles, err
```
joelanford, line 335: When we call this function, we call the second return value `oldBundles` in the calling scope, but here it looks like its all of the bundles that match the selector.  Should we go ahead of make sure that `b` is not present in `existingBundles`?

Or change the calling scope variable name to `existingBundles`?

internal/util/util.go
timflannagan: I might have deleted my focus for the day but I wonder whether we should be injecting the core.rukpak.io/owner-kind label on the generated Bundle resource so we only requeue owning BundleInstance resources vs. any BundleInstance that has matching labels?

internal/provisioner/plain/controllers/bundleinstance_controller.go
timflannagan: What happens if we have a bug in the comparison logic such that deleting the Bundle with the oldest creation timestamp just continuously results in new Bundles being generated? I had added this condition check mainly to prevent hotlooping behavior while I was still introducing tests, so I only intended for this to be a short term bandaid. I think we can either remove this logic in this PR, or implement a more robust check that mirrors the deployment implementation to prevent this behavior as a follow-up. Thoughts?

internal/provisioner/plain/README.md
```diff
@@ -95,9 +95,9 @@ Surfacing the content of a bundle in a more user-friendly way, via a plugin or a
 
 The `BundleInstance` API is meant to indicate the version of the bundle that should be active within the cluster. Given
 two unpacked bundles in the cluster,
-`my-bundle-v0.0.1` and `my-bundle-v0.0.2`, the `spec.bundleName` field of the related `BundleInstance` can be updated to
+`my-bundle-v0.0.1` and `my-bundle-v0.0.2`, the `spec.Bundle.Name` field of the related `BundleInstance` can be updated to
```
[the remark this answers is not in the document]
timflannagan: Yep - updated the documentation in the last force push.

internal/provisioner/plain/controllers/bundleinstance_controller.go
joelanford: I'm good either way. It seems like once we work any kinks out, we shouldn't hotloop because the next reconciliation we'll see that the latest bundle matches the desired template (unless of course that template changes again)

internal/util/util.go
joelanford: I have to reboot my memory on that, to be honest. IIRC those labels we're mainly used for `SelectorsByObject` to limit what the informers watch/cache.

internal/util/util_test.go
joelanford: I personally like ginkgo/gomega, but I don't think my opinion should be weighted as heavily as others. Do what folks think is most productive.

internal/util/util.go
joelanford: I think deployments/replica sets can technically adopt overlapping sets of pods. I have no idea what that would mean or whether there's an actual use case for that.

internal/provisioner/plain/controllers/bundleinstance_controller.go
timflannagan: I ended up dropping bubbling up any status conditions when encountering an error while cleaning up stale Bundles while we figure out the correct UX for this.

internal/provisioner/plain/controllers/bundleinstance_controller.go
timflannagan: See https://github.com/operator-framework/rukpak/pull/293#discussion_r865332918 - I ended up deciding to drop this implementation in favor of logging and returning an error. We can iron out the desired UX over time?

internal/provisioner/plain/README.md
timflannagan: I think this is worth chasing but I think we'll end up overhauling the BI's status (and therefore implicitly the printer columns) relatively soon, and I think we'll end up mirroring the deployment printer columns where it uses booleans vs. strings to expose state.

api/v1alpha1/bundleinstance_types.go
timflannagan: I'd like to keep the changes here minimal given the size of this PR already :laughing:. That said: I don't have a strong opinion here as these changes would only affect the Go types as users would still be typing `template` and `selector` as that's the json key. Maybe we can revisit this over time?

api/v1alpha1/bundleinstance_types.go
```diff
@@ -46,18 +46,28 @@ const (
 type BundleInstanceSpec struct {
 	// ProvisionerClassName sets the name of the provisioner that should reconcile this BundleInstance.
 	ProvisionerClassName string `json:"provisionerClassName"`
+	// Label selector for Bundles resources.
+	Selector *metav1.LabelSelector `json:"selector"`
```
exdx, line 51: Is having the selector mandatory an implication of the implementation? If I just want to create an embedded BI, I would have to provide a label selector as well, is that UX a bit complex? Could we not just create a default selector if one is not provided by the client?

api/v1alpha1/bundleinstance_types.go
```diff
@@ -46,18 +46,28 @@ const (
...
-	BundleName string `json:"bundleName"`
+// BundleTemplate defines the desired state of a Bundle resource
+type BundleTemplate struct {
+	// Standard object's metadata.
+	// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata
+	// +optional
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	// Specification of the desired behavior of the Bundle.
+	// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#spec-and-status
+	// +optional
```
exdx, line 66: Is there a use-case where the template is provided but the spec is omitted? ie should the spec be optional?

api/v1alpha1/bundleinstance_types.go
```diff
@@ -57,6 +57,8 @@ type BundleTemplate struct {
 	// Standard object's metadata.
 	// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata
 	// +optional
+	// +kubebuilder:pruning:PreserveUnknownFields
+	// +kubebuilder:validation:Schemaless
```
exdx, line 62: How come we don't want to validate the schema for this object? Is it to make the YAML UX more ergonomic, or to allow the client to pass in additional fields we don't necessarily reconcile on?

internal/util/util.go
```diff
@@ -38,22 +40,73 @@ func BundleInstanceProvisionerFilter(provisionerClassName string) predicate.Pred
...
+			continue
+		}
+		return bundle
+	}
+	return nil
+}
+
+// CheckDesiredBundleTemplate is responsible for determining whether the existingBundle
+// parameter is semantically equal to the desiredBundle Bundle template.
+func CheckDesiredBundleTemplate(existingBundle *rukpakv1alpha1.Bundle, desiredBundle *rukpakv1alpha1.BundleTemplate) bool {
```
exdx, line 98: nit: does this function have to be exported if its only called in this package in the above function?

exdx (commented):
Nice work! Just some thoughts. The detailed description made it easier to review.

api/v1alpha1/bundleinstance_types.go
tylerslaton: Sure, we can revisit at another time. I will say that my original point was more to edit that `json` values as well. I'll update that diff and unresolve so you can see this comment but feel free to close it right after. I think it is fine to edit later since we're still in iteration station.

timflannagan: Ah okay that makes more sense. In that case, that's where I'd start pushing back as this PR is implementing the design that had outlined those fields. We can revisit that concern over time.

internal/util/sort.go
timflannagan: That seems like an interesting implementation. I think I can simplify the sorting mechanism anyways so I'll play around with that locally and push up a separate commit if I can something working.

timflannagan: Updated the implementation to use sort.Slice in https://github.com/operator-framework/rukpak/pull/293/commits/bd136663720eb0439aef70265af7b29c88d4e586

timflannagan:
Notes from call for follow ups:

- Blocking:
  - Add support for reconciling metadata.Annotations within a Bundle's template
    - Joe: hash the template, and use that hash for that Bundle's metadata.Name so the lookup logic for an existing Bundle is more straight forward
  - Label selectors (spec.Selector) being a required field for all provisioners?
    - Do we need label selectors? Rely on the core.rukpak.io/owner-kind label instead?
    - Use case is users want to inject their own labels/annotations for the generated Bundle
    - Punt on a top-level spec.Select metav1.LabelSelector field until we have a concrete use case where this is necessarily
- Follow up:
  - Re-evaluate whether we need a safeguard around a limit on how many Bundles should be generated at a time
    - Do we need more than two Bundles at a time (one active/one pending)
    - Upgrade fails we roll back?
    - Keeping multiple helm release secrets around at a time
  - Shore up the UX around the BI printer columns to more-or-less emulate the Deployment/RS boolean values over string
  - Do we have a need for 1:N BundleInstance resources for subscription-esq scenarios?
    - Current expectation is that a higher level controller would manage an individual BundleInstance resource
    - Note: there's no adoption mechanism for an existing BundleInstance resources
      - RukPak's plain provisioner packaged as a BundleInstance resource

api/v1alpha1/bundleinstance_types.go
timflannagan: Dropped this field in the most recent set of commits.

internal/util/util.go
timflannagan: I think you were looking at an older version of this PR as I've since changed it to be called in tests as well.

internal/util/util_test.go
```diff
@@ -0,0 +1,154 @@
...
+				existingBundle: &rukpakv1alpha1.Bundle{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "stub",
+						Labels: map[string]string{
+							"stub": "stub",
+						},
+					},
+					Spec: rukpakv1alpha1.BundleSpec{
+						ProvisionerClassName: "non-existent-provisioner-class-name",
+					},
```
akihikokuroda, line 68: nit: This is an invalid bundle.  It may be better adding the `Source` as-is here.

timflannagan: Yep - I'll probably hold off on addressing this though as I think I'll refactor this test to use gomega as a follow-up.

api/v1alpha1/bundleinstance_types.go
joelanford: I don't see anything like this in the PodTemplateSpec, so I wonder if/how they solved it for deployments/statefulsets/etc. Seems a little scary to allow arbitrary fields to be added and for us not to do validation of anything inside, unless I'm misinterpreting.

api/v1alpha1/bundleinstance_types.go
joelanford: Seems like the adoption use case maybe?

internal/provisioner/plain/README.md
```diff
@@ -123,57 +127,44 @@ Once the cluster has been setup, take the following steps:
...
 spec:
-  source:
-    type: image
-    image:
-      ref: quay.io/tflannag/bundle:combo-operator-v0.0.1
   provisionerClassName: core.rukpak.io/plain
+  template:
+    metadata:
+      labels:
+        subscription: combo
```
joelanford, line 143: nit: people will likely copy this pattern if its in our docs. curious what the thought process is behind the use of "subscription" as a term here. maybe "package" or even more generic, "app", instead?

internal/util/util.go
joelanford: Here's what deployment does for listing its replica sets. Looks like it uses the label selector and then furhter filters by the owner ref.

internal/provisioner/plain/controllers/bundleinstance_controller.go
```diff
@@ -72,15 +82,11 @@ type BundleInstanceReconciler struct {
...
-// TODO(user): Modify the Reconcile function to compare the state specified by
-// the BundleInstance object against the actual cluster state, and then
-// perform operations to make the cluster state reflect the state specified by
-// the user.
 //
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.9.2/pkg/reconcile
 func (r *BundleInstanceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	l := log.FromContext(ctx)
+	l := log.FromContext(ctx).WithValues("req", req.NamespacedName)
```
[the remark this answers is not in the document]
timflannagan: This was just a debug artifact -- dropping in the next force push.

README.md
timflannagan: Ended up dropping this field as it wasn't clear whether we'll need to support it in the short term.

api/v1alpha1/bundleinstance_types.go
timflannagan: I had copied the [PodTemplateSpec](https://github.com/kubernetes/api/blob/a2ee8c7e6bdc1744015c35711607d5403bbd1d39/core/v1/types.go#L3953) type definitions for this, but it's unclear to me what the rationale was making the spec field optional (and following the git blame trail doesn't help illuminate that decision either).

I'm fine with making this required until we have a concrete use case?

api/v1alpha1/bundleinstance_types.go
timflannagan: > Seems a little scary to allow arbitrary fields to be added and for us not to do validation of anything inside, unless I'm misinterpreting

This seems valid but I'll caution that adding arbitrary metadata fields doesn't result in the plain provisioner respecting them. The current implementation only respects propagating the labels/annotations right now.

api/v1alpha1/bundleinstance_types.go
timflannagan: Actually: does leaving it optional give us more leeway here? Wouldn't making it required -> optional down the line to support something like adoption require us to make a breaking change to the API?

exdx (approved):

internal/util/util.go
timflannagan: We hashed this out in the [rukpak-dev slack channel](https://github.com/operator-framework/rukpak/pull/293/commits/d7bb57295d7037817838f6222d05cf28f526ec52).

Update the implementation to use label selectors in https://github.com/operator-framework/rukpak/pull/293/commits/d7bb57295d7037817838f6222d05cf28f526ec52.

api/v1alpha1/bundleinstance_types.go
joelanford: Right, I guess my concern is the UX of "we'll let you create stuff in the BI template, but then we'll silently ignore some of it"

It'd be best, IMO, to block (via CRD schema or a webhook) BI creation/update if it contains something that we know we'll ignore or will cause a Bundle creation failure.

internal/provisioner/plain/controllers/bundleinstance_controller.go
timflannagan: > Would it make sense to get rid of the BI HasValidBundle condition and just take the latest bundle's Unpacked condition directly?

I'm not opposed to this suggestion but it's also a problem with the existing implementation so I'd like to hold off on changing this here.

api/v1alpha1/bundleinstance_types.go
joelanford: required -> optional is non-breaking because all existing instance will have set it already.
optional -> required is breaking because there will be some instances that have not set it, and now we're requiring them to be set.

timflannagan: Updated to make spec required in the last force push.

joelanford (approved):
Awesome work! Excited to get this in.

internal/provisioner/plain/controllers/bundleinstance_controller.go
timflannagan: Created #333

api/v1alpha1/bundleinstance_types.go
timflannagan: Resolved in [slack](https://kubernetes.slack.com/archives/C038B7MF75M/p1651866914692939?thread_ts=1651865835.897279&cid=C038B7MF75M). Joe had discovered that there's a controller-gen marker for this, and we removed the kubebuilder preserveUnknownFields and schemaLess markers from these set of changes.

internal/provisioner/plain/controllers/bundleinstance_controller.go
timflannagan: Created https://github.com/operator-framework/rukpak/issues/334 as a follow-up item.

PR #303, opened 2022-05-03
internal,test: Add a package that contains the unique provisioner IDs

timflannagan:
Introduce a types.go package that contains the constant variables used throughout the plain provisioner implementation. Update references to that plain provisioner unique ID to point to that consistent constant variable vs. scattering this information across the repository.

### DIFF
--- a/internal/provisioner/plain/controllers/bundle_controller.go
+++ b/internal/provisioner/plain/controllers/bundle_controller.go
@@ -43,6 +43,7 @@ import (
 
 	rukpakv1alpha1 "github.com/operator-framework/rukpak/api/v1alpha1"
 	"github.com/operator-framework/rukpak/internal/git"
+	plain "github.com/operator-framework/rukpak/internal/provisioner/plain/types"
 	"github.com/operator-framework/rukpak/internal/storage"
 	"github.com/operator-framework/rukpak/internal/updater"
 	"github.com/operator-framework/rukpak/internal/util"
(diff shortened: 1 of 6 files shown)


tylerslaton (approved):
Not blocking, but maybe each individual provisioner should own its own ID?

timflannagan:
> Not blocking, but maybe each individual provisioner should own its own ID?

I definitely agree: I didn't see a super clean way to do this with our current plain provisioner directory structure, but I'm open to ideas.

oceanc80:
Maybe something like `rukpak/internal/provisioner/plain/id.go`?

tylerslaton:
@oceanc80 Yeah that makes sense structurally but unfortunately you can't import `main` packages in Golang. It'll provide you with this error:

```console
main.go:6:2: import "github.com/operator-framework/rukpak/internal/provisioner/plain" is a program, not an importable package
```

internal/provisioner/ids/ids.go
```diff
@@ -0,0 +1,5 @@
+package provisioner
+
+const (
+	PlainID = "core.rukpak.io/plain"
```
exdx, line 4: nit: would it be helpful to have a comment here explaining what this id is and what it's being used for? Especially if we expect to group other provisioner IDs here as well eventually.

exdx (approved):
LGTM. I think its fine moving this into its own package or leaving it under the provisioner package for now.

internal/provisioner/ids/ids.go
timflannagan: Updated in the last force push.

oceanc80 (approved):

PR #304, opened 2022-05-03
Add quickstart section to root-level README

### LINKED ISSUE #249: Consolidate quickstart guides
It looks like there are a variety of quickstart guides (e.g. [this one](https://github.com/operator-framework/rukpak/blob/main/docs/plain-bundle-spec.md#quickstart) and [this one](https://github.com/operator-framework/rukpak/tree/main/internal/provisioner/plain)) and it would be a bit clearer to consolidate them to streamline the first time UX.

# Acceptance Criteria
- [ ] Clear step-by-step guide for getting started with RukPak, ideally in the main README

exdx:
Closes #249

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,24 @@ the content.
 The RukPak project is community driven and is part of the broader Kubernetes ecosystem. New contributors are welcome and
 highly encouraged. See the [contributing guidelines](CONTRIBUTING.md) to get started.
 
+## Quickstart
+
+The RukPak project consists of a series of controllers, known as [provisioners](#provisioner), that install and manage
+content on a Kubernetes cluster. See [below](#components) for a more detailed look into the APIs that RukPak provides.
+
+The provisioner currently implemented and bundled with RukPak is known as the plain provisioner. To get started with
+this provisioner on a local kind cluster,
+see [the quickstart section](./internal/provisioner/plain/README.md#Running-locally) of the plain provisioner README.
+There will be other provisioners added to the RukPak project that support different content types.
+
+The plain provisioner is able to source and unpack plain bundles. To learn more about the plain bundle format,
+see [the plain bundle spec](./docs/plain-bundle-spec.md).
+
+This project uses GitHub issues and milestones to prioritize and keep track of ongoing work. To see the current state of
+the project, checkout the [open issues](https://github.com/operator-framework/rukpak/issues) and
+[recent milestones](https://github.com/operator-framework/rukpak/milestones). There is a lot of development work ongoing
+and plenty of opportunities to get involved and [contribute](CONTRIBUTING.md)!
+
 ## Components
 
 RukPak is composed of two primary APIs, [Bundle](#bundle) and [BundleInstance](#bundleInstance), as well as the concept
@@ -108,3 +126,5 @@ actions to the cluster, it will not allow it go be applied. In the context of Ru
 `BundleInstance` resolution. 
 
 To read more about this webhook, and learn how to disable this default behavior, view the `crdvalidator` [documentation](cmd/crdvalidator/README.md).
+The `plain` provisioner is able to unpack a given `plain+v0` bundle onto a cluster and then instantiate it, making the
+content of the bundle available in the cluster.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,7 +8,7 @@ to help RukPak contributors.
 ## Communication
 
 - Email: [operator-framework-olm-dev](mailto:operator-framework-olm-dev@googlegroups.com)
-- Slack: [#olm-dev](https://kubernetes.slack.com/archives/C0181L6JYQ2)
+- Slack: [#rukpak-dev](https://kubernetes.slack.com/archives/C038B7MF75M)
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -21,6 +21,23 @@ the content.
 The RukPak project is community driven and is part of the broader Kubernetes ecosystem. New contributors are welcome and
 highly encouraged. See the [contributing guidelines](CONTRIBUTING.md) to get started.
 
+## Quickstart
+
+The RukPak project consists of a series of controllers, known as provisioners, that install and manage content on a
```
timflannagan, line 26: Link to the provisioner header in the README here?

README.md
```diff
@@ -21,6 +21,23 @@ the content.
...
+this provisioner, see [the quickstart section](./internal/provisioner/plain/README.md#Running-locally) of the plain
+provisioner README. There will be other provisioners added to the RukPak project that support different content types.
+
+The plain provisioner is able to source and unpack plain bundles. To learn more about the plain bundle format,
+see [the plain bundle spec](./docs/plain-bundle-spec.md).
+
+This project uses GitHub issues and milestones to prioritize and keep track of ongoing work. To see the current state of
+the project, checkout the [open issues](https://github.com/operator-framework/rukpak/issues) and
+[recent milestones](https://github.com/operator-framework/rukpak/milestones). There is a lot of development work ongoing
+and lots of opportunities to get involved and contribute!
```
timflannagan, line 39: Link to the CONTRIBUTING.md documentation?

timflannagan (approved):

timflannagan:
Nice work as always - only question: we're referencing an issue that's centered around consolidating quickstart guides yet it seems like these changes are more geared towards an introduction of the plain provisioner, vs. actually consolidating quickstart guides. Should we create a different ticket here instead?

README.md
```diff
@@ -21,6 +21,23 @@ the content.
 The RukPak project is community driven and is part of the broader Kubernetes ecosystem. New contributors are welcome and
 highly encouraged. See the [contributing guidelines](CONTRIBUTING.md) to get started.
 
+## Quickstart
+
+The RukPak project consists of a series of controllers, known as provisioners, that install and manage content on a
+Kubernetes cluster. See below for a more detailed look into the APIs that RukPak provides.
```
tylerslaton, line 27: Might be worth linking to where this starts in the document so that a user can easily navigate to it.

README.md
```diff
@@ -21,6 +21,23 @@ the content.
...
+
+The provisioner currently implemented and bundled with RukPak is known as the plain provisioner. To get started with
+this provisioner, see [the quickstart section](./internal/provisioner/plain/README.md#Running-locally) of the plain
```
oceanc80, line 30: Might be worth specifying that this quickstart is for running locally vs. on a cluster.

exdx:
> Nice work as always - only question: we're referencing an issue that's centered around consolidating quickstart guides yet it seems like these changes are more geared towards an introduction of the plain provisioner, vs. actually consolidating quickstart guides. Should we create a different ticket here instead?

My impression of the issue was that the quickstarts were scattered and not easy to navigate, and that we should introduce a section to the root-level README that helps to navigate the different docs. I didn't think that the intent was to consolidate necessarily, since each doc is valuable on its own and has its own concept of quickstart/demo. The goal of the new section being introduced here is to make it easier for a first time user to onboard and navigate the different documents. Unless I misunderstood the spirit of the original issue, I don't think a new ticket is necessary -- maybe @oceanc80 could clarify if this PR doesn't meet the requirements?

oceanc80:
@exdx I think once we have more provisioners it'd be nice to surface examples of running locally and on a cluster with the plain provisioner directly in this readme, but I think for now as long as we have the link to the existing docs, that should be fine. We could maybe also add the [cluster quickstart link](https://github.com/operator-framework/rukpak/blob/main/docs/plain-bundle-spec.md#quickstart) as well just so both the local and cluster quickstarts are linked from this readme, but overall I think this is good for now.

oceanc80 (approved):

exdx:
> @exdx I think once we have more provisioners it'd be nice to surface examples of running locally and on a cluster with the plain provisioner directly in this readme, but I think for now as long as we have the link to the existing docs, that should be fine. We could maybe also add the [cluster quickstart link](https://github.com/operator-framework/rukpak/blob/main/docs/plain-bundle-spec.md#quickstart) as well just so both the local and cluster quickstarts are linked from this readme, but overall I think this is good for now.

I agree, yes once we have more provisioners we will have to reorganize the docs quite a bit, right now they are overly specified towards the plain provisioner and plain bundle format. 

For the plain bundle spec, I think it's still targeting a local installation, it just describes how to build a plain bundle image, which is slightly tangential to running rukpak locally or on a real cluster. I think you bring up a good point, that we should probably have some kind of target and documentation around installing rukpak on a real (non-kind) cluster. Right now we only have a `run` target to deploy rukpak which assumes its dealing with a kind cluster (we removed other `deploy` and `install` targets because they were causing unnecessary confusion). I will go ahead and open an issue for that.

exdx:
#312

PR #305, opened 2022-05-03
api,manifests: Add a short name for the BundleInstance API

timflannagan:
Update the kubebuilder markers for the BI types.go definition and add a shortName configuration that allows users to interact with bundleinstance resources locally without typing out the full name.

### DIFF
--- a/api/v1alpha1/bundleinstance_types.go
+++ b/api/v1alpha1/bundleinstance_types.go
@@ -61,8 +61,8 @@ type BundleInstanceStatus struct {
 }
 
 //+kubebuilder:object:root=true
-//+kubebuilder:resource:scope=Cluster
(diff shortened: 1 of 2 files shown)


timflannagan:
This is potentially controversial as the `bi` short name seems like it could cause collision down the line with unrelated APIs. I'll let others chime in though.

tylerslaton (approved):
I can't think of a really good substitute for just `bi`/`bis` that is useful for a UX. Unless someone can think of a collision off the top of their head I think we should be good to get this in as is.

timflannagan:
This is a bit backwards, but I'm going to create a tracking issue so it's slightly more clear what landed in the 0.4 milestone.

Issue #311, opened 2022-05-03
Define API for bundle addressability


joelanford:
Today, there's no first-class API that enables consumers of bundles (other than provisioners themselves) to fetch and view bundle content. In #292, we moved bundle content to an emptyDir volumeMount only visible to the provisioner pod.

If you know where to look (and you have the permission), you can run:
`kubectl exec <provisionerPodName> -n rukpak-system -- cat /var/cache/bundles/<bundleUID>.tgz` and then pipe that output to your local tar invocation of choice to view the file/directory contents of the archive or extract the full contents on your local file system.

However that command is only useful if you know that the provisioner has implemented storage with the `LocalDirectory` storage implementation. What if the provisioner used something else (e.g. S3)?

We should define a new API that makes bundle content generally accessible no matter the underlying provisioner/storage implementation. One strawman idea is to update our specs to state that:
1. All bundle provisioners _must_ host bundle content on a local port.
2. All bundle provisioners _must_ set a new status field called `Bundle.status.contentAddress` on `Bundle` objects they manage, such that when a client performs an http GET request to that address, the provisioner returns the tar.gz contents of that bundle.
3. (Maybe?) the bundle content endpoint should be protected by the kube-rbac-proxy and a specific RBAC rule for a well-defined non-resource URL (e.g. `bundles/content`), so that only clients with read permission for bundle content are able to fetch it.

joelanford:
The implementation for this was merged in #315. However there are still a few follow-ups that need to happen to fully close out this issue:
1. Add e2e tests that exercise the content http server of the plain provisioner (including authorization via kube-rbac-proxy)
2. Update documentation about this feature
3. Update any general provisioner specifications we have to make clear that other bundle provisioners are required to implement this.

tylerslaton:
I can take a stab at finishing up that clean-up if you're okay with it @joelanford?

exdx:
Remaining work: e2e tests and then we can close this issue.

exdx:
Based on the outcome of grooming, we can close this issue out and open a follow-up for the e2e test specifically.

PR #313, opened 2022-05-03
Remove bundle.status.info field

### LINKED ISSUE #314: Make `bundle.status` generic to the content of the bundle
The bundle status field currently has an `info` field that is defined as:
```go
type BundleInfo struct {
	Package string         `json:"package"`
	Name    string         `json:"name"`
	Version string         `json:"version"`
	Objects []BundleObject `json:"objects,omitempty"`
}

type BundleObject struct {
	Group     string `json:"group"`
	Version   string `json:"version"`
	Kind      string `json:"kind"`
	Name      string `json:"name"`
	Namespace string `json:"namespace"`
}
```

This makes an assumption that all bundles will be expandable to a list of kubernetes objects. However there's nothing about the general bundle spec that says that bundles must contain objects. For example, a helm chart could be a bundle. A Helm chart contains chart metadata, a default values file, and a bunch of templates. It isn't until you render a chart that you have plain manifests. And a provisioner may decide that the rendering step should be defered to `BundleInstance` reconciliation, which would require the bundle to be stored as metadata, default values, and templates.

For now, I propose the we remove the `BundleInfo` field from the bundle status. This _is_ a regression in UX for the plain provisioner because it takes visibility away from users wanting to see some information about the bundle contents.

However, I don't think the bundle status is the correct location for such potentially verbose information. I think we should cover this visibility concern via #311.

joelanford:
This field is not generic enough to cover all potential bundle formats.
For example, a helm chart does not contain specific `client.Object`s.
Rather, it contains a helm chart directory containing metadata, default
values, and some templates, which can be _rendered_ into manifests.

Thus, we should remove this field. If we later determine that there are
other fields that _do_ apply to all bundles, we can add them in
follow-on pull requests.

Closes #314 


timflannagan:
Should we create a tracking issue for this?

exdx (commented):
I agree with your sentiment that not all Bundle types would conform to have this type of information on them, but I really like the deep level of information that we have now on the Bundle type 🤷 . It makes them much less opaque than in other systems, where you can see at a glance exactly what is installed. 

Is there any way forward without removing the field?

joelanford:
@timflannagan Done, see #314 

@exdx I agree that this makes UX worse for the plain provisioner. However I think it outright precludes a helm or kustomize provisioner from integrating well (or any other provisioner for a bundle format that isn't just straight manifests). I think we can mostly solve this concern with #311.

Issue #210, opened 2022-04-01
Introduce minimal documentation for cutting new releases


timflannagan:
Goal: introduce some light documentation that's centered around how to cut releases, and which steps aren't automated by goreleaser right now.

timflannagan:
Note: just dumping the command I used to create the latest rukpak.yaml release manifest for the v0.2.0 release so I don't forget again how I did this:

```bash
yq manifests/*.yaml manifests/**/*.yaml internal/provisioner/plain/manifests/*.yaml | sed "s/:latest/:v0.2.0/g" > rukpak.yaml
```

timflannagan:
And it looks like I forgot to override the default unpacker image used in the plain provisioner deployment manifest:

```yaml
          args:
          - --unpack-image
          - quay.io/operator-framework/plain-provisioner:v0.2.0
 ```

timflannagan:
Moving this to the 0.3 backlog just so we're forced to have something in place in the short term.

timflannagan:
Moved this to 0.4 as I just pushed the 0.3 up locally. Note: the command I had outlined in https://github.com/operator-framework/rukpak/issues/210#issuecomment-1086356056 won't work anymore now that we're working towards a kustomize-based approach for stamping out the rukpak-related manifests.

Side note: it would be nice to have automation that tests the release quickstart manifests when a new release has been cut. I think that would just involve the following steps:

- Create a kind cluster
- Install cert manager
- Wait until cert manager is available
- Apply the rukpak.yaml release manifests
- Wait until the rukpak stack is ready (although I don't think we have any general checks for this right now)
- Load the e2e testdata images via `make kind-load-bundles`
- Run the e2e suite against that custom rukpak deployment via `make test-e2e`

exdx:
I really like the idea of an automated smoke test when a new release tag is pushed. 

Side note: from the conversation here https://github.com/operator-framework/rukpak/pull/316#discussion_r865133374 it seems like we don't necessarily want to include the cert-manager installation step as part of the rukpak install. Maybe we can remove that portion from the release description going forward and instruct users to install cert-manager first (linking them to the appropriate doc). It may make more sense since outside of CI environments they may have a version of it on-cluster already.

timflannagan:
> I really like the idea of an automated smoke test when a new release tag is pushed.

Just for housekeeping: related to https://github.com/operator-framework/rukpak/issues/263.

timflannagan:
[...]

PR #316, opened 2022-05-04
Add install section to the root-level README

### LINKED ISSUE #312: Support deploying rukpak to non-kind clusters
Currently, the only way to deploy rukpak is through the `run` target, which assumes that the cluster being deployed to is a kind cluster (it attempts to `kind load` the provisioner image onto the cluster). Other targets to deploy rukpak were removed in #209 because they were potentially confusing to new users of the project. That said, Rukpak should provide a way to deploy itself to a non-CI cluster.

A/C:
- provide a makefile target that enables rukpak to be deployed to a live cluster (based on a real tag)
- update documentation to reflect the possibility of installing to a non-CI cluster

exdx:
This PR adds an installation section to the top-level README so new users of the project can at a glance see how to quickly install RukPak. Additional discussion around this in #312.

Closes #312

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,8 +28,9 @@ content on a Kubernetes cluster. See [below](#components) for a more detailed lo
 
 The provisioner currently implemented and bundled with RukPak is known as the plain provisioner. To get started with
 this provisioner on a local kind cluster,
-see [the quickstart section](./internal/provisioner/plain/README.md#Running-locally) of the plain provisioner README.
-There will be other provisioners added to the RukPak project that support different content types.
+see [the quickstart section](./internal/provisioner/plain/README.md#Running-locally) of the plain provisioner README. To
+install the latest version of the provisioner on an existing cluster, see the [installation guide](#install). There will
+be other provisioners added to the RukPak project that support different content types.
 
 The plain provisioner is able to source and unpack plain bundles. To learn more about the plain bundle format,
 see [the plain bundle spec](./docs/plain-bundle-spec.md).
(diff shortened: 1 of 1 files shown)


tylerslaton (approved):
🚀

README.md
```diff
@@ -39,6 +40,27 @@ the project, checkout the [open issues](https://github.com/operator-framework/ru
 [recent milestones](https://github.com/operator-framework/rukpak/milestones). There is a lot of development work ongoing
 and plenty of opportunities to get involved and [contribute](CONTRIBUTING.md)!
 
+## Install
+
+The recommended way of installing RukPak is via a tagged release from
+the [releases](https://github.com/operator-framework/rukpak/releases) page. There are detailed instructions provided in
+the release notes on how to install a particular release. The only requirement is to have a `kubectl` client available
```
joelanford, line 47: We can link directly to the `rukpak.yaml` file of the latest release like this:
```
kubectl apply -f https://github.com/operator-framework/rukpak/releases/latest/download/rukpak.yaml
```

Only caveat is that if we ever backport something and cut a release of an old patch version, I'm pretty sure GH will treat that as the latest. Basically "latest" to github is based on release timestamp, not semver.

exdx: Ok that's fair -- we can include applying the rukpak.yaml file directly in the README -- I was unsure since there is a cert-manager dependency, but I can elaborate on that as well.

joelanford: I wonder if we should include a copy of the cert-manager manifests we depend on in our release. Then we could do:
```
kubectl apply -f https://github.com/operator-framework/rukpak/releases/latest/download/cert-manager.yaml
kubectl apply -f https://github.com/operator-framework/rukpak/releases/latest/download/rukpak.yaml
```

joelanford: Or we could just call cert-manager out as a dependency and not include any official instructions on installing it?

exdx: Hmm, I'm not sure about vendoring cert-manager in, since it's a popular project chances are some users already have some version of it installed. We would benefit from just linking to an existing cert-manager on the cluster and avoid the issue of trying to install it if it's already present. We could attempt to conditionally install the dependency but it seems needlessly complicated. 

In our official documentation for the [v0.3 release](https://github.com/operator-framework/rukpak/releases/tag/v0.3.0) we do state a dependency on a specific version of cert-manager. I suppose we could remove that and include a note stating that the released version is dependent on cert-manager instead? 

I'm wondering if it makes sense to leave the release format as-is and from the root README just provide the rukpak installation link and explain that there is a dependency on cert-manager (with a link to how to install it on their doc site). That way users have flexibility in how they choose to install cert-manager.

joelanford: +1 to that approach. Except that for future releases, I think we could leave out the kubectl apply step that references the cert-manager manifest.

exdx: Cool, I can reference that in #210 so we can consider it when tackling that issue.

PR #318, opened 2022-05-04
Add markdown linter GH Action

exdx:
This PR introduces a new GH Action specifically for linting the markdown files in the repository. It relies on the https://github.com/github/super-linter project which seems to be the most maintained GH Action for linting and specifically for markdown. It only lints markdown files and does not use any of the other linters included, since golangci-lint is used for Go code in the project.

### DIFF
--- a/.github/workflows/sanity.yaml
+++ b/.github/workflows/sanity.yaml
@@ -21,7 +21,15 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: Run linting checks
+      - name: Run golangci linting checks
         uses: "golangci/golangci-lint-action@v2"
         with:
           version: "v1.45.2"
+
+      - name: Lint markdown files
+        uses: github/super-linter/slim@v4
+        env:
+          VALIDATE_ALL_CODEBASE: true
+          DEFAULT_BRANCH: main
+          # only runs the markdown linter
+          VALIDATE_MARKDOWN: true
--- a/README.md
+++ b/README.md
@@ -1,7 +1,6 @@
+# RukPak
 [![License](http://img.shields.io/:license-apache-blue.svg)](http://www.apache.org/licenses/LICENSE-2.0.html)
 [![Slack Channel](https://img.shields.io/badge/chat-4A154B?logo=slack&logoColor=white "Slack Channel")](https://kubernetes.slack.com/archives/C038B7MF75M)
-
-# RukPak
 
 RukPak runs in a Kubernetes cluster and defines an API for installing cloud native bundle content.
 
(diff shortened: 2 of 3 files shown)


exdx:
I suppose we could move this under the existing `sanity/lint` action?

oceanc80:
> I suppose we could move this under the existing `sanity/lint` action?
That seems reasonable to me.

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,57 @@
...
+      # Run Linter against code base #
+      ################################
+      - name: Lint Code Base
+        uses: github/super-linter/slim@v4
+        env:
+          VALIDATE_ALL_CODEBASE: true
+          DEFAULT_BRANCH: main
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          # only runs the markdown linter
+          VALIDATE_MARKDOWN: true
```
joelanford, line 57: Nit: remove the `######` comments (or at least the border comment lines that are just `###########`?

I feel like all of us are familiar enough with GH action config files that those lines amount to vertical-space-taking noise.

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,57 @@
...
+        uses: actions/checkout@v3
+        with:
+          # Full git history is needed to get a proper list of changed files within `super-linter`
+          fetch-depth: 0
+
+      ################################
+      # Run Linter against code base #
+      ################################
+      - name: Lint Code Base
+        uses: github/super-linter/slim@v4
```
joelanford, line 51: should we move this to the existing sanity workflow to keep it consolidated with the other linting related tasks?

.github/workflows/sanity.yaml
```diff
@@ -21,7 +21,15 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: Run linting checks
+      - name: Run golangci linting checks
         uses: "golangci/golangci-lint-action@v2"
         with:
           version: "v1.45.2"
+
+      - name: Lint markdown files
```
timflannagan, line 29: Should we move this into its own individual job.

timflannagan (approved):

timflannagan:
Also: should we separate out the markdown linting violations into it's own commit?

.github/workflows/sanity.yaml
exdx: Sure, it probably makes more sense that way

exdx:
> Also: should we separate out the markdown linting violations into it's own commit?

Yes, they should be in their own commit. I squashed everything but will try to split it out 👍

.github/workflows/sanity.yaml
exdx: Actually, I kind of like having the markdown lint step be under the `lint` job since its technically also linting. The two lint steps share the same checkout action too which is nice. I don't know that we want a `sanity/markdown` job to run standalone, as it kind of clutters the UI, versus just keeping the markdown linter as part of the `sanity/lint` job and diagnosing any failures there as general lint errors.

timflannagan: Yep that's fine with me. The only consideration here is that multiple steps within an individual job are run sequentially vs. jobs which run in parallel.

exdx: Good point -- I think since the markdown job is fairly quick we can get away with this, but it's worth keeping in mind.

PR #320, opened 2022-05-05
WIP: FIX: adds GHA for e2e testing on rukpak release

### LINKED ISSUE #263: Introduce automation for testing new rukpak releases
Goal: introduce a new GHA workflow that's responsible for running e2e tests against new rukpak releases.

AC:
- Create a kind cluster
- Mirror the quickstep steps that are documented in the "Installation" section of a release
- Wait until the rukpak stack is healthy
- Load the e2e testdata images into that kind cluster
- Run the e2e suite against that custom rukpak release

nsapse:
adds a Github Action to test any potential releases of rukpak

Closes #263


.github/workflows/release-e2e.yaml
```diff
@@ -0,0 +1,51 @@
...
+      run: |
+        curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
+        sudo install -o root -g root -m 0755 kubectl /usr/local/bin/kubectl
+
+    - name: release-install
+      needs: [start_kind_cluster, install_kubectl]
+      run: |
+        kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.7.1/cert-manager.yaml
+        kubectl wait --for=condition=Available --namespace=cert-manager deployment/cert-manager-webhook --timeout=60s
+        kubectl apply -f https://github.com/operator-framework/rukpak/releases/download/v0.3.0/rukpak.yaml
```
tylerslaton, line 42: I'm not so sure that we need this step. The `make e2e` section will already do this install. Additionally, the release we are pointing at here is static so it won't change across runs.

.github/workflows/release-e2e.yaml
```diff
@@ -0,0 +1,51 @@
...
+        KIND_OS: kind-linux-amd64
+      run: |
+        curl -sLo kind "https://github.com/kubernetes-sigs/kind/releases/download/${KIND_VERSION}/${KIND_OS}"
+        chmod +x kind
+        sudo mv kind /bin/
+
+    - name: start_kind_cluster
+      run: |
+        # /bin/kind create cluster
+        make kind-cluster
```
exdx, line 29: nit: this target deletes a cluster first before creating a new one with a specific name. In this CI context though, there is never going to be a pre-existing cluster. Does it make more sense to simply create a cluster here with the default name? 
Ala
```
kind create cluster
```

exdx, line 29: I think we may also need to export the kubeconfig per https://github.com/operator-framework/operator-lifecycle-manager/blob/3002cf79ce867e32d30d2190df642eaa7f449294/.github/workflows/run-kind-local.yml#L23

.github/workflows/release-e2e.yaml
exdx, line 42: Can we infer the version of the release that we are testing based off the metadata provided in the release? We need to ensure that this step is applying the $latest release at each run.

exdx, line 42: Should we wait for the plain provisioner to rollout as well, like we do for cert-manager? Something like
```
kubectl wait --for=condition=Available --namespace=rukpak-system deployment/plain-provisioner --timeout=60s
```

.github/workflows/release-e2e.yaml
```diff
@@ -0,0 +1,51 @@
...
+        kubectl apply -f https://github.com/operator-framework/rukpak/releases/download/v0.3.0/rukpak.yaml
+
+    - name: load_bundles
+      needs: [release-install]
+      run: |
+        make kind-load-bundles
+
+    - name: Run e2e tests
+      needs: [load_bundles]
+      run: make e2e KIND=/bin/kind
```
exdx, line 51: `make e2e` actually builds the container and stands up a new cluster, which is redundant... We can run the e2e tests via ginkgo directly? Something like
```
go run github.com/onsi/ginkgo/v2/ginkgo ./test/e2e
```
we could potentially add a new target for this as well.

exdx (commented):
Nice! Just some thoughts.

.github/workflows/release-e2e.yaml
timflannagan: We already have an existing `test-e2e` Makefile target which runs ginkgo under-the-hood. I'd recommend re-using that target, and refactoring this check to avoid using the `e2e` target like Dan had mentioned as that target is performing a bunch of installation check that aren't relevant to running e2e tests against an existing rukpak installation that uses the release tagged images.

.github/workflows/release-e2e.yaml
timflannagan: Yup +1 here: we _might_ be able to get away with using the latest endpoint, but we also might be able to infer the tag here as we're only running this workflow during tag event triggers? I think my preference is highly skewed towards the latter implementation. 

The [release workflow](https://github.com/operator-framework/rukpak/blob/main/.github/workflows/release.yaml#L39) has a conditional that checks whether the git reference we're processing is a tag, and parses that value to pull out the tag value. We'd have to update this logic to use a variable in the URL:

```yaml
kubectl apply -f https://github.com/operator-framework/rukpak/releases/download/${IMAGE_TAG}/rukpak.yaml
```

.github/workflows/release-e2e.yaml
```diff
@@ -0,0 +1,51 @@
...
+        make kind-cluster
+
+
+    - name: install_kubectl
+      run: |
+        curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
+        sudo install -o root -g root -m 0755 kubectl /usr/local/bin/kubectl
+
+    - name: release-install
+      needs: [start_kind_cluster, install_kubectl]
```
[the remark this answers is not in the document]
nsapse: deleted

.github/workflows/release-e2e.yaml
nsapse: That makes sense to me. I was being lazy and trying to use a pre-made target to get functionality for free but I didn't realize it was creating potential issues.

.github/workflows/release-e2e.yaml
nsapse: That makes sense to me. Will add.

joelanford: Also the webhook. I've noticed local e2e runs that fail because the webhook pod isn't serving by the time we start hammering the API server with bundles.

timflannagan: @joelanford I've seen this exact behavior locally when copy-pasting our installation release instructions. Do we need an issue for fixing this behavior?

nsapse: Sorry, beginner question - a rollout wait for another webhook beyond the cert manager one (i.e. not just this `kubectl wait --for=condition=Available --namespace=cert-manager deployment/cert-manager-webhook --timeout=60s`?)

nsapse: Based on your reply it seems like leaving this is probably what's best. Am I understanding correctly, Tim?

.github/workflows/release-e2e.yaml
nsapse: Done

.github/workflows/release-e2e.yaml
exdx: Yes, I would add some checks to ensure that the additional resources and webhooks are up and available before starting the e2e tests. Something similar to #328 
```
	kubectl wait --for=condition=Available --namespace=rukpak-system deployment/plain-provisioner --timeout=60s
	kubectl wait --for=condition=Available --namespace=rukpak-system deployment/rukpak-core-webhook --timeout=60s
	kubectl wait --for=condition=Available --namespace=crdvalidator-system deployment/crd-validation-webhook --timeout=60s
```

.github/workflows/release-e2e.yaml
nsapse: Switched targets.

.github/workflows/release-e2e.yaml
timflannagan: @nsapse I think this is the only blocking comment on my end now as we can't hardcode the rukpak version (e.g. v0.3.0 in this case) due to the increase in maintenance over time.

nsapse: Pulled checks from the new release instructions.

nsapse: I couldn't get the example you linked working so I piped the tags through sort and took the most recent with head. It works but let me know if it's too much of a hack.

timflannagan: I _think_ I just had to do something similar to this yesterday: https://github.com/operator-framework/rukpak/blob/main/.github/workflows/release.yaml#L58. You should be able to just parse the tag reference using that `${GITHUB_REF#refs/tags/}` value?

nsapse: For some reason when I do that the URL I get is `https://github.com/operator-framework/rukpak/releases/download/refs/heads/release_actions/rukpak.yaml` i.e. the results of `${GITHUB_REF#refs/tags/}` is `relase_actions`

timflannagan: Yeah you'd need to add a conditional that checks whether the current HEAD is a tag reference. You can use the `if` field for conditional statements (and the link I had sent has the conditional you need on line 55).

nsapse: That's with the check pulled from your example. I must be implementing it incorrectly. I'll play around with it.

nsapse: Alright, working.

.github/workflows/release-e2e.yaml
```diff
@@ -0,0 +1,57 @@
+---
+name: release-e2e
+
+on:
+  release:
+    types: [created]
+  workflow_dispatch:
```
joelanford, line 7: Should this be on tag pushes rather than release creation?

Also, what's the purpose of running this on workflow_dispatch events?

.github/workflows/release-e2e.yaml
```diff
@@ -0,0 +1,57 @@
...
+        sudo mv kind /bin/
+
+    - name: start_kind_cluster
+      run: |
+        /bin/kind create cluster
+        /bin/kind export kubeconfig
+
+    - name: install_kubectl
+      run: |
+        curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
```
joelanford, line 34: Nit: should we think about pinning this based on the contents of go.mod (ala https://github.com/operator-framework/rukpak/blob/2d3120c7f7c97ab52b58c55f0161707b90566b72/Makefile#L78)

.github/workflows/release-e2e.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+    - name: Set the release related variables
+      run: |
+          if [[ $GITHUB_REF == refs/tags/* ]]; then
+          echo IMAGE_TAG="${GITHUB_REF#refs/tags/}" >> $GITHUB_ENV
+          fi
+
+    - name: release-install
+      run: |
+        kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.7.1/cert-manager.yaml
```
joelanford, line 45: Nit: The cert-manager version seems like another one that we'll forget to keep updated. Is there a way to glean this version from a centralized place in the repo?

.github/workflows/release-e2e.yaml
```diff
@@ -0,0 +1,57 @@
...
+        go-version: '~1.18'
+
+    - name: install_kind_binary
+      env:
+        KIND_VERSION: v0.12.0
+        KIND_OS: kind-linux-amd64
+      run: |
+        curl -sLo kind "https://github.com/kubernetes-sigs/kind/releases/download/${KIND_VERSION}/${KIND_OS}"
+        chmod +x kind
+        sudo mv kind /bin/
```
joelanford, line 25: should we `make kind` and then use `hack/tools/bin/kind`? That way we get the version of kind that the repo specifies?

Right now, we're pretty loose (read: we don't care what it is) with the kind version in the Makefile. Perhaps we should?

.github/workflows/release-e2e.yaml
joelanford: > Should this be on tag pushes rather than release creation?

nvmd, I see that we need the release to exist so that we can use the release `rukpak.yaml` manifest.

timflannagan: Yeah, we can probably drop the worklow_dispatch event trigger here.

nsapse: Will drop. My apologies. That's debugging cruft I forgot to remove so I could trigger it manually for testing.

.github/workflows/release-e2e.yaml
nsapse: Yeah, is is already out of date with the latest release (1.8) so point well taken. I'll look into how to glean.

timflannagan: @tylerslaton Reminded me that we have a Makefile variable for the cert-mgr version that I exported recently:

https://github.com/operator-framework/rukpak/blob/2d3120c7f7c97ab52b58c55f0161707b90566b72/Makefile#L16

I'm not sure how you can source a Makefile environment from a GHA environment outside of using Makefile targets though. Maybe here we could piggy back off the `make cert-mgr` target:

https://github.com/operator-framework/rukpak/blob/2d3120c7f7c97ab52b58c55f0161707b90566b72/Makefile#L109-L111

Which appears to emulate the same apply/wait commands?

.github/workflows/release-e2e.yaml
nsapse: I don't see a `kind` target in the rukpak Makefile.

.github/workflows/release-e2e.yaml
nsapse: Well, that makes it easy. Will switch to piggybacking.

.github/workflows/release-e2e.yaml
exdx: Yes, I think it makes sense to pin a specific version of kind in our Makefile, to make our CI more hermetic

.github/workflows/release-e2e.yaml
exdx: The way core kubernetes components are versioned in go.mod is via minor version updates, that translate to the major versions. So the v0.23.1 tag of the compoenent is actually representing v1.23.1 of kubernetes which is fairly recent (1.24 is currently the head).

.github/workflows/release-e2e.yaml
nsapse: Would it make sense to pin the Makefile to the most recent release? Set the version in the Makefile directly and manually update it? Something else entirely?

timflannagan: This seems like a reasonable request so I opened https://github.com/operator-framework/rukpak/issues/365 which can be done out of band from these changes.

.github/workflows/release-e2e.yaml
nsapse: Huh, is there a straightforward way to do the translation?  After googling for a while I can't find one. I could just `sed` out the leading 0 and replace it with a 1 but that seems like a bit of a hack since presumably at some point K8 will move on to major version 2.

.github/workflows/release-e2e.yaml
nsapse: Thanks, Tim. Assigned that to myself. Might as well do this correctly.

.github/workflows/release-e2e.yaml
joelanford: That makefile shell command specifically looks for a 0 and replaces it with a 1 (it also replaces that patch version with `x`, which is essentially a wildcard as far as the setup-envtest tool is concerned. However it seems like that command only works in the context of a Makefile. I tried copy/pasting into my terminal and it didn't do any substitution.

TL;DR: I think its safe to just sub the major 0 version for a 1 with `s/^v0/v1/`

.github/workflows/release-e2e.yaml
nsapse: Merged in the change - will rebase and use it in this PR.

.github/workflows/release-e2e.yaml
nsapse: Alright, fixed this now that I finished the upstream stuff.

Makefile
```diff
@@ -194,7 +195,7 @@ GINKGO := $(abspath $(TOOLS_BIN_DIR)/ginkgo)
 CONTROLLER_GEN := $(abspath $(TOOLS_BIN_DIR)/controller-gen)
 SETUP_ENVTEST := $(abspath $(TOOLS_BIN_DIR)/setup-envtest)
 GORELEASER := $(abspath $(TOOLS_BIN_DIR)/goreleaser)
-KIND := $(abspath $(TOOLS_BIN_DIR)/KIND)
+KIND := $(abspath $(TOOLS_BIN_DIR)/kind)
```
timflannagan, line 198: Separate commit or move this out into it's own PR?

.github/workflows/release-e2e.yaml
```diff
@@ -0,0 +1,54 @@
...
+          IMAGE_TAG=${GITHUB_REF#refs/tags/} >> $GITHUB_ENV
+          fi
+
+    - name: install_cert_manager
+      run: |
+        make cert-mgr
+
+    - name: release_install
+      run: |
+        kubectl apply -f https://github.com/operator-framework/rukpak/releases/download/${IMAGE_TAG}/rukpak.yaml
```
timflannagan, line 44: I think this might be off as we echo the IMAGE_TAG variable we built up in a previous job to github's environment, which can be indexed from the [`env` context](https://docs.github.com/en/actions/learn-github-actions/contexts#env-context):

```
kubectl apply -f https://github.com/operator-framework/rukpak/releases/download/${{ env.IMAGE_TAG }}/rukpak.yaml

File: .github/workflows/release-e2e.yaml [hunk 11]
```diff
@@ -0,0 +1,54 @@
...
+        kubectl wait --for=condition=Available --namespace=rukpak-system deployment/plain-provisioner --timeout=60s
+        kubectl wait --for=condition=Available --namespace=rukpak-system deployment/core-webhook --timeout=60s
+        kubectl wait --for=condition=Available --namespace=crdvalidator-system deployment/crd-validation-webhook --timeout=60s
+
+    - name: load_bundles
+      run: |
+        make kind-load-bundles
```
timflannagan, line 51: Do we need to override the KIND variable here too?

.github/workflows/release-e2e.yaml
```diff
@@ -0,0 +1,54 @@
...
+
+    - name: install_kubectl
+      run: |
+        curl -LO "https://dl.k8s.io/release/${KUBECTL_VERSION}/bin/linux/amd64/kubectl"
+        sudo install -o root -g root -m 0755 kubectl /usr/local/bin/kubectl
+
+    - name: Set the release related variables
+      run: |
+          if [[ $GITHUB_REF == refs/tags/* ]]; then
+          IMAGE_TAG=${GITHUB_REF#refs/tags/} >> $GITHUB_ENV
```
timflannagan, line 35: nit: identing?

.github/workflows/release-e2e.yaml
```diff
@@ -0,0 +1,54 @@
...
+  release-e2e-kind:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - uses: actions/setup-go@v2
+      with:
+        go-version: '~1.18'
+
+    - name: install_kind_binary
+      env:
```
timflannagan, line 18: nit: remove the empty `env` field here?

timflannagan (commented):
I think we're getting closer. I had a couple of more comments/nits/etc.

.github/workflows/release-e2e.yaml
nsapse: fixed

.github/workflows/release-e2e.yaml
nsapse: fixed

.github/workflows/release-e2e.yaml
```diff
@@ -0,0 +1,54 @@
...
+        go-version: '~1.18'
+
+    - name: install_kind_binary
+      run: |
+        make kind
+
+    - name: start_kind_cluster
+      run: |
+        hack/tools/bin/kind create cluster
+        hack/tools/bin/kind export kubeconfig
```
tylerslaton, line 24: With #384 these two jobs can be condensed to something like:
```suggestion
   - name: start_kind_cluster
      run: make kind-cluster
```

tylerslaton, line 24: With #384 these two jobs can be condensed to something like:
```suggestion
   - name: start_kind_cluster
      run: make kind-cluster
```

tylerslaton: After that though, I'm good with this merging and will approve 👍🏻

nsapse: Integrated your new actions. Also, reviewing that PR the UX improvements are great. Awesome job.

.github/workflows/release-e2e.yaml
```diff
@@ -0,0 +1,49 @@
...
+            IMAGE_TAG=${GITHUB_REF#refs/tags/} >> $GITHUB_ENV
+          fi
+
+    - name: install_cert_manager
+      run: |
+        make cert-mgr
+
+    - name: release_install
+      run: |
+        kubectl apply -f https://github.com/operator-framework/rukpak/releases/download/${{ IMAGE_TAG }}/rukpak.yaml
```
timflannagan, line 38: Same comment as before: I think we need to use {{ env.IMAGE_TAG }} here if we're indexing into the github environment context.

.github/workflows/release-e2e.yaml
```diff
@@ -0,0 +1,49 @@
...
+      run: |
+          if [[ $GITHUB_REF == refs/tags/* ]]; then
+            IMAGE_TAG=${GITHUB_REF#refs/tags/} >> $GITHUB_ENV
+          fi
+
+    - name: install_cert_manager
+      run: |
+        make cert-mgr
+
+    - name: release_install
```
timflannagan, line 36: nit: Install the rukpak release manifests

.github/workflows/release-e2e.yaml
```diff
@@ -0,0 +1,49 @@
...
+        curl -LO "https://dl.k8s.io/release/${KUBECTL_VERSION}/bin/linux/amd64/kubectl"
+        sudo install -o root -g root -m 0755 kubectl /usr/local/bin/kubectl
+
+    - name: Set the release related variables
+      run: |
+          if [[ $GITHUB_REF == refs/tags/* ]]; then
+            IMAGE_TAG=${GITHUB_REF#refs/tags/} >> $GITHUB_ENV
+          fi
+
+    - name: install_cert_manager
```
timflannagan, line 32: nit: Install the cert-manager dependency

.github/workflows/release-e2e.yaml
```diff
@@ -0,0 +1,49 @@
...
+    - uses: actions/checkout@v1
+    - uses: actions/setup-go@v2
+      with:
+        go-version: '~1.18'
+
+    - name: start_kind_cluster
+      run: |
+        make kind-cluster
+
+    - name: install_kubectl
```
timflannagan, line 21: nit: Download kubectl

.github/workflows/release-e2e.yaml
```diff
@@ -0,0 +1,49 @@
...
+jobs:
+  release-e2e-kind:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - uses: actions/setup-go@v2
+      with:
+        go-version: '~1.18'
+
+    - name: start_kind_cluster
```
timflannagan, line 17: nit: Start a kind cluster

Makefile
```diff
@@ -113,7 +114,7 @@ install: generate cert-mgr ## Install rukpak
 run: build-container kind-cluster kind-load install ## Build image, stop/start a local kind cluster, and run operator in that cluster
 
 cert-mgr: ## Install the certification manager
-	kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/$(CERT_MGR_VERSION)/cert-manager.yaml
+	kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/${CERT_MGR_VERSION}/cert-manager.yaml
```
timflannagan, line 117: Mainly curious - any idea on why this change was necessary?

nsapse: I was getting bugs running this locally. I assume it's something related to `$()` being used to launch a subshell but that's a pure guess because I'm a bash noob tbh. I can change it back.

.github/workflows/release-e2e.yaml
```diff
@@ -0,0 +1,49 @@
...
+      with:
+        go-version: '~1.18'
+
+    - name: Start a kind cluster
+      run: |
+        make kind-cluster
+
+    - name: Download kubectl
+      run: |
+        curl -LO "https://dl.k8s.io/release/${{env.KUBECTL_VERSION}}/bin/linux/amd64/kubectl"
```
timflannagan, line 23: I realize we're not using a Makefile target to download kubectl so the `export KUBECTL_VERSION=...` in the Makefile won't propagate to github's environment context. I think we do the following instead:

- Remove the exported variable from the Makefile
- Update this command to inline a shell command: `curl -LO "https://dl.k8s.io/release/$(go list -m k8s.io/kubectl | cut -d" " -f2 | sed 's/^v0/v1/')/bin/linux/amd64/kubectl"`

.github/workflows/release-e2e.yaml
```diff
@@ -0,0 +1,41 @@
+---
+name: release-e2e
```
timflannagan, line 2: Follow-up: update this workflow to automatically create a github issue whenever the job fails, e.g. we fail to successfully install the release, or e2e tests fail, etc.).

Makefile
```diff
@@ -112,6 +112,11 @@ install: generate cert-mgr ## Install rukpak
 
 run: build-container kind-cluster kind-load install ## Build image, stop/start a local kind cluster, and run operator in that cluster
 
+wait:
+	kubectl wait --for=condition=Available --namespace=$(RUKPAK_NAMESPACE) deployment/plain-provisioner --timeout=60s
+	kubectl wait --for=condition=Available --namespace=$(RUKPAK_NAMESPACE) deployment/rukpak-core-webhook --timeout=60s
+	kubectl wait --for=condition=Available --namespace=crdvalidator-system deployment/crd-validation-webhook --timeout=60s
```
timflannagan, line 118: This was added in #392 so we just need to rebase this PR to pick up this new Makefile target.

nsapse: Rebased in.

timflannagan (approved):
Let's get one more approval before merging.

exdx (approved):
LGTM, very nice

timflannagan:
Follow-up(s):

- Automate this ensure process
  - One implementation is by using a custom PAT token during the release workflow
- Ensure a github issue is created when a step in this workflow fails

timflannagan:
@nsapse feel free to merge

PR #327, opened 2022-05-06
docs: add note about spin-up time to README

tylerslaton:
Signed-off-by: Tyler Slaton <tyslaton@redhat.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,8 +36,8 @@ the release notes on how to install a particular release. The only requirement i
 that is configured to target the cluster to install to.
 
 > Note: RukPak depends on [cert-manager](https://cert-manager.io/) for creating and managing certificates for its webhooks.
-> cert-manager should be installed prior to installing RukPak. See the cert-manager [installation docs](https://cert-manager.io/docs/installation/)
-> for more information on how to install cert-manager.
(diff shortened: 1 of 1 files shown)


exdx (commented):
I'm not sure if we want to include this documentation or solve this in-code (#328).

Overall though, if a user is installing from a release, I think we could provide a quick script that ensures the installation is up and ready? This seems like a good candidate for the rukpak CLI tool potentially

exdx (approved):

tylerslaton:
> I'm not sure if we want to include this documentation or solve this in-code (#328).
> 
> Overall though, if a user is installing from a release, I think we could provide a quick script that ensures the installation is up and ready? This seems like a good candidate for the rukpak CLI tool potentially

I agree with you on all of this. I think my preference would be to solve it in code but we can have this as a back-up if we don't feel that the code implementation is quite right.

joelanford:
@exdx I suggested this approach to Tyler because I figured it'd be pretty easy to forget updating that script or instruction if we:
1. added or removed a thing that needed to be waited on
2. changed the name or namespace of something that needed to be waited on.

joelanford (approved):

PR #328, opened 2022-05-06
Add condition checks to install target

exdx:
Add checks to ensure that the deployments bundled with the default
rukpak installations are Available before returning from install. This
prevents users accidentally creating rukpak resources before the
controllers are initialized.

This does hardcode the namespace/name of the deployments, but since we don't currently provide a way to configure those it seems like an acceptable compromise.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,7 @@ VERSION_PATH := $(PKG)/internal/version
 GIT_COMMIT ?= $(shell git rev-parse HEAD)
 PKGS = $(shell go list ./...)
 CERT_MGR_VERSION=v1.7.1
+RUKPAK_NAMESPACE ?= rukpak-system
 
 CONTAINER_RUNTIME ?= docker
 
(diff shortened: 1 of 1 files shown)


akihikokuroda (approved):
LGTM. Thanks!

Makefile
```diff
@@ -99,6 +99,9 @@ kind-cluster: ## Standup a kind cluster for e2e testing usage
 
 install: generate kustomize cert-mgr ## Install rukpak
 	$(KUSTOMIZE) build manifests | kubectl apply -f -
+	kubectl wait --for=condition=Available --namespace=rukpak-system deployment/plain-provisioner --timeout=60s
+	kubectl wait --for=condition=Available --namespace=rukpak-system deployment/rukpak-core-webhook --timeout=60s
+	kubectl wait --for=condition=Available --namespace=crdvalidator-system deployment/crd-validation-webhook --timeout=60s
```
tylerslaton, line 104: Maybe we could make a variable called `NAMESPACE` and then reference it here?

tylerslaton (approved):
Should we edit the release notes to reflect these changes as well? Mainly so users can follow the same steps when officially installing.

Makefile
exdx: That's fair, lets do that

exdx:
> Should we edit the release notes to reflect these changes as well? Mainly so users can follow the same steps when officially installing.

I'm not sure about editing the existing release notes but maybe we can do so moving forward? cc @timflannagan

timflannagan:
Let's just edit the v0.3.0 release notes? Can we double check that the additional wait commands actually work on a cluster before editing that release?

exdx:
> Let's just edit the v0.3.0 release notes? Can we double check that the additional wait commands actually work on a cluster before editing that release?

Sure, I can confirm, and if they work add them in.

timflannagan:
SGTM - thanks!

exdx:
Looks good. I did notice that there is a discrepancy in the name of the core webhook deployment:

* in the v0.3 release it's called `core-webhook`  
* on master it's called `rukpak-core-webhook`

I believe this is a result of the recent kustomize work (#235 specifically) and if it's intentional and we intend on shipping it with that new name in v0.4 we will have to remember to modify that check in the instructions to refer to the new deployment name.

PR #329, opened 2022-05-06
e2e: Accomodate multiple resource does not exist errors

timflannagan:

### DIFF
--- a/test/e2e/plain_provisioner_test.go
+++ b/test/e2e/plain_provisioner_test.go
@@ -845,7 +845,11 @@ var _ = Describe("plain provisioner bundleinstance", func() {
 					WithTransform(func(c *metav1.Condition) metav1.ConditionStatus { return c.Status }, Equal(metav1.ConditionFalse)),
 					WithTransform(func(c *metav1.Condition) string { return c.Reason }, Equal(rukpakv1alpha1.ReasonInstallFailed)),
 					WithTransform(func(c *metav1.Condition) string { return c.Message },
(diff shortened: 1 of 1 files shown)


timflannagan:
Loosely related to #330. See the [conversation-within-a-conversation](https://kubernetes.slack.com/archives/C038B7MF75M/p1651865835897279) in the rukpak-dev channel.

It sounds like this is the wrong approach.

/hold

test/e2e/plain_provisioner_test.go
```diff
@@ -845,7 +845,11 @@ var _ = Describe("plain provisioner bundleinstance", func() {
 					WithTransform(func(c *metav1.Condition) metav1.ConditionStatus { return c.Status }, Equal(metav1.ConditionFalse)),
 					WithTransform(func(c *metav1.Condition) string { return c.Reason }, Equal(rukpakv1alpha1.ReasonInstallFailed)),
 					WithTransform(func(c *metav1.Condition) string { return c.Message },
-						ContainSubstring(`unable to recognize "": no matches for kind "OperatorGroup" in version "operators.coreos.com/v1"`)),
+						Or(
+							ContainSubstring(`unable to recognize "": no matches for kind "OperatorGroup" in version "operators.coreos.com/v1"`),
```
joelanford, line 849: Can you add a comment here explaining the scenario that requires this? I feel like we'll quickly forget why this is necessary, which may lead to cargo-culting or at least losing the thread on a possible solution (e.g. wrap both of these errors in the controller such that our test can look for the wrapper)

timflannagan: Yeah definitely. I'll add a TODO to get back to this tomorrow.

PR #341, opened 2022-05-09
Add warning message regarding BundleInstances

exdx:
Adds a warning message to the BI section indicating that anyone who can create a BI can gain cluster-admin privileges.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,10 @@ spec:
 
 ### BundleInstance
 
+> :warning: A BundleInstance changes the state of the Kubernetes cluster by installing and removing objects. It's important
+> to verify and trust the content that is being installed, and limit access (via RBAC) to the BundleInstance API to only those
+> who require those permissions.
+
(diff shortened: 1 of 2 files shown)


timflannagan (approved):
A couple of comments but nothing blocking:

- This seems geared towards the plain provisioner implementation
- We mentioned auditing Bundles but we don't provide any documentation for how someone can achieve that

exdx:
> A couple of comments but nothing blocking:
> 
>     * This seems geared towards the plain provisioner implementation
> 
>     * We mentioned auditing Bundles but we don't provide any documentation for how someone can achieve that

I think those are good points -- I'm going to move this specific warning to the plain provisioner documentation and add a more generic one in the main README. Regarding auditing, I guess we can be a little hand-wavey for now since there is no mechanism in place (or even an issue tracking one) but this issue is solved in the community, for example https://github.com/Shopify/kubeaudit

PR #347, opened 2022-05-10
internal: Add support for configuring the Bundle content storage directory

timflannagan:
Add support for overriding the default plain provisioner storage directory so developing locally via `go run internal/provisioners/plain/main.go` works.

### DIFF
--- a/internal/provisioner/plain/main.go
+++ b/internal/provisioner/plain/main.go
@@ -57,14 +57,17 @@ func init() {
 }
 
 func main() {
-	var httpBindAddr string
-	var httpExternalAddr string
-	var enableLeaderElection bool
-	var probeAddr string
-	var systemNamespace string
-	var unpackImage string
-	var rukpakVersion bool
-	var gitClientImage string
+	var (
+		httpBindAddr         string
+		httpExternalAddr     string
+		enableLeaderElection bool
+		probeAddr            string
+		systemNamespace      string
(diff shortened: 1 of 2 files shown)


internal/provisioner/plain/main.go
```diff
@@ -74,7 +77,8 @@ func main() {
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
 	flag.BoolVar(&rukpakVersion, "version", false, "Displays rukpak version information")
-	flag.StringVar(&gitClientImage, "git-client-image", "alpine/git:v2.32.0", "Configures which git container image to use to clone bundle git repos")
+	flag.StringVar(&storageDirectory, "storage-dir", storage.DefaultBundleCacheDir, "Configures the directory that is used to store Bundle contents.")
```
joelanford, line 80: Even though we set the default here correctly, do you think it's worthwhile setting `--storage-dir` explicitly in the deployment spec so that if we ever decide to change the volumeMount, its more obvious that we also need to change this flag's value?

timflannagan: Yeah I had already made this change but it looks like it got lost while cherrypicking this change to it's own branch. I'll force push those deployment manifest changes in a second.

tylerslaton (approved):
💯

PR #348, opened 2022-05-10
docs: provisioner specs and usage of contentURL

joelanford:
One of the follow-ups for #311 was to document the new `status.contentURL` field, provide an example for how a client could view the content of a bundle by fetching that URL, and to describe the expectations for provisioner implementations with respect to this field.

This PR:
- updates the plain provisioner doc to include a concrete example of fetching the bundle data
- adds a draft provisioner spec that documents most of the things we (currently) expect a provisioner to do.

### DIFF
--- a/docs/provisioner-spec.md
+++ b/docs/provisioner-spec.md
@@ -1,0 +1,54 @@
+# Provisioner Spec [DRAFT]
+
+## Overview
+
+A provisioner is a controller responsible for reconciling `Bundle` and/or `BundleInstance` objects using
+provisioner-specific logic, but with a consistent API. This provisioner concept is inspirted by and therefore very
+similar to native Kubernetes's `Ingress` API and the ecosystem of ingress controllers.
+
+The idea of a provisioner is to unpack bundle content and install that bundle content onto a cluster, and the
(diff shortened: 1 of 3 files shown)


timflannagan (approved):
The root README.md outlines the provisioner concept. Should we link to this markdown document there as well?

Nice work - I only saw nits that I didn't feel like surfacing as comments. I'm fine with having this be a working doc and committing this draft.

joelanford:
Good catch. I'll add a link from the main readme.

docs/provisioner-spec.md
```diff
@@ -0,0 +1,54 @@
+# Provisioner Spec [DRAFT]
+
+## Overview
```
exdx, line 3: Really nice doc! Maybe we want to mention somewhere some examples of provisioners that we could imagine would exist, or that could we plan on implementing? That may be helpful in demystifying the concept more for folks reading the doc.

exdx (approved):

docs/provisioner-spec.md
joelanford: Let's do that in a follow up. I didn't want to cover any more than I had to to meet what we need for #311.

PR #349, opened 2022-05-10
Use the Updater pattern in the BI controller 

exdx:
Currently the Bundle pattern uses an `Updater` to manage its status updates. A list of status closures are passed in to the Updater, which then evaluates them when the reconciliation loop ends. This is an elegant pattern, and so this PR adds similar functionality to the BI controller, so both controllers have a unified approach to managing their status. 

Note: I attempted to do this via generics and could not get something to work. If we feel like a generic solution is better (to merge the two very similar Bundle and BundleInstance Updaters) I think that it's a good approach to take as well, potentially as a followup?


internal/updater/updater.go
```diff
@@ -64,7 +64,7 @@ func (u *Updater) Apply(ctx context.Context, b *rukpakv1alpha1.Bundle) error {
 	})
 }
 
-func EnsureCondition(condition metav1.Condition) UpdateStatusFunc {
+func EnsureCondition(condition metav1.Condition) UpdateBundleStatusFunc {
```
joelanford, line 67: Is it fair for `BundleStatus` to get to use this function name? :)

In all seriousness, I wonder if we should have `updater/bundle` and `updater/bundleinstance` subpackages such that we don't have to worry about de-conflicting names like this? Or maybe just demote this to `EnsureBundleCondition`? I worry about the length of the function names with the latter option though.

exdx: Yes -- I agree that the naming conventions currently proposed in this PR add a bit of complexity. I can move the updaters into their own subpackages and import them, which hopefully makes things more clear and avoids collisions.

internal/updater/updater.go
```diff
@@ -28,24 +28,24 @@ import (
 	rukpakv1alpha1 "github.com/operator-framework/rukpak/api/v1alpha1"
 )
 
-func New(client client.Client) Updater {
-	return Updater{
+func NewBundleUpdater(client client.Client) BundleUpdater {
```
tylerslaton, line 31: I'm a bit partial to the idea of naming this file `updated_bundle.go` as it matches the `updated_bi.go` format. It also makes it obvious what this is an updater of since these functions are not generically applicable. Thoughts?

exdx: That makes sense, I agree. I'm going to break out the bundle and bundle-instance updaters into their own sub-packages to make this separation even more clear

exdx:
Updated to incorporate the latest feedback

exdx:
Broke out the set of changes into two separate commits so it should be easier to review.

exdx:
Broke out the set of changes into two separate commits so it should be easier to review.

internal/provisioner/plain/controllers/bundleinstance_controller.go
```diff
@@ -246,48 +267,52 @@ func (r *BundleInstanceReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+					Type:    rukpakv1alpha1.TypeInstalled,
+					Status:  metav1.ConditionFalse,
+					Reason:  rukpakv1alpha1.ReasonCreateDynamicWatchFailed,
+					Message: err.Error(),
+				}))
 			return ctrl.Result{}, err
 		}
 
-		u := &unstructured.Unstructured{Object: uMap}
+		uu := &unstructured.Unstructured{Object: uMap}
```
tylerslaton, line 280: nit: `uu` is a little confusing here. Maybe we could rename to something a bit more verbose?

```suggestion
		unstructuredObj := &unstructured.Unstructured{Object: uMap}
```

tylerslaton (approved):
Nice, I like this structure.

internal/provisioner/plain/controllers/bundleinstance_controller.go
exdx: Sounds good

tylerslaton (approved):
👍🏻 Nice cleanup!

PR #350, opened 2022-05-10
Fix lint issues found from golangci-lint run

exdx:
Fixes a duplicate package import and removes a function that is no
longer being used.

These issues were found by simply running `golangci-lint run` locally -- doesn't look like our remote CI didn't pick these up, or at least causes a failure in that case.


timflannagan (approved):

timflannagan:
@exdx I think you're running into this because you're running golangci-lint directly locally vs. `make lint` which should be using the same versioned binary as CI. Maybe we need to add the `tidy` target as a prerequisite if you're running that lint target, and you're still running into skew.

I'm going to manually merge this in the meantime because I realize we can bump the golangci-lint minor version that re-enables some of the linters (that were disabled during the bump to Go 1.18) that enforce these checks.

PR #351, opened 2022-05-10
hack/.github: Bump the golangci-lint tool version

timflannagan:
Update the golangci-lint tool version to v1.46.0 release which re-enables a couple of core linters that were disabled when the golangci-lint had bumped to Go 1.18.


timflannagan:
Follow up to #350

joelanford:
Also bump: https://github.com/operator-framework/rukpak/blob/426ebcf078eaf661ac36abf7e3aacefff0cffb0b/.github/workflows/sanity.yaml#L27

timflannagan:
Nice catch - updated. I was under the impression that we ran `make lint` to fuel that sanity check but I'll hold off on consolidating because that golangci-lint action likely does a bunch of caching under-the-hood.

PR #352, opened 2022-05-11
internal: Ignore unpack Pod already exist errors in the Bundle controller

timflannagan:
The `r.ensureUnpackPod(...)` attempts to create or re-create a Pod that's responsible for sourcing and unpacking Bundle contents. When that method returns an error from the `util.CreateOrRecreate(...)` logic where the generated unpack Pod's metadata.Name is already present on the cluster:

https://github.com/operator-framework/rukpak/blob/d868cf228aecc4bcd1ff1d1a3b32ac4177ca6717/internal/provisioner/plain/controllers/bundle_controller.go#L259

We can imply that this is because we're processing an out-of-date cache resource, and ignore that error as the Bundle's manager instance watches for Pod events:

https://github.com/operator-framework/rukpak/blob/d868cf228aecc4bcd1ff1d1a3b32ac4177ca6717/internal/provisioner/plain/controllers/bundle_controller.go#L437

### DIFF
--- a/internal/provisioner/plain/controllers/bundle_controller.go
+++ b/internal/provisioner/plain/controllers/bundle_controller.go
@@ -30,6 +30,7 @@ import (
 
 	"github.com/nlepage/go-tarfs"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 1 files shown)


joelanford:
Out of curiosity, do you know what sequence of events is causing the bundle reconcile to be called where we end up with the unpack pod being stale? Looking at the `SetupWithManager` function, nothing sticks out to me as something that might result in a bunch of concurrent requests for the same bundle.

My only guess is that we're creating the pod and updating the status, and we're somehow re-reconciling due to the status update before the cache contains the new pod. Is there a chance that we're doing `c.Status().Update()` somewhere before we create the pod?

timflannagan:
I was under the impression that the requeue behavior happens when the Pod has been first created:

https://github.com/operator-framework/rukpak/blob/d868cf228aecc4bcd1ff1d1a3b32ac4177ca6717/internal/provisioner/plain/controllers/bundle_controller.go#L158-L159

We get back that `op == controllerutil.OperationResultCreated` reconcile result, but at that point, I guess the cache should already be up to date? This was only really reproducible for me when I had a bad unpack Pod that was being generated.

joelanford:
> at that point, I guess the cache should already be up to date?

Correct. If the event that triggers the reconcile is the "pod create" event, then the cache will contain the pod object associated with that event.

timflannagan:
Actually I'm not sure this is much of an issue right now. Worst case is we perform unnecessary work requeuing an apierrors.IsAlreadyExists which isn't the end of the world.

PR #358, opened 2022-05-12
Makefile: Surface release related targets in help output

timflannagan:
Move the release/quickstart Makefile targets outside of the hack/tools section, and into their own section. Update both of those targets and add a comment that surfaces when `make help` is run.


tylerslaton (approved):
Would these ever be ran directly by a user? If not, should we be showing them in the `help` target?

timflannagan:
> Would these ever be ran directly by a user? If not, should we be showing them in the help target?

Depends on what user means in this context as we surface dev/CI style targets that I wouldn't expect the general user to interact with.

exdx (approved):
I'm ok either way regarding surfacing this section in `make help`, but I think it's fine as-is

PR #359, opened 2022-05-12
.github: Refactor the sanity workflow

timflannagan:
Update the lint GHA workflow job to use the lint Makefile target to ensure that both local/CI environments use the same entrypoint and version of golangci-lint. Add a step that caches the build and module paths to cut down on the overall runtime when running the lint job.

Separate the step responsible for linting the markdown files in the repository from lint job, into it's own job so that it can run in parallel with the other jobs defined in the sanity workflow.

### DIFF
--- a/.github/workflows/sanity.yaml
+++ b/.github/workflows/sanity.yaml
@@ -35,6 +35,11 @@ jobs:
 
       - name: Run golangci linting checks
         run: make lint
+  markdown:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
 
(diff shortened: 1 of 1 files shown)


timflannagan:
Okay that's definitely a fairly significant performance regression. Looks like some light caching is needed.

/hold

timflannagan:
Looking at https://github.com/golangci/golangci-lint-action#how-to-use, we should be fine to simply cache the build and module paths without sacrificing CI runtimes.

timflannagan:
Looks like caching those build/module paths worked.

/hold cancel

exdx (approved):
LGTM

PR #360, opened 2022-05-12
.github: Update the TODO workflow event triggers

timflannagan:
Restrict the TODO workflow to only run on pull requests and during push events to the main branch.


.github/workflows/todo.yaml
```diff
@@ -1,6 +1,9 @@
 name: todo
 on:
-- push
+  pull_request:
```
joelanford, line 3: Would this cause TODO issues to be created on every PR push that includes a new/different TODO comment?

timflannagan: Hmm maybe the pull_request trigger is the wrong one here. I just wanted to limit this action running on tag events in my fork given our dev model is merging PRs to the main branch right now. I'll update any remove the pull_request field here.

exdx (approved):

PR #357, opened 2022-05-11
Add uninstall target

exdx:
Adds an uninstall target to conveniently remove rukpak from the cluster
entirely. Relies on the kustomize output to build a list of objects to
remove.

<details>

```
rukpak(main): kustomize build manifests | kubectl delete -f -
namespace "crdvalidator-system" deleted
namespace "rukpak-system" deleted
customresourcedefinition.apiextensions.k8s.io "bundleinstances.core.rukpak.io" deleted
customresourcedefinition.apiextensions.k8s.io "bundles.core.rukpak.io" deleted
serviceaccount "crd-validation-webhook" deleted
serviceaccount "plain-provisioner-admin" deleted
serviceaccount "rukpak-core-admin" deleted
clusterrole.rbac.authorization.k8s.io "bundle-reader" deleted
clusterrole.rbac.authorization.k8s.io "crd-validation-webhook" deleted
clusterrole.rbac.authorization.k8s.io "plain-provisioner-admin" deleted
clusterrolebinding.rbac.authorization.k8s.io "crd-validation-webhook" deleted
clusterrolebinding.rbac.authorization.k8s.io "plain-provisioner-admin" deleted
service "crd-validation-webhook" deleted
service "plain-provisioner" deleted
service "rukpak-webhook-service" deleted
deployment.apps "crd-validation-webhook" deleted
deployment.apps "plain-provisioner" deleted
deployment.apps "rukpak-core-webhook" deleted
certificate.cert-manager.io "crd-validation-webhook-certificate" deleted
certificate.cert-manager.io "rukpak-webhook-certificate" deleted
issuer.cert-manager.io "selfsigned" deleted
issuer.cert-manager.io "rukpak-selfsigned" deleted
validatingwebhookconfiguration.admissionregistration.k8s.io "crd-validation-webhook" deleted
validatingwebhookconfiguration.admissionregistration.k8s.io "rukpak-validating-webhook-configuration" deleted
```
</details>

PR #361, opened 2022-05-12
hack,Makefile: Remove kustomize as a direct tool dependency

timflannagan:
Follow-up to #357 which uses kubectl's kustomize command to manage our kustomize manifests.

Refactor any Makefile targets that rely on kustomize with kubectl commands:

- Update the install Makefile command and replace `kustomize build` with `kubectl apply -k manifests`.
- Update the quickstart Makefile command to use `kubectl create --dry-run` and pipe that output to sed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -98,8 +98,8 @@ kind-cluster: ## Standup a kind cluster for e2e testing usage
 
 ##@ install/run:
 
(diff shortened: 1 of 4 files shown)


exdx (approved):
🎉

PR #366, opened 2022-05-13
Git source refactor with go library

### LINKED ISSUE #285: Support private git source bundle

### LINKED ISSUE #291: Refactor sourcing bundles from git implementation to use a git library
Currently the Bundle API supports sourcing content from a git repository, and the plain provisioner is able to clone and persist contents from a bundle git repo to the default storage mechanism (see #179). The existing implementation relies on creating an underlying git-client pod (in this case using an `alpine/git` image) to do the actual cloning and checking out of the git repository on the remote. This approach works, but is difficult to extend to add things like authentication (for sourcing from private repositories) and is not very maintainable long term. 

The alternative approach would be to encode the git specific logic into the provisioner, such that it could checkout the bundle contents from the git repository without relying on an underlying pod to do the unpacking. Some potential git libraries in Go that could be used to accomplish this:
* [go-git](https://github.com/go-git/go-git)
* [git2go](https://github.com/libgit2/git2go)

akihikokuroda:
Closes #291 

This uses `go-git` package for git access.


internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -463,53 +475,151 @@ func bundleImagePod(pod *corev1.Pod, source rukpakv1alpha1.ImageSource, unpackIm
...
+	}
+	var ref string
+	if rev.Branch != "" {
+		ref = rev.Branch
+	} else if rev.Tag != "" {
+		ref = rev.Tag
+	} else {
+		ref = rev.Commit
+	}
+	contentInfo, _, err := client.Contents.List(ctx, repo, path, ref, scm.ListOptions{})
```
joelanford, line 604: The implementation of `client.Contents` for a github-based `scm.Client` seems specific to GitHub's API (see [here](https://github.com/drone/go-scm/blob/766414330d77aa62abbb2e3c951ebdd06a237718/scm/driver/github/content.go#L20))

I have my doubts that this library works with arbitrary git repository URLs.

akihikokuroda: Looks like the way to [create the `client`](https://github.com/drone/go-scm/tree/766414330d77aa62abbb2e3c951ebdd06a237718#getting-started) is different for each type of repository.   Then the drivers for each repository are used via the `client` for each function.

akihikokuroda: The public git and the on-prem git uses the same client.  It can support both by giving the right URL.  The different client is necessary to support Bitbucket, Gitea, Gitlab, etc..

joelanford: Right but again, why are we using a client that needs specific knowledge of the hosted git service? We need to generically support git the protocol, not propietary git services provided by GitHub, Gitlab, Gitea, etc.

If we go with go-scm, it means:
1. We have to do URL matching to figure out which service it is.
2. We have to maintain the mapping between hostname and service (i.e. I could host a gitea instance at git.rukpak.io and put a bundle there. How does this code know that git.rukpak.io is a Gitea instance?)
3. Even if we could feasibly do 2, that would breakdown pretty quickly for things like `localhost` and in-cluster service hostnames like `git.my-git-namespace.svc`
4. We can only ever support the specific SCMs supported by this library (i.e. It could never clone from an nginx-based git server)

akihikokuroda: OK.  I give up go-scm.  Should I close this PR and you take over with your POC?  Or I can refactor with the go-git?

akihikokuroda: @joelanford I refactored using the `go-git` packages.  I took some logic from your POC branch.  Thanks!

internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -464,53 +476,186 @@ func bundleImagePod(pod *corev1.Pod, source rukpakv1alpha1.ImageSource, unpackIm
...
+		return "", "", err
+	}
+	return string(secret.Data["username"]), string(secret.Data["accesstoken"]), nil
+}
 
-	pod = addUnpackerInitContainer(pod, unpackImage)
+func (r *BundleReconciler) handleGitSource(ctx context.Context, u *updater.Updater, bundle *rukpakv1alpha1.Bundle) error {
+	if bundle.Status.Phase == rukpakv1alpha1.PhaseUnpacked {
+		return nil
+	}
```
joelanford, line 505: This is functionally different than the previous code, right?

With the pod-based unpacker, we don't check the phase of the bundle before deciding to proceed with ensuring the unpack pod exists. In that scenario, even if the bundle says it is unpacked, we re-create the unpack pod, re-unpack it, and update the stored bundle (which potentially includes changes)

To get 1-to-1 behavior, is it possible to remove this early exit for now?  If I'm understanding correctly, what will happen if we remove this is:
1. First pass through, we clone the repo, build the bundle tar.gz, store it, and then update `bundle.status` with the details.
2. We run reconciliation again for this bundle because of the status update in 1. This time, we again clone the repo, build the bundle tar.gz, and store it, but the status doesn't change, so this ends the reconciliation for the original change.

We end up with 2 clones, which isn't ideal. But I think we should tackle the question of how to handle that separately. For example, I could ask "what if the status says "Unpacked", but I can see that it isn't present in my storage? I think there are a few questions along those lines that make this somewhat complex.

internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -464,53 +476,186 @@ func bundleImagePod(pod *corev1.Pod, source rukpakv1alpha1.ImageSource, unpackIm
...
+
+	var bundleFS fs.FS = &billyFS{wt.Filesystem}
+
+	// Subdirectory
+	if gitsource.Directory != "" {
+		directory := gitsource.Directory
+		if directory[0] == '.' {
+			directory = directory[2:]
+		}
```
joelanford, line 568: A `.`  as the first character of a directory name doesn't always mean it will be `./`.

Can `filepath.Clean` clean things up for us? After we clean, we probably need to fail if the path starts with `../`.

api/v1alpha1/bundle_types.go
```diff
@@ -81,6 +81,11 @@ type GitSource struct {
 	// is required. Setting more than one field or zero fields will result in an
 	// error.
 	Ref GitRef `json:"ref"`
+	// SecretName contains the secret name that has authorization information for HTTPS protocol and is in the namespace that the provisioner is deployed.
+	// The secret is expected to contain `data.username` and `data.accesstoken` for the username and personal access token, respectively.
+	SecretName string `json:"secretName,omitempty"`
+	// If SslNoVerify is set true, the server certificate is ignored.
+	SslNoVerify bool `json:"sslNoVerify,omitempty"`
```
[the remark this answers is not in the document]
akihikokuroda: OK. I'll split it.  Thanks!

internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -464,53 +476,186 @@ func bundleImagePod(pod *corev1.Pod, source rukpakv1alpha1.ImageSource, unpackIm
...
+}
 
-	pod = addUnpackerInitContainer(pod, unpackImage)
+func (r *BundleReconciler) handleGitSource(ctx context.Context, u *updater.Updater, bundle *rukpakv1alpha1.Bundle) error {
+	if bundle.Status.Phase == rukpakv1alpha1.PhaseUnpacked {
+		return nil
+	}
+	gitsource := bundle.Spec.Source.Git
+	if gitsource.Repository == "" {
+		return updateStatusUnpackFailing(u, fmt.Errorf("missing git information"))
```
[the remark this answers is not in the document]
akihikokuroda: That's correct.  I'll add comments.

internal/provisioner/plain/controllers/bundle_controller.go
joelanford: > After we clean, we probably need to fail if the path starts with ../

In fact, this seems like a thing we should check and fail on in the validation webhook.

internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -464,53 +476,186 @@ func bundleImagePod(pod *corev1.Pod, source rukpakv1alpha1.ImageSource, unpackIm
...
+	fi, err := f.Filesystem.Stat(path)
+	return &billyFile{file, fi, err}, nil
+}
+
+func (f *billyFS) ReadDir(name string) ([]fs.DirEntry, error) {
+	fis, err := f.Filesystem.ReadDir(name)
+	if err != nil {
+		return nil, err
+	}
+	entries := []fs.DirEntry{}
```
joelanford, line 646: Nit: (probably from my own code? :sweat_smile:) 
```suggestion
	entries := make([]fs.DirEntry, 0, len(fis))
```

internal/provisioner/plain/controllers/bundle_controller.go
akihikokuroda: I took out it.  It is working as you describe here now.

internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -75,6 +80,7 @@ type BundleReconciler struct {
 //+kubebuilder:rbac:groups=core.rukpak.io,resources=bundles/finalizers,verbs=update
 //+kubebuilder:rbac:groups=core,resources=pods,verbs=list;watch;create;delete
 //+kubebuilder:rbac:groups=core,resources=pods/log,verbs=get
+//+kubebuilder:rbac:groups=core,resources=secrets,verbs=get
```
joelanford, line 83: This can also be moved to the git+auth follow-up PR, right?

joelanford: As an aside, we might want to maintain this rule separately because we'll only need secret access in the rukpak-system namespace, so we just need a role/rolebinding.

joelanford:
@akihikokuroda We just merged #191, so all of the git sourcing logic has been moved to https://github.com/operator-framework/rukpak/blob/main/internal/source/git.go

Sorry for the conflict that PR created. It should be a fairly simple rebase, but let us know if you have questions!

akihikokuroda:
@joelanford HI!  I rebased.  Please take a look.  Thanks!

internal/source/git.go
```diff
@@ -2,145 +2,91 @@ package source
...
-	op, err := g.ensureUnpackPod(ctx, bundle, pod)
-	if err != nil {
-		return nil, err
-	}
-	if op == controllerutil.OperationResultCreated || op == controllerutil.OperationResultUpdated || pod.DeletionTimestamp != nil {
-		return &Result{State: StatePending}, nil
+	gitsource := bundle.Spec.Source.Git
+	if gitsource.Repository == "" {
+		// This should never happen because the validation webhook rejects git bundles without repository
+		return nil, fmt.Errorf("missing git information")
```
exdx, line 34:
```suggestion
		return nil, errors.New("missing git source information: repository must be provided")
```

internal/source/git.go
```diff
@@ -2,145 +2,91 @@ package source
...
-	case corev1.PodFailed:
-		return nil, g.failedPodResult(ctx, pod)
-	case corev1.PodSucceeded:
-		return g.succeededPodResult(ctx, bundle, pod)
-	default:
-		return nil, g.handleUnexpectedPod(ctx, pod)
+	// Set options for clone
+	cloneOpts := git.CloneOptions{
+		URL:      gitsource.Repository,
+		Progress: &bytes.Buffer{},
```
exdx, line 40: Is this the setting for where the git progress logs should be sent to? Maybe we could pipe them to stdout so users can see those logs? It would be easier for users to diagnose errors that way, at the expense of some potentially noisy logs.

internal/source/git.go
```diff
@@ -2,145 +2,91 @@ package source
...
-		pod.Spec.Containers[0].Command = []string{"/unpack", "--bundle-dir", "/bundle"}
-		pod.Spec.Containers[0].VolumeMounts = []corev1.VolumeMount{{Name: "bundle", MountPath: "/bundle"}}
+	var bundleFS fs.FS = &billyFS{wt.Filesystem}
 
-		pod.Spec.Volumes = []corev1.Volume{
-			{Name: "bundle", VolumeSource: corev1.VolumeSource{EmptyDir: &corev1.EmptyDirVolumeSource{}}},
+	// Subdirectory
+	if gitsource.Directory != "" {
+		directory := filepath.Clean(gitsource.Directory)
+		if directory[:3] == "../" {
```
exdx, line 80: This is a good catch and maybe something maybe we could move to the schema for validation in a follow-up? Based on the `Clean` documentation though, I think it replaces all relative paths so this condition would never be hit? https://pkg.go.dev/path/filepath#Clean

Also, would we want to check for `/` as an invalid directory as well?

exdx (approved):
Nice, this looks a lot better! Just had some small comments. 

Since the e2e tests are passing I think we have confidence this will work.

internal/source/git.go
akihikokuroda: I'll add the check for '/'.  I'll create a follow up for the webhook changes.  Thanks!

internal/source/git.go
akihikokuroda: I'll update.  Thanks!

internal/source/git.go
akihikokuroda: OK.  I'll put the progress into the log.  It's usually a few lines and it doesn't look too bad.  Thanks!

internal/source/git.go
joelanford: My understanding is that `Clean` can't do anything about a path with `../` prefixes. That makes sense because `Clean` doesn't assume CWD when it's working.

In cases where `../` is in the middle of a path, it _can_ clean those up because it can just remove the leading directories from the path.

internal/source/git.go
```diff
@@ -1,146 +1,92 @@
...
-}
 
-func (g *Git) succeededPodResult(ctx context.Context, bundle *rukpakv1alpha1.Bundle, pod *corev1.Pod) (*Result, error) {
-	bundleFS, err := g.getBundleContents(ctx, pod)
-	if err != nil {
-		return nil, fmt.Errorf("get bundle contents: %w", err)
-	}
 	resolvedSource := &rukpakv1alpha1.BundleSource{
 		Type: rukpakv1alpha1.SourceTypeGit,
 		// TODO: improve git source implementation to return result with commit hash.
```
joelanford, line 92: Do we have access to the git commit hash now that we're doing this all with a git library?

I'm curious if we can "resolve" the `bundle.Spec.Source.Git` struct by translating it to one that uses `commit` instead of tag or branch.

internal/provisioner/plain/main.go
```diff
@@ -175,14 +175,7 @@ func main() {
...
 		},
-		rukpakv1alpha1.SourceTypeGit: &source.Git{
-			Client:          mgr.GetClient(),
-			KubeClient:      kubeClient,
-			ProvisionerName: plainBundleProvisionerName,
-			PodNamespace:    ns,
-			UnpackImage:     unpackImage,
-			GitClientImage:  gitClientImage,
-		},
+		rukpakv1alpha1.SourceTypeGit: &source.Git{},
```
exdx, line 178: nit: one other thing I realized is we can remove the git client image `StringVar` from the command line options in line 82 since we no longer rely on an external git image to resolve git references.

internal/source/git.go
akihikokuroda: It seems possible to resolve it to the hash.  `func (*Repository) ResolveRevision (rev plumbing.Revision) (*plumbing.Hash, error)` of `github.com/go-git/go-git/v5/Repository` can translate the tag and branch to the hash then the hash can be used to reset the repo to the commit.

joelanford: I don't think we need to actually reset the repo (unless we have to). I'm just asking if its possible to lookup the hash of the current worktree and then use that in the `status.resolvedSource`.

joelanford: This isn't a blocker, btw. I think we can do this as a follow-up.

internal/provisioner/plain/main.go
akihikokuroda: I missed it.  I'll take it out.  Thanks.

internal/source/git.go
```diff
@@ -1,146 +1,92 @@
...
-	case corev1.PodFailed:
-		return nil, g.failedPodResult(ctx, pod)
-	case corev1.PodSucceeded:
-		return g.succeededPodResult(ctx, bundle, pod)
-	default:
-		return nil, g.handleUnexpectedPod(ctx, pod)
+	// Set options for clone
+	cloneOpts := git.CloneOptions{
+		URL:      gitsource.Repository,
+		Progress: os.Stdout,
```
joelanford, line 40: Also for follow-up: it might be nice to save this off to a `bytes.Buffer` and embed it in an error if the clone fails.

That way we don't pollute the controller logs with the git clone output all the time, but we do get the output when its helpful to have it.

exdx: Yes, I suggested os.Stdout instead of a buffer originally, but only printing it conditionally makes even more sense.

internal/source/git.go
akihikokuroda: > can we use ResolveRevision to populate the bundle.status.resolvedSource?

yes, we can use the hash for all types.

akihikokuroda:
There are 3 follow-ups so far.
1. add check for git directory in the validation webhook
2. improve the provisioner log by putting the git clone progress only on clone error.
3. fill the bundle.status.resolvedSource with the hash value of the commit.

joelanford (approved):

PR #372, opened 2022-05-16
doc: add putting imagePullSecret in the service account

### LINKED ISSUE #363: Pull private bundle images via secrets in the service-account
Rukpak added support for private bundle images in #275 via a `pullSecret` field in the spec of the bundle. If this field is specified, the secret gets referenced in the `spec.ImagePullSecrets` of the underlying unpacker pod that persists the contents of the bundle image. 

It is also possible to pull private images by adding the required secrets to the service account being used by the unpacker pod -- this is the `default` service account in the namespace by default. This way of accessing private container images may be more desirable to users who have setup their namespace secret configuration already. Kubernetes may have also changed the default behavior of service accounts to bind all existing dockerconfig secrets in the namespace. 

We should potentially update the documentation around private images to reference the service account based approach to secret management as well. cc @akihikokuroda

akihikokuroda:

Closes #363

### DIFF
--- a/docs/plain-bundle-spec.md
+++ b/docs/plain-bundle-spec.md
@@ -193,7 +193,9 @@ docker push quay.io/my-registry/rukpak:example
 kubectl create secret docker-registry mysecret --docker-server=quay.io --docker-username="your user name" --docker-password="your password" --docker-email="your e-mail adress" -n rukpak-system
 ```
 
-4. Create a Bundle referencing a private image registry:
+### Use the secret to pull the private image
+
+#### Method 1:  Create a Bundle referencing a private image registry
 
 ```bash
 kubectl apply -f -<<EOF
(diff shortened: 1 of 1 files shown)


akihikokuroda:
I'll update.  Thanks!

docs/plain-bundle-spec.md
```diff
@@ -210,3 +212,9 @@ spec:
   provisionerClassName: core.rukpak.io/plain
 EOF
 ```
+
+#### Method 2: Add the secret to the `imagePullSecrets` in the `default` service account in the privisioner deployed namespace
```
exdx, line 216:
```suggestion
#### Method 2: Add the secret to the `imagePullSecrets` in the `default` service account in the provisioner deployed namespace
```

docs/plain-bundle-spec.md
```diff
@@ -210,3 +212,9 @@ spec:
...
+
+```bash
+kubectl patch serviceaccount default -p '{"imagePullSecrets": [{"name": "myrsecret"}]}' -n rukpak-system
```
exdx, line 219:
```suggestion
kubectl patch serviceaccount default -p '{"imagePullSecrets": [{"name": "mysecret"}]}' -n rukpak-system
```

Does this patch merge the secret with other potential secrets referenced by the SA, or do they get dropped?

exdx (commented):
Thanks for adding this. 

My only concern is that we rely on the `default` service account in the namespace, which is somewhat of an anti-pattern. Using the `default` account is considered bad for security purposes. We don't really have a story here, but may have to address this in the future.

docs/plain-bundle-spec.md
akihikokuroda: Thanks!  I fix it.

docs/plain-bundle-spec.md
akihikokuroda: It's a good question.  This replaces the secrets already in the service account.  I'll add a  note about it.  Thanks!

PR #375, opened 2022-05-16
 Add the kind binary to the hack/tools go module

### LINKED ISSUE #365: Add the kind binary to the hack/tools go module
Update the hack/tools/go.mod and add the github.com/kubernetes-sigs/kind as a tooling dependency. This can avoid any skew between local and CI environments by ensuring both environments use the same version.

AC:

- Update the [hack/tools](https://github.com/operator-framework/rukpak/blob/main/Makefile#L189) section of the Makefile, and add the requisite targets
- Update hack/tools/go.mod and add the kind tool as a dependency

nsapse:
- Closes #365
- Update the hack/tools/go.mod and add the github.com/kubernetes-sigs/kind
as a tooling dependency.
- Creates kind endpoint in Makefile to allow installation for
  testing/hack purposes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ export IMAGE_REPO ?= quay.io/operator-framework/rukpak
 export IMAGE_TAG ?= latest
 IMAGE?=$(IMAGE_REPO):$(IMAGE_TAG)
 KIND_CLUSTER_NAME ?= kind
-KIND := kind
+KIND_OS = kind-linux-amd64
 BIN_DIR := bin
 TESTDATA_DIR := testdata
 VERSION_PATH := $(PKG)/internal/version
@@ -116,7 +116,7 @@ uninstall: ## Remove all rukpak resources from the cluster
 ##################
 # Build and Load #
 ##################
-.PHONY: build plain unpack core build-container kind-load kind-load-bundles kind-cluster
+.PHONY: build plain unpack core build-container kind-load kind-load-bundles kind-cluster kind
 
 ##@ build/load:
 
@@ -188,19 +188,21 @@ TOOLS_BIN_DIR := $(TOOLS_DIR)/bin
 
 ##@ hack/tools:
 
-.PHONY: golangci-lint ginkgo controller-gen goreleaser
+.PHONY: golangci-lint ginkgo controller-gen goreleaser kind
 
 GOLANGCI_LINT := $(abspath $(TOOLS_BIN_DIR)/golangci-lint)
 GINKGO := $(abspath $(TOOLS_BIN_DIR)/ginkgo)
 CONTROLLER_GEN := $(abspath $(TOOLS_BIN_DIR)/controller-gen)
 SETUP_ENVTEST := $(abspath $(TOOLS_BIN_DIR)/setup-envtest)
 GORELEASER := $(abspath $(TOOLS_BIN_DIR)/goreleaser)
+KIND := $(abspath $(TOOLS_BIN_DIR)/KIND)
 
 controller-gen: $(CONTROLLER_GEN) ## Build a local copy of controller-gen
 ginkgo: $(GINKGO) ## Build a local copy of ginkgo
 golangci-lint: $(GOLANGCI_LINT) ## Build a local copy of golangci-lint
 setup-envtest: $(SETUP_ENVTEST) ## Build a local copy of envtest
 goreleaser: $(GORELEASER) ## Builds a local copy of goreleaser
+kind: $(KIND) ## Builds a local copy of kind
 
 $(CONTROLLER_GEN): $(TOOLS_DIR)/go.mod # Build controller-gen from tools folder.
 	cd $(TOOLS_DIR); go build -tags=tools -o $(BIN_DIR)/controller-gen sigs.k8s.io/controller-tools/cmd/controller-gen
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -141,6 +142,11 @@ build-container: BIN_DIR:=$(BIN_DIR)/$(GOOS)
 build-container: build ## Builds provisioner container image locally
 	$(CONTAINER_RUNTIME) build -f Dockerfile -t $(IMAGE) $(BIN_DIR)
 
+kind:
```
tylerslaton, line 145: Testing locally, I believe that you can remove this entirely. The target you defined on 223 should handle this and you just need to call `$(KIND)`. You can also remove line 10 as it isn't needed either. 

Maybe there is some environmental skew that is making it so I'm not experiencing the need for this?

Makefile
```diff
@@ -7,7 +7,8 @@ export IMAGE_REPO ?= quay.io/operator-framework/rukpak
 export IMAGE_TAG ?= latest
 IMAGE?=$(IMAGE_REPO):$(IMAGE_TAG)
 KIND_CLUSTER_NAME ?= kind
-KIND := kind
+KIND := kind-load
```
tylerslaton, line 10: See other comment, but I believe you can delete this line.

tylerslaton (commented):
Nice QOL change. Looks like the verify target in CI is failing due to some skew in the go mod. Are you running `make verify` locally?

Makefile
nsapse: Thanks for the heads up on these two.

tylerslaton (approved):
🚀

go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	github.com/onsi/gomega v1.18.1
 	github.com/operator-framework/api v0.13.0
 	github.com/operator-framework/helm-operator-plugins v0.0.9
-	github.com/spf13/cobra v1.3.0
+	github.com/spf13/cobra v1.4.0
```
exdx, line 13: nit: I wouldn't expect the base go.mod to have a diff if we are only adding a tool, which has a separate hacks/tools/go.mod

Makefile
```diff
@@ -7,7 +7,7 @@ export IMAGE_REPO ?= quay.io/operator-framework/rukpak
 export IMAGE_TAG ?= latest
 IMAGE?=$(IMAGE_REPO):$(IMAGE_TAG)
 KIND_CLUSTER_NAME ?= kind
-KIND := kind
+KIND_OS = kind-linux-amd64
```
timflannagan, line 10: Are we using this KIND_OS variable anywhere?

Makefile
```diff
@@ -116,7 +116,7 @@ uninstall: ## Remove all rukpak resources from the cluster
 ##################
 # Build and Load #
 ##################
-.PHONY: build plain unpack core build-container kind-load kind-load-bundles kind-cluster
+.PHONY: build plain unpack core build-container kind-load kind-load-bundles kind-cluster kind
```
timflannagan, line 119: This looks like a duplicate .PHONY target as we also have one in the hack/tools section

Makefile
```diff
@@ -212,3 +214,5 @@ $(SETUP_ENVTEST): $(TOOLS_DIR)/go.mod # Build setup-envtest from tools folder.
 	cd $(TOOLS_DIR); go build -tags=tools -o $(BIN_DIR)/setup-envtest sigs.k8s.io/controller-runtime/tools/setup-envtest
 $(GORELEASER): $(TOOLS_DIR)/go.mod # Build goreleaser from tools folder.
 	cd $(TOOLS_DIR); go build -tags=tools -o $(BIN_DIR)/goreleaser github.com/goreleaser/goreleaser
+$(KIND): $(TOOLS_DIR)/go.mod
```
timflannagan, line 217: We need to update the hack/tools/go.mod and add the kind dependency too.

exdx (commented):
Looks good. We should be sure that any dependencies are updated in the hack/tools/go.mod

Also, we should add the kind dependency as a blank underscore in the hack/tools/tools.go file first. Then run go mod tidy to ensure that the tools go.mod is up to date. 

See https://github.com/operator-framework/rukpak/pull/217/ for an example of a PR that added a new tool (kustomize)

timflannagan:
It looks like this got merged which isn't the end of the world, but I'm going to re-open that tracking issue as I think we need to update the hack/tools/go.mod also.

timflannagan:
Another thing that we'll need to update is the e2e workflow to use the `e2e` Makefile target.

nsapse:
Oof, my bad for merging this when approved. I'll open a second PR to address these comments.

Makefile
nsapse: We were using it in the previous kind target - I don't think we are anymore. Good point. Will double check then remove.

Makefile
nsapse: Added for redundancy but will remove.

tylerslaton:
> Oof, my bad for merging this when approved. I'll open a second PR to address these comments.

We should probably have a 2 person approval requirement for PR merges. @timflannagan @exdx thoughts?

timflannagan:
I'm probably a -1 on 2 person approval requirements tbh

timflannagan:
I think we can use PR size as a determinant for how many approvals we'll need while we continue iterating. It's not the end of the world that something got prematurely merged.

tylerslaton:
@timflannagan Yeah, that's fine with me.

Makefile
nsapse: adding

Makefile
```diff
@@ -188,19 +188,21 @@ TOOLS_BIN_DIR := $(TOOLS_DIR)/bin
...
 
-.PHONY: golangci-lint ginkgo controller-gen goreleaser
+.PHONY: golangci-lint ginkgo controller-gen goreleaser kind
 
 GOLANGCI_LINT := $(abspath $(TOOLS_BIN_DIR)/golangci-lint)
 GINKGO := $(abspath $(TOOLS_BIN_DIR)/ginkgo)
 CONTROLLER_GEN := $(abspath $(TOOLS_BIN_DIR)/controller-gen)
 SETUP_ENVTEST := $(abspath $(TOOLS_BIN_DIR)/setup-envtest)
 GORELEASER := $(abspath $(TOOLS_BIN_DIR)/goreleaser)
+KIND := $(abspath $(TOOLS_BIN_DIR)/KIND)
```
akihikokuroda, line 198: Should it be `KIND := $(abspath $(TOOLS_BIN_DIR)/kind`?

nsapse: it should - good catch. I modified this in a follow up commit.

PR #376, opened 2022-05-16
Addresses Open Issues with Commit #375

nsapse:
Attempts to fix open comments left on [PR#375](https://github.com/operator-framework/rukpak/pull/375) after merge.

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -17,13 +17,9 @@ jobs:
         go-version: '~1.18'
 
     - name: Install kind binary
-      env:
-        KIND_VERSION: v0.12.0
-        KIND_OS: kind-linux-amd64
       run: |
-        curl -sLo kind "https://github.com/kubernetes-sigs/kind/releases/download/${KIND_VERSION}/${KIND_OS}"
-        chmod +x kind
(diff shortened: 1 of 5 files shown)


timflannagan (commented):
Thanks for the quick follow up - a couple of things:

- Can we drop the unrelated commit that introduces a new GHA workflow?
- Can we update the commit to describe adding kind to the hack/tools module?
- Can we update the e2e workflow and use `kind` Makefile target instead of download the kind binary manually?

timflannagan:
@nsapse Can you add the "Closes #365" to your PR description?

nsapse:
I've re-run the tests twice now and it keeps failing `the bundle name is too long` which seems oddly specific and it's not immediately clear to me why changing the kind binary would modify that.

nsapse:
Looking further the test shows that the code is failing as expected but with a slightly different error message than predicted. Is it possible someone modified the error being returned without updating the tests?

nsapse:
Failed test was fixed by rolling back kind from 0.13.0 -> 0.12.0. Had inadvertently upgraded as part of original pr.

timflannagan (approved):

exdx (approved):

PR #377, opened 2022-05-16
Add unique hash label to dynamic Bundle metadata

### LINKED ISSUE #334: Update the hash-based implementation for the plain provisioner's BI controller
Follow-up to the comment in https://github.com/operator-framework/rukpak/pull/293#discussion_r865173446 which suggested we explore the usage of hashing the BundleTemplate spec.Template field and using that hash value as the generated Bundle's metadata.Name.

exdx:
Adds a unique hash label to the metadata of embedded bundles in the BI spec. The hash label is called `core.rukpak.io/bundle-template-hash` and the value is the fnv32a hash of the bundle template upon creation. 

Closes #334

### DIFF
--- a/internal/provisioner/plain/controllers/bundleinstance_controller.go
+++ b/internal/provisioner/plain/controllers/bundleinstance_controller.go
@@ -343,20 +343,23 @@ func reconcileDesiredBundle(ctx context.Context, c client.Client, bi *rukpakv1al
 	}
 
 	// check whether there's an existing Bundle that matches the desired Bundle template
-	// specified in the BI resource, and if not, generate a new matches the template.
+	// specified in the BI resource, and if not, generate a new Bundle that matches the template.
 	b := util.CheckExistingBundlesMatchesTemplate(existingBundles, bi.Spec.Template)
 	if b == nil {
 		controllerRef := metav1.NewControllerRef(bi, bi.GroupVersionKind())
+		hash := util.GenerateTemplateHash(bi.Spec.Template)
+
 		labels := bi.Spec.Template.Labels
 		if len(labels) == 0 {
 			labels = make(map[string]string)
 		}
 		labels[util.CoreOwnerKindKey] = rukpakv1alpha1.BundleInstanceKind
 		labels[util.CoreOwnerNameKey] = bi.GetName()
+		labels[util.CoreBundleTemplateHashKey] = hash
 
 		b = &rukpakv1alpha1.Bundle{
 			ObjectMeta: metav1.ObjectMeta{
-				Name:            util.GenerateBundleName(bi.GetName()),
+				Name:            util.GenerateBundleName(bi.GetName(), hash),
 				OwnerReferences: []metav1.OwnerReference{*controllerRef},
 				Labels:          labels,
 				Annotations:     bi.Spec.Template.Annotations,
--- a/internal/util/util.go
+++ b/internal/util/util.go
@@ -3,6 +3,7 @@ package util
 import (
 	"context"
 	"fmt"
+	"hash/fnv"
 	"io/ioutil"
 	"sort"
 	"time"
@@ -107,23 +108,34 @@ func CheckExistingBundlesMatchesTemplate(existingBundles *rukpakv1alpha1.BundleL
 }
 
 // CheckDesiredBundleTemplate is responsible for determining whether the existingBundle
-// parameter is semantically equal to the desiredBundle Bundle template.
+// hash is equal to the desiredBundle Bundle template hash.
 func CheckDesiredBundleTemplate(existingBundle *rukpakv1alpha1.Bundle, desiredBundle *rukpakv1alpha1.BundleTemplate) bool {
-	// Create a copy of the existingBundle resource and delete the known set of
-	// auto-generated labels that the BI controller injects when comparing the
-	// current and desired set of labels.
-	bundle := existingBundle.DeepCopy()
-	delete(bundle.Labels, CoreOwnerKindKey)
-	delete(bundle.Labels, CoreOwnerNameKey)
-	return equality.Semantic.DeepEqual(bundle.Spec, desiredBundle.Spec) &&
-		equality.Semantic.DeepEqual(existingBundle.Annotations, desiredBundle.Annotations) &&
-		equality.Semantic.DeepEqual(bundle.Labels, desiredBundle.Labels)
+	if len(existingBundle.Labels) == 0 {
+		// Existing Bundle has no labels set, which should never be the case.
+		// Return false so that the Bundle is forced to be recreated with the expected labels.
+		return false
+	}
+
+	existingHash, ok := existingBundle.Labels[CoreBundleTemplateHashKey]
+	if !ok {
+		// Existing Bundle has no template hash associated with it.
+		// Return false so that the Bundle is forced to be recreated with the template hash label.
+		return false
+	}
+
+	// Check whether the hash of the desired bundle template matches the existing bundle on-cluster.
+	desiredHash := GenerateTemplateHash(desiredBundle)
+	return existingHash == desiredHash
 }
 
-// GenerateBundleName is responsible for generating a unique
-// Bundle metadata.Name using the biName parameter as the base.
-func GenerateBundleName(biName string) string {
-	return fmt.Sprintf("%s-%s", biName, rand.String(5))
+func GenerateTemplateHash(template *rukpakv1alpha1.BundleTemplate) string {
+	hasher := fnv.New32a()
+	DeepHashObject(hasher, template)
+	return rand.SafeEncodeString(fmt.Sprint(hasher.Sum32()))
+}
+
+func GenerateBundleName(biName, hash string) string {
+	return fmt.Sprintf("%s-%s", biName, hash)
 }
 
 // SortBundlesByCreation sorts a BundleList's items by it's
--- a/test/e2e/plain_provisioner_test.go
+++ b/test/e2e/plain_provisioner_test.go
@@ -677,8 +677,9 @@ var _ = Describe("plain provisioner bundleinstance", func() {
 		})
 		It("should generate a Bundle that contains the correct labels", func() {
 			expectedLabels := map[string]string{
-				util.CoreOwnerKindKey: rukpakv1alpha1.BundleInstanceKind,
-				util.CoreOwnerNameKey: bi.GetName(),
+				util.CoreOwnerKindKey:          rukpakv1alpha1.BundleInstanceKind,
+				util.CoreOwnerNameKey:          bi.GetName(),
+				util.CoreBundleTemplateHashKey: "54dd864b69",
 			}
 			Eventually(func() (map[string]string, error) {
 				if err := c.Get(ctx, client.ObjectKeyFromObject(bi), bi); err != nil {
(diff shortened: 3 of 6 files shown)


internal/util/util.go
```diff
@@ -115,15 +117,32 @@ func CheckDesiredBundleTemplate(existingBundle *rukpakv1alpha1.Bundle, desiredBu
...
+		return false
+	}
+
+	// Split up the bundle name into the BI name + hash.
+	// Since bundles are immutable, the name of the bundle is guaranteed to conform to what it was upon creation.
+	hash := strings.Split(bundle.GetName(), "-")[1]
+
+	hasher := fnv.New32a()
+	DeepHashObject(hasher, desiredBundle.Spec)
+	result := rand.SafeEncodeString(fmt.Sprint(hasher.Sum32()))[:10]
```
exdx, line 132: not really happy with the arbitrary indexing here, but otherwise do we run into issues where the size of the hash overflows the limit for the size of a kube object (or even our webhook at 40 chars)?

internal/util/util.go
```diff
@@ -115,15 +117,32 @@ func CheckDesiredBundleTemplate(existingBundle *rukpakv1alpha1.Bundle, desiredBu
...
 	delete(bundle.Labels, CoreOwnerKindKey)
 	delete(bundle.Labels, CoreOwnerNameKey)
-	return equality.Semantic.DeepEqual(bundle.Spec, desiredBundle.Spec) &&
-		equality.Semantic.DeepEqual(existingBundle.Annotations, desiredBundle.Annotations) &&
-		equality.Semantic.DeepEqual(bundle.Labels, desiredBundle.Labels)
+
+	if !equality.Semantic.DeepEqual(existingBundle.Annotations, desiredBundle.Annotations) ||
+		!equality.Semantic.DeepEqual(bundle.Labels, desiredBundle.Labels) {
+		return false
+	}
```
joelanford, line 124: Do we need any of these comparisons anymore? I would assume that we could include `template.Metadata` and `template.Spec` in our hash computation. And then all we need to do is compute the hash and see if it matches the pod's suffix.

exdx: That sounds more reasonable -- thanks.

internal/provisioner/plain/controllers/bundleinstance_controller.go
```diff
@@ -356,7 +356,7 @@ func reconcileDesiredBundle(ctx context.Context, c client.Client, bi *rukpakv1al
 
 		b = &rukpakv1alpha1.Bundle{
 			ObjectMeta: metav1.ObjectMeta{
-				Name:            util.GenerateBundleName(bi.GetName()),
+				Name:            util.GenerateBundleName(bi.GetName(), bi.Spec.Template),
```
joelanford, line 359: Should we just pass `bi` as a single parameter to this function?

internal/util/util.go
```diff
@@ -107,23 +109,35 @@ func CheckExistingBundlesMatchesTemplate(existingBundles *rukpakv1alpha1.BundleL
...
 	// current and desired set of labels.
-	bundle := existingBundle.DeepCopy()
-	delete(bundle.Labels, CoreOwnerKindKey)
-	delete(bundle.Labels, CoreOwnerNameKey)
-	return equality.Semantic.DeepEqual(bundle.Spec, desiredBundle.Spec) &&
-		equality.Semantic.DeepEqual(existingBundle.Annotations, desiredBundle.Annotations) &&
-		equality.Semantic.DeepEqual(bundle.Labels, desiredBundle.Labels)
+	existingBundle = existingBundle.DeepCopy()
+	delete(existingBundle.Labels, CoreOwnerKindKey)
+	delete(existingBundle.Labels, CoreOwnerNameKey)
```
joelanford, line 119: Since we're no longer doing comparisons using `existingBundle`, can we drop these lines as well?

internal/util/util.go
```diff
@@ -107,23 +109,35 @@ func CheckExistingBundlesMatchesTemplate(existingBundles *rukpakv1alpha1.BundleL
...
+
+	// Split up the bundle name to get the template hash.
+	// Since bundles are immutable, the name of the bundle is guaranteed to conform to what it was upon creation.
+	existingHash := strings.Split(existingBundle.GetName(), "-")[1]
```
tylerslaton, line 123: I believe this logic breaks if the `existingBundle` is named something like `super-cool-bundle` since it is grabbing a different part of the name. Maybe we could consider rehashing the `existingBundle` in this function as a solution?

I added this test case into `TestCheckDesiredBundleTemplate` to check that this is actually a problem:

<details>
<summary>TestCheckDesiredBundleTemplate test case</summary>

```
{
	name: "True/BundleMatchesTemplate",
	args: args{
		existingBundle: &rukpakv1alpha1.Bundle{
			ObjectMeta: metav1.ObjectMeta{
				Name: "stub-123-6f74b48d4",
				Labels: map[string]string{
					"stub-123": "stub-123",
				},
			},
			Spec: sampleSpec,
		},
		desiredBundle: &rukpakv1alpha1.BundleTemplate{
			ObjectMeta: metav1.ObjectMeta{
				Name: "stub-123",
				Labels: map[string]string{
					"stub-123": "stub-123",
				},
			},
			Spec: sampleSpec,
		},
	},
	want: true,
}
```

</details>

tylerslaton (commented):
This new hash seems useful to outline through documentation so that a user could know what the hash attached to their Bundle actually is. Should we add that somewhere?

internal/util/util.go
joelanford: I don't think rehashing will work because extra stuff gets added to `bundle.metadata` when it's created (resourceVersion, creationTimestamp, generation, etc)

I think rather than parsing the name of the existing bundle, we can store the computed hash in a label on the bundle when we create it and then look it up there when we need to do this comparison. If you look at a replicaset, you'll see `metadata.labels["pod-template-hash"]` with a value equal to the name suffix. We could name ours `bundle-template-hash`.

exdx: Yes, I realized this was an issue over the weekend -- basically we would have to check the suffix of the name (the last `-`), instead of what was immediately after the first `-`.

As joe pointed out, parsing the name is fairly brittle. I do think that keeping a hash of the template as a label and using that for comparisons is more robust, so I'm leaning towards redoing this towards that implementation instead.

tylerslaton: Sounds good to me! With whatever we implement it would be great to see a test case (like what I linked) come along with it to verify we're not regressing.

exdx:
I've redone the implementation based on a label-based approach -- I think it's what we were looking for. Still seeing some e2e failures that I need to track down further.

exdx:
Also, as an aside: I noticed we don't use the name of the Bundle in the embedded spec at all (even if it's provided) and just use the BI name + hash -- should we sometimes use the name of the embedded Bundle if it's provided in the name instead?

joelanford (approved):
Nice! Once the e2e's are passing, this LGTM

tylerslaton (approved):
Nice, looks great! For the underlying Bundle name not being used, I'm not super sure what it would be used for. That being said, I would expect it to be used somehow if I provide it and then eventually go looking for that generated Bundle as a user.

test/e2e/plain_provisioner_test.go
```diff
@@ -1294,7 +1295,7 @@ var _ = Describe("plain provisioner garbage collection", func() {
 					return false
 				}
 				installedBundleName := bi.Status.InstalledBundleName
-				return installedBundleName != "" && installedBundleName != originalBundleName
+				return installedBundleName != ""
```
timflannagan, line 1298: This check still seems relevant given we're testing that a new Bundle has been created?

exdx: Yes, it's definitely relevant, but the name is now the same between the original bundle and the new bundle, since the template hasn't changed in this test, we just delete/recreate the Bundle. I put these e2e changes in their own commit to highlight. 

Maybe comparing something else here, like resourceVersion, would help indicate that the resources are different even though the name is the same?

exdx: Actually, maybe the UID makes more sense in this context

exdx:
Got the e2e to pass -- now I'm seeing a strange issue where the unit tests are failing on GH Actions but passing locally. Triaging further.

joelanford:
> it feels like we should name the bundle after what's provided

Can we see what happens if you do this with a Deployment that contains a pod template with `.metadata.name` set? I'd be very surprised if that was honored. I'd expect that to fail validation or to be ignored.

internal/provisioner/plain/controllers/bundleinstance_controller.go
```diff
@@ -343,20 +343,23 @@ func reconcileDesiredBundle(ctx context.Context, c client.Client, bi *rukpakv1al
 	}
 
 	// check whether there's an existing Bundle that matches the desired Bundle template
-	// specified in the BI resource, and if not, generate a new matches the template.
+	// specified in the BI resource, and if not, generate a new Bundle that matches the template.
 	b := util.CheckExistingBundlesMatchesTemplate(existingBundles, bi.Spec.Template)
 	if b == nil {
 		controllerRef := metav1.NewControllerRef(bi, bi.GroupVersionKind())
+		hash := util.GenerateTemplateHash(bi.Spec.Template)
```
timflannagan, line 350: Should we add a comment here that explains why we're generating the template hash before building up the set of labels?

internal/util/util.go
```diff
@@ -107,23 +108,32 @@ func CheckExistingBundlesMatchesTemplate(existingBundles *rukpakv1alpha1.BundleL
...
+	}
+
+	var existingHash string
+	if hash, ok := existingBundle.Labels[CoreBundleTemplateHashKey]; ok {
+		existingHash = hash
+	} else {
+		// Existing Bundle has no template hash associated with it.
+		// Return false so that the Bundle is forced to be recreated with the template hash label.
+		return false
+	}
```
timflannagan, line 126: super nit: flip this conditional to !ok so we can exit early and remove the if/else conditional structure?

internal/provisioner/plain/controllers/bundleinstance_controller.go
```diff
@@ -343,20 +343,23 @@ func reconcileDesiredBundle(ctx context.Context, c client.Client, bi *rukpakv1al
...
 			labels = make(map[string]string)
 		}
 		labels[util.CoreOwnerKindKey] = rukpakv1alpha1.BundleInstanceKind
 		labels[util.CoreOwnerNameKey] = bi.GetName()
+		labels[util.CoreBundleTemplateHashKey] = hash
 
 		b = &rukpakv1alpha1.Bundle{
 			ObjectMeta: metav1.ObjectMeta{
-				Name:            util.GenerateBundleName(bi.GetName()),
+				Name:            fmt.Sprintf("%s-%s", bi.GetName(), hash),
```
timflannagan, line 362: I _think_ we can still have a helper function that's responsible for generating the Bundle's metadata.Name but I don't have strong opinions here.

timflannagan (approved):
Nice! lgtm.

internal/provisioner/plain/controllers/bundleinstance_controller.go
exdx: Yes, that sounds good. Basically, my investigation found that that the hash is different on macOS versus linux, most likely due to the representation of the pointer being passed in on different platforms. The FNV hash itself is pretty straightforward and not likely to produce different results, so it must be the underlying data. 

Wherever we use the underlying hash library implementation, we just check for equality later on on-cluster (here and in OLM) which works for that specific os/arch combination. There are no tests that check "given an object, this is its hash" because those tests would seemingly produce different results on different platforms. 

So the end result is I pulled out the hardcoded hashes out of the tests, which watered down the value of the unit tests some, but enable them to pass on different platforms. The alternative would be to make these tests GOOS/GOARCH aware, which seems like overkill, since we have good feedback from the e2e regardless.

Issue #92, opened 2022-03-02
Support referencing and mounting local Bundle content


timflannagan:
Right now the Bundle API only supports referencing a spec.Image container image, which gets pulled down through CRI and unpacked at runtime. It would be nice to be able to support configuring a Bundle to source local cluster contents, e.g. a mounted ConfigMap or PVC filesystem path, so we can avoid having to build/push a scratch-based container image.

Example Bundle YAML manifest outlined in the e2e strawman:

```yaml
kind: Bundle
metadata: 
  name: plumbus-operator.v0.9.3
spec:
  class: deppy.resolveset
  refs:
  - file://content
  volumeMounts:
  - mountPath: /content
    configMap:
      name: local 
      namespace: plumbus
```

We could introduce a spec.Source/spec.Refs field to the Bundle API where we only have a single supported source, which is the image source provider, and build out support for referencing local Kubernetes resource for Bundle content.

timflannagan:
Moving this as 0.1 candidate.

timflannagan:
A couple of things I was thinking about locally when playing around with this yesterday:

- Updating the spec.Image field to be spec.Source and ensuring all source types have a valid prefix, e.g. docker://... or image://... for container images, file://... for anything mountable in k8s, s3://... and whether that's a surprising from a UX perspective.
- How does the provisioner behave when there's updates to the underlying PVC/ConfigMap that's referenced in the spec.Source key? Do we setup something like a fsnotify watcher, or does this come back full circle to the Bundle immutability conversation? My current thinking is that I'd like to be able to declaratively make changes to my Bundle's source-of-truth (e.g. a ConfigMap/PVC in this case) and see the provisioner propagate those changes. Do I need to create another Bundle in order to enable that use case? That seems like it waters down the value proposition but maybe I'm thinking about this in the wrong way.
- How does the provisioner handle flapping between source types? Does this wrap back into the Bundle immutability conversation, or do we need a somewhat first-class way of ensuring Bundles that declare a certain source type _cannot_ be changed after creation?
- It feels like we need validation for when the spec.Source: file://... key is present to ensure that the spec.VolumeMounts configuration is also present for how the e2e strawman outlined in the Bundle API.
- Do we support updating the volume mounts specified for a Bundle?

timflannagan:
[...]

PR #379, opened 2022-05-17
rukpakctl scaffolding

### LINKED ISSUE #369: Add initial scaffolding for rukpakctl
`rukpakctl` is a new CLI tool that interacts with rukpak resources. See [the design doc](https://hackmd.io/AJ8ygfzbTrmXv_CIiUCPtQ?view) for a more detailed document outlining the tool. This issue is for adding the baseline code necessary to add the binary to the project. 

A/C:
* Add new folder and main.go under /cmd
* Bootstrap a cobra CLI tool in the new folder
* Add Makefile targets

akihikokuroda:

Closes #369 

This has just scaffolding for the rukpakctl.  Are there anything to be added in this PR?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -115,13 +115,13 @@ uninstall: ## Remove all rukpak resources from the cluster
 ##################
 # Build and Load #
 ##################
-.PHONY: build plain unpack core build-container kind-load kind-load-bundles kind-cluster
+.PHONY: build plain unpack core rukpakctl build-container kind-load kind-load-bundles kind-cluster
 
 ##@ build/load:
 
 # Binary builds
 VERSION_FLAGS=-ldflags "-X $(VERSION_PATH).GitCommit=$(GIT_COMMIT)"
-build: plain unpack core crdvalidator
+build: plain unpack core crdvalidator rukpakctl
 
 plain:
 	CGO_ENABLED=0 go build $(VERSION_FLAGS) -o $(BIN_DIR)/$@ ./internal/provisioner/plain
@@ -134,6 +134,9 @@ core:
 
 crdvalidator:
 	CGO_ENABLED=0 go build $(VERSION_FLAGS) -o $(BIN_DIR)/$@ ./cmd/crdvalidator
+
+rukpakctl:
+	CGO_ENABLED=0 go build $(VERSION_FLAGS) -o $(BIN_DIR)/$@ ./cmd/rukpakctl
 
 build-container: export GOOS=linux
 build-container: BIN_DIR:=$(BIN_DIR)/$(GOOS)
@@ -215,3 +218,4 @@ $(GORELEASER): $(TOOLS_DIR)/go.mod # Build goreleaser from tools folder.
 	cd $(TOOLS_DIR); go build -tags=tools -o $(BIN_DIR)/goreleaser github.com/goreleaser/goreleaser
 $(KIND): $(TOOLS_DIR)/go.mod
 	cd $(TOOLS_DIR); go build -tags=tools -o $(BIN_DIR)/kind sigs.k8s.io/kind
+
--- a/cmd/rukpakctl/cmd/root.go
+++ b/cmd/rukpakctl/cmd/root.go
@@ -1,0 +1,47 @@
+/*
+Copyright © 2022 NAME HERE <EMAIL ADDRESS>
+
+*/
+package cmd
+
+import (
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+// rootCmd represents the base command when called without any subcommands
+var rootCmd = &cobra.Command{
+	Use:   "rukpakctl",
+	Short: "A brief description of your application",
+	Long: `A longer description that spans multiple lines and likely contains
+examples and usage of using your application. For example:
+
+Cobra is a CLI library for Go that empowers applications.
+This application is a tool to generate the needed files
+to quickly create a Cobra application.`,
+	// Uncomment the following line if your bare application
+	// has an action associated with it:
+	// Run: func(cmd *cobra.Command, args []string) { },
+}
+
+// Execute adds all child commands to the root command and sets flags appropriately.
+// This is called by main.main(). It only needs to happen once to the rootCmd.
+func Execute() {
+	err := rootCmd.Execute()
+	if err != nil {
+		os.Exit(1)
+	}
+}
+
+func init() {
+	// Here you will define your flags and configuration settings.
+	// Cobra supports persistent flags, which, if defined here,
+	// will be global for your application.
+
+	// rootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/.rukpakctl.yaml)")
+
+	// Cobra also supports local flags, which will only run
+	// when this action is called directly.
+	rootCmd.Flags().BoolP("toggle", "t", false, "Help message for toggle")
+}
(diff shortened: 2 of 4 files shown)


exdx (commented):
Thanks for putting together the scaffolding. I know from the [design doc](https://hackmd.io/AJ8ygfzbTrmXv_CIiUCPtQ) we had quite a number of possible commands, but I think it would make sense to just add one at a time. 

The two commands that seem the most plausible were one that unpacks the tar.gz bundle from the volume and prints the contents as a YAML stream, and a command that could create a BI on cluster from a local set of manifests (this probably requires #92 though). Since that's the case, can we simply drop all the other commands for now and just introduce the `unpack` command for now? It would effectively do what's described in the [plain provisioner script](https://github.com/operator-framework/rukpak/blob/main/internal/provisioner/plain/README.md#make-bundle-content-available-but-do-not-install-it). It might make more sense to bring in the unpack command as a follow-up though. 

My thinking with this PR is we can have just a minimal stub for the binary, so we can develop commands iteratively afterwards. Just a main.go and some Makefile targets with the basic directory structure is enough.

Makefile
```diff
@@ -216,3 +221,5 @@ $(GORELEASER): $(TOOLS_DIR)/go.mod # Build goreleaser from tools folder.
 	cd $(TOOLS_DIR); go build -tags=tools -o $(BIN_DIR)/goreleaser github.com/goreleaser/goreleaser
 $(KIND): $(TOOLS_DIR)/go.mod
 	cd $(TOOLS_DIR); go build -tags=tools -o $(BIN_DIR)/kind sigs.k8s.io/kind
+$(COBRA_CLI): $(TOOLS_DIR)/go.mod
```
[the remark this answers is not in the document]
akihikokuroda: OK.  I'll take it out.

cmd/rukpakctl/cmd/root.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func init() {
+	cobra.OnInitialize(initConfig)
+
+	// Here you will define your flags and configuration settings.
+	// Cobra supports persistent flags, which, if defined here,
+	// will be global for your application.
+
+	rootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/.rukpakctl.yaml)")
```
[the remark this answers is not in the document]
akihikokuroda: OK.  I'll take this out, too.

akihikokuroda:
Thanks for review.  I'll take out unnecessary pieces for now.  I'll just put the `unpack` command stub.

cmd/rukpakctl/main.go
```diff
@@ -0,0 +1,11 @@
+/*
+Copyright © 2022 NAME HERE <EMAIL ADDRESS>
+
+*/
```
timflannagan, line 4: nit: we can drop this copyright header?

cmd/rukpakctl/cmd/root.go
```diff
@@ -0,0 +1,47 @@
...
+package cmd
+
+import (
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+// rootCmd represents the base command when called without any subcommands
+var rootCmd = &cobra.Command{
```
timflannagan, line 14: I think my preference would be to have a constructor that returns a *cobra.Command:

```go
func NewRootCmd() {
    return &cobra.Command {
        ...
    }
```

cmd/rukpakctl/cmd/root.go
```diff
@@ -0,0 +1,47 @@
...
+	// has an action associated with it:
+	// Run: func(cmd *cobra.Command, args []string) { },
+}
+
+// Execute adds all child commands to the root command and sets flags appropriately.
+// This is called by main.main(). It only needs to happen once to the rootCmd.
+func Execute() {
+	err := rootCmd.Execute()
+	if err != nil {
+		os.Exit(1)
```
timflannagan, line 33:
```suggestion
                fmt.Println(os.Stderr, err)
		os.Exit(1)
```

cmd/rukpakctl/cmd/root.go
```diff
@@ -0,0 +1,47 @@
...
+// rootCmd represents the base command when called without any subcommands
+var rootCmd = &cobra.Command{
+	Use:   "rukpakctl",
+	Short: "A brief description of your application",
+	Long: `A longer description that spans multiple lines and likely contains
+examples and usage of using your application. For example:
+
+Cobra is a CLI library for Go that empowers applications.
+This application is a tool to generate the needed files
+to quickly create a Cobra application.`,
```
timflannagan, line 22: nit: Can we populate these short/long descriptions?

cmd/rukpakctl/main.go
```diff
@@ -0,0 +1,11 @@
...
+package main
+
+import "github.com/operator-framework/rukpak/cmd/rukpakctl/cmd"
+
+func main() {
+	cmd.Execute()
```
timflannagan, line 10: If we refactor the root cobra.Command into a constructor function vs. relying on a package-level variable, we can do the following:

```suggestion
	cmd := root.NewCmd()
	if err := cmd.Execute(); err != nil {
	    fmt.Fprintln(os.Stderr, err)
	    os.Exit(1)
        }
```

cmd/rukpakctl/cmd/root.go
```diff
@@ -0,0 +1,47 @@
+/*
+Copyright © 2022 NAME HERE <EMAIL ADDRESS>
+
+*/
+package cmd
```
timflannagan, line 5:
```suggestion
package root
```

cmd/rukpakctl/cmd/unpack.go
```diff
@@ -0,0 +1,40 @@
...
+package cmd
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+// unpackCmd represents the unpack command
+var unpackCmd = &cobra.Command{
```
timflannagan, line 14: I think we can either drop this unpack command, or apply the same feedback where we can use a constructor that returns a *cobra.Command, which the root.NewCmd constructor can register:

```go
package root

import (
    ...
)

func NewCmd() *cobra.Command {
    cmd := &cobra.Command{
        ...
    }

    cmd.AddCommand(unpack.NewCmd())
    ...
    return cmd
}
```

timflannagan (commented):
Thanks for taking a stab at this - I had a couple of comments, but I think we can use a constructor function that returns built up cobra.Commands vs. relying on package-level variable declarations and `init` functions. Thoughts?

cmd/rukpakctl/cmd/root.go
timflannagan: Actually: what we can do here is move this root.go into it's own namespace, e.g. `package root` so the cmd/rukpakctl/main.go can call it via `root.NewCmd()`.

akihikokuroda:
@timflannagan Thanks for review.  The contents in the `rukpakctl` are generated by 
```
cobra-cli init
cobra-cli add unpack
```
[as-is](https://github.com/spf13/cobra-cli/blob/main/README.md).  I thought it was intention of [this issue](https://github.com/operator-framework/rukpak/issues/369).  I can take your comments and make changes if it should be done in this issue.  WDYT?

timflannagan:
I'm fine with a follow up here. It looks like Dan already approved so feel free to merge.

PR #380, opened 2022-05-18
BI reconciliation refactor

### LINKED ISSUE #283: Refactor the main reconciliation logic for the plain provisioner's BI controller
Goal: refactor the main Reconcile method for the BI controller into individual methods to help reduce the amount of complexity (and overall LoC) for the driver function.

Right now the plain provisioner's main Reconcile method is reaching a complexity threshold which makes it difficult to read, reason about, and extend when interacting with that part of the codebase.

The controller is responsible for the following control flow:

- Get the BI resource from the informer cache
- Check whether the corresponding Bundle resource is reporting an Unpacked phase
- Get the unpacked Bundle contents from the storage layer's cache
- Build up a helm chart based on the desired set of client.Object resources
- Ensure the release state (e.g. install or upgrade) logic is performed
- Ensure the desired objects have a dynamic watch present in the internal mapping of GVK watches

As the BI API surface gets extended to support things like label selection (#94) or embedded Bundles (#73), we'll need to introduce more complexity into this controller.

akihikokuroda:

Closes #283

### DIFF
--- a/internal/provisioner/plain/controllers/bundledeployment_controller.go
+++ b/internal/provisioner/plain/controllers/bundledeployment_controller.go
@@ -93,6 +93,7 @@ func (r *BundleDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	l.V(1).Info("starting reconciliation")
 	defer l.V(1).Info("ending reconciliation")
 
+	// Get the BD resource from the informer cache
 	bd := &rukpakv1alpha1.BundleDeployment{}
 	if err := r.Get(ctx, req.NamespacedName, bd); err != nil {
 		return ctrl.Result{}, client.IgnoreNotFound(err)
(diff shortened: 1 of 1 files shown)


akihikokuroda:
@timflannagan Thanks for your review.  I'll take your suggestions and see how they make the code look better.

tylerslaton (commented):
Thanks for the contribution! I agree that there are some updates that could be made to make this controller's logic cleaner. If you rebase this PR you'll notice some changes from to the `bundle_controller.go` file. 

I believe that we should move towards updating this controller to be more inline with the patterns on display in the `bundle_controller.go` file. Namely around this [updateStatusX()](https://github.com/operator-framework/rukpak/blob/main/internal/provisioner/plain/controllers/bundle_controller.go#L174-L186) style of functions that solely manage updating the status.

akihikokuroda:
@tylerslaton I'll work on UpdateStatusX pattern.  Thanks!

internal/provisioner/plain/controllers/bundleinstance_controller.go
```diff
@@ -164,9 +155,68 @@ func (r *BundleInstanceReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+				Type:    rukpakv1alpha1.TypeHasValidBundle,
+				Status:  status,
+				Reason:  reason,
+				Message: message,
+			}),
+		)
+		return false, ctrl.Result{}, nil
+	}
+	return true, ctrl.Result{}, nil
+}
```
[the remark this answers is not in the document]
akihikokuroda: Thanks!  I fix this.

internal/provisioner/plain/controllers/bundledeployment_controller.go
```diff
@@ -134,39 +183,14 @@ func (r *BundleDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
-		}))
-
-	desiredObjects, err := r.loadBundle(ctx, bundle, bd.GetName())
-	if err != nil {
-		u.UpdateStatus(
-			updater.EnsureCondition(metav1.Condition{
-				Type:    rukpakv1alpha1.TypeHasValidBundle,
-				Status:  metav1.ConditionFalse,
-				Reason:  rukpakv1alpha1.ReasonBundleLoadFailed,
-				Message: err.Error(),
```
exdx, line 162: We have pretty granular conditions here in case of errors, which I think is very helpful for the user. By refactoring to simply return an error in the reconciliation loop, we lose a lot of information. Could we refactor but keep these informative conditions?

akihikokuroda: Thanks for review!  It seems that `git diff` got confused.  The new code is like [this](https://github.com/operator-framework/rukpak/blob/c010d2ba7aaa42afe72bff30cf30d0b7c6fffe82/internal/provisioner/plain/controllers/bundledeployment_controller.go#L133).  It is returning the err from the loadBundle as a return value of the reconciler.
```
        // Get the unpacked Bundle contents from the storage layer's cache
        desiredObjects, err := r.loadBundle(ctx, bundle, bd.GetName())
        if err != nil {
                updateStatusHasValidBundle(&u, metav1.ConditionFalse, rukpakv1alpha1.ReasonBundleLoadFailed, err.Error())
                return ctrl.Result{}, err
        }
``` 
The `updateStatusHasValidBundle` set the same status.

akihikokuroda:
I close this for now.  I can work for the issue later.

PR #384, opened 2022-05-19
feat(makefile): update installation UX for Kind to be more seamless

tylerslaton:
Signed-off-by: Tyler Slaton <tyslaton@redhat.com>

## Summary 
Updating the UX of utilizing `kind` locally to be more seamless. 

The UX goes from:

```
make kind
make kind-cluster
make run
```

to

```
make kind-cluster
make run
```

Additionally, you can change which binary you're using (your own personal, the repo's, etc) by specifying the KIND variable.

```
make kind-cluster KIND=/usr/local/bin/kind
make run
```

Lastly, this PR also updates the `e2e` CI job to benefit from this change.

## Notes
A follow-up PR will be responsible for updating the docs to reflect this, found in #373.

PR #385, opened 2022-05-19
Makefile: automatically build and use hack/tools/bin/kind in kind-related targets

joelanford:


timflannagan (approved):
It looks like there's also some overlap here between this and #385?

joelanford:
Sweet, yeah I'll close in favor of #384.

Issue #251, opened 2022-04-14
Installing the Combo Operator documentation references broken images


oceanc80:
The [Installing the Combo Operator ](https://github.com/operator-framework/rukpak/tree/main/internal/provisioner/plain#installing-the-combo-operator) instructions reference broken images. Both `quay.io/tflannag/bundle:combo-operator-v0.0.1` and `quay.io/tflannag/bundle:combo-operator-v0.0.2` need to be updated to the new flattened `manifests` directory structure as the existing images both fail to be unpacked due to the nested `crd` directory.

# Acceptance Criteria
- [ ] Ensure documentation steps work as expected

PR #386, opened 2022-05-19
Fix the Combo Operator Installation docs

oceanc80:
The [Installing the Combo Operator](https://github.com/operator-framework/rukpak/tree/main/internal/provisioner/plain#installing-the-combo-operator) instructions referenced broken images. This PR updates those references to the correct images.

Closes: https://github.com/operator-framework/rukpak/issues/251

Cleaner version of [this PR](https://github.com/operator-framework/rukpak/pull/373).

### DIFF
--- a/internal/provisioner/plain/README.md
+++ b/internal/provisioner/plain/README.md
@@ -103,7 +103,8 @@ RBAC permissions to access bundle content.
 
 As an example, a client outside the cluster can view the file contents from a bundle named `my-bundle` by running
 the following script:
-```console
+
+```bash
 BUNDLE_NAME=my-bundle
 
 kubectl create sa fetch-bundle -n default
@@ -134,24 +135,29 @@ resources referenced by the bundle are present on the cluster.
 
 ### Setup
 
-To experiment with the `plain` provisioner locally, first setup a local cluster, or simply
-have [kind](https://kind.sigs.k8s.io/) installed locally.
+To experiment with the `plain` provisioner locally, take the following steps to
+create a local [kind](https://kind.sigs.k8s.io/) cluster and deploy the provisioner onto it:
 
-Once the cluster has been setup, take the following steps:
+```bash
+# Clone the repository
+git clone https://github.com/operator-framework/rukpak
 
-* Clone the repository via `git clone https://github.com/operator-framework/rukpak`
-* Navigate to the repository via `cd rukpak`
-* Run `make run` to build and deploy the provisioner onto the local cluster.
+# Navigate to the repository
+cd rukpak
+
+# Start a local kind cluster then build and deploy the provisioner onto it
+make run
+```
 
 ### Installing the Combo Operator
 
-From there, create some `Bundles` and `BundleInstance` types to see the provisioner in action. For an example bundle to
+From there, create some Bundles and BundleInstance types to see the provisioner in action. For an example bundle to
 use, the [combo operator](https://github.com/operator-framework/combo) is a good candidate.
 
-Create the combo `BundleInstance` referencing the desired combo `Bundle` configuration.
+Create the combo BundleInstance referencing the desired combo Bundle configuration:
 
-```console
-$ kubectl apply -f -<<EOF
+```bash
+kubectl apply -f -<<EOF
 apiVersion: core.rukpak.io/v1alpha1
 kind: BundleInstance
 metadata:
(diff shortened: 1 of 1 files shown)


internal/provisioner/plain/README.md
```diff
@@ -235,51 +279,58 @@ spec:
...
 
 ```console
+# Delete the combo BundleInstance
+kubectl delete bundleinstances.core.rukpak.io combo
+```
+
+A message should show that the `BundleInstance` was deleted and now the cluster state is the same as it was
+prior to installing the operator.
+
+```text
```
timflannagan, line 326:
```suggestion
```console
```

internal/provisioner/plain/README.md
```diff
@@ -235,51 +279,58 @@ spec:
...
 ```
 
-Notice that the container image has changed since we had first installed the combo operator.
+Notice that the container image has changed to `v0.0.2` since we first installed the combo operator.
 
-### Deleting the Combo Operator
+### Deleting the Combo Operator and Local Kind Cluster
 
-To cleanup from the installation, simply remove the BundleInstance from the cluster. This will remove all references
+To clean up from the installation, simply remove the `BundleInstance` from the cluster. This will remove all references
```
timflannagan, line 313: nit: do we introduce inconsistency using `BundleInstance` vs. BundleInstance?

internal/provisioner/plain/README.md
```diff
@@ -134,24 +135,29 @@ resources referenced by the bundle are present on the cluster.
 
 ### Setup
 
-To experiment with the `plain` provisioner locally, first setup a local cluster, or simply
-have [kind](https://kind.sigs.k8s.io/) installed locally.
+To experiment with the `plain` provisioner locally, take the following steps to
+create a local kind cluster and deploy the provisioner onto it:
```
timflannagan, line 139: It looks like we dropped the link to the kind documentation here. Is that still worth linking out to that documentation?

internal/provisioner/plain/README.md
```diff
@@ -166,50 +172,88 @@ spec:
...
 $ kubectl -n combo delete deployments.apps combo-operator
 deployment.apps "combo-operator" deleted
 ```
 
-Check for the deployment again, it will be back on the cluster. The provisioner ensures that all resources required for
-the BundleInstance to run are accounted for on-cluster.
+The provisioner ensures that all resources required for the BundleInstance to run are accounted for on-cluster.
+So if we check for the deployment again, it will be back on the cluster:
+
+```text
```
timflannagan, line 252:
```suggestion
```bash
```

internal/provisioner/plain/README.md
```diff
@@ -235,51 +279,58 @@ spec:
...
 EOF
 ```
 
-And wait until the newly generated `combo-xzfxv` Bundle is reporting an Unpacked status:
+Once the newly generated `combo-xzfxv` Bundle is reporting an Unpacked status, the BundleInstance `combo`
+resource should now point to the new Bundle version. The combo-operator deployment
+in the combo namespace should also be healthy and contain a new container image:
 
-```console
+```text
```
timflannagan, line 291:
```suggestion
```bash
```

timflannagan (approved):

internal/provisioner/plain/README.md
oceanc80: I'll have to check but I thought I made that change to make it consistent in that section of the doc.

internal/provisioner/plain/README.md
oceanc80: I had removed it because we automatically use kind, but it might be nice to leave the link in so that people can read more about it if they're not familiar.

internal/provisioner/plain/README.md
```diff
@@ -166,50 +172,88 @@ spec:
...
 deployment.apps "combo-operator" deleted
 ```
 
-Check for the deployment again, it will be back on the cluster. The provisioner ensures that all resources required for
-the BundleInstance to run are accounted for on-cluster.
+The provisioner ensures that all resources required for the BundleInstance to run are accounted for on-cluster.
+So if we check for the deployment again, it will be back on the cluster:
+
+```text
+$ kubectl -n combo get deployments.apps combo-operator
```
tylerslaton, line 253: nit: Should probably decide if we want to have the `$` on a lot of these as it seems like it is inconsistent. 
```suggestion
kubectl -n combo get deployments.apps combo-operator
```

tylerslaton (approved):
Nice, looks good to me.

internal/provisioner/plain/README.md
oceanc80: It looks like it's kind of inconsistent throughout the doc. Is there a preference for one way over the other?
cc: @exdx

timflannagan: I'm fine with whatever as long as we're consistent. I think we typically use BundleInstance without the code brackets throughout the rest of the documentation set, so it's also possible we have this issue outside of this doc as well.

internal/provisioner/plain/README.md
oceanc80: I tried to split up the commands and the console output so the user can use the copy function in github to quickly grab the command. I've also tried to make the color coding consistent so it's more obvious which is which, but I didn't go back and make it consistent across the whole doc so I'm going to do another pass to just make sure it all flows.

internal/provisioner/plain/README.md
```diff
@@ -166,50 +172,93 @@ spec:
...
+```bash
+# Check the combo operator deployment
+kubectl -n combo get deployments.apps combo-operator
+
+# Check that the operator is present
+kubectl -n combo get deployments.apps combo-operator -o yaml | grep 'image:' | xargs
+```
+
+The deployment should show that the operator is ready and available and the operator
+should be present:
```
exdx, line 228:
```suggestion
The deployment should show that the operator is ready and available:
```
nit: the second half of the sentence is repetitive

exdx (approved):
Looks good! Just some small nits.

PR #387, opened 2022-05-23
POC: registry+v1 bundle provisioner

joelanford:
This PR builds on the source interface PR to add a new registry+v1 bundle provisioner that converts a registry+v1 bundle to a plain+v1 bundle prior to storing the bundle in the provisioner storage.

There are lots of small pieces of this PR that can be broken out, where I discovered problems that weren't apparent without another provisioner in the mix.

Here's an example of a BundleInstance that you can install with this PR:
```yaml
kind: BundleInstance
metadata:
  name: prometheus-operator
spec:
  template:
    spec:
      provisionerClassName: core.rukpak.io/registry
      source:
        type: image
        image:
          ref: quay.io/operatorhubio/prometheus@sha256:5b04c49d8d3eff6a338b56ec90bdf491d501fe301c9cdfb740e5bff6769a21ed
```


internal/convert/registryv1.go
```diff
@@ -265,18 +370,111 @@ func generateName(base string, o interface{}) string {
...
+		Subjects: subjects,
+		RoleRef: rbacv1.RoleRef{
+			APIGroup: rbacv1.GroupName,
+			Kind:     "ClusterRole",
+			Name:     roleName,
+		},
+	}
+}
+
+func keys[Key comparable, Value any](m map[Key]Value) []Key {
```
timflannagan, line 466: :eyes:

internal/provisioner/registry/controllers/bundle_controller.go
```diff
@@ -0,0 +1,285 @@
...
+			Type:    rukpakv1alpha1.TypeUnpacked,
+			Status:  metav1.ConditionFalse,
+			Reason:  rukpakv1alpha1.ReasonUnpackFailed,
+			Message: err.Error(),
+		}),
+	)
+	return err
+}
+
+func getObjects(bundleFS fs.FS) ([]client.Object, error) {
```
timflannagan, line 238: maybe we could provide a closure here for how to handle building up the output array of client.Object(s) here if we're using this logic in multiple provisioner implementations?

internal/provisioner/registry/types/types.go
```diff
@@ -0,0 +1,6 @@
+package registry
+
+const (
+	// ProvisionerID is the unique plain provisioner ID
```
timflannagan, line 4: nit: s/plain/registry

internal/util/util.go
```diff
@@ -37,6 +41,31 @@ func BundleInstanceProvisionerFilter(provisionerClassName string) predicate.Pred
...
 
+func MapOwneeToBundle(ctx context.Context, cl client.Client, log logr.Logger, provisionerID string) handler.MapFunc {
+	return func(obj client.Object) []reconcile.Request {
+		var bundleKey *types.NamespacedName
+		for _, ref := range obj.GetOwnerReferences() {
+			if ref.Kind == rukpakv1alpha1.BundleKind &&
+				ref.APIVersion == rukpakv1alpha1.GroupVersion.String() &&
+				ref.Controller != nil && *ref.Controller {
+				bundleKey = &types.NamespacedName{Name: ref.Name}
+			}
```
timflannagan, line 52: helper function?

manifests/provisioners/registry/resources/deployment.yaml
```diff
@@ -0,0 +1,49 @@
...
+          ports:
+            - containerPort: 8443
+              protocol: TCP
+              name: https
+        - name: manager
+          image: quay.io/operator-framework/rukpak:latest
+          imagePullPolicy: IfNotPresent
+          command: ["/registry"]
+          args:
+            - "--unpack-image=quay.io/operator-framework/rukpak:latest"
```
timflannagan, line 38: :+1:

internal/storage/localdir.go
```diff
@@ -25,15 +28,36 @@ const DefaultBundleCacheDir = "/var/cache/bundles"
...
 
-func (s *LocalDirectory) Load(_ context.Context, owner client.Object) (fs.FS, error) {
-	bundleFile, err := os.Open(s.bundlePath(owner.GetName()))
+func (s *LocalDirectory) Load(ctx context.Context, owner client.Object) (fs.FS, error) {
+	bundle := owner.(*rukpakv1alpha1.Bundle)
+
+	// FIXME(TODO): use a split storage? i.e. if the bundle is a plain bundle,
+	//  we know its "ours" and have it stored locally, so we can just load from
+	//  disk. If it isn't ours, we need to pull from its content URL. For now,
+	//  always pulling from the content URL should be fine.
```
timflannagan, line 40: How bad of an idea would it be to first check local storage by stating the expected tar.gz file, and then fall back to network requests when we get a cache miss?

internal/provisioner/registry/main.go
```diff
@@ -0,0 +1,214 @@
...
+		LeaderElectionID:       "510f803c.olm.operatorframework.io",
+		NewCache: cache.BuilderWithOptions(cache.Options{
+			SelectorsByObject: cache.SelectorsByObject{
+				&rukpakv1alpha1.BundleInstance{}: {},
+				&rukpakv1alpha1.Bundle{}:         {},
+			},
+			DefaultSelector: cache.ObjectSelector{
+				Label: dependentSelector,
+			},
+		}),
```
timflannagan, line 123: Do we need a helper function that's responsible building up a custom cache when we start investing in a shared library approach?

internal/convert/registryv1.go
```diff
@@ -1,34 +1,146 @@
...
+	reg := RegistryV1{}
+	fileData, err := fs.ReadFile(rv1, filepath.Join("metadata", "annotations.yaml"))
+	if err != nil {
+		return nil, err
+	}
+	annotationsFile := registry.AnnotationsFile{}
+	if err := yaml.Unmarshal(fileData, &annotationsFile); err != nil {
+		return nil, err
+	}
+	reg.PackageName = annotationsFile.Annotations.PackageName
```
timflannagan, line 54: do we need to nil check this to avoid a panic here?

.goreleaser.template.yml
```diff
@@ -115,6 +115,7 @@ release:
     kubectl wait --for=condition=Available --namespace=cert-manager deployment/cert-manager-webhook --timeout=60s
     kubectl apply -f https://github.com/operator-framework/rukpak/releases/download/{{ .Tag }}/rukpak.yaml
     kubectl wait --for=condition=Available --namespace=rukpak-system deployment/plain-provisioner --timeout=60s
+    kubectl wait --for=condition=Available --namespace=rukpak-system deployment/registry-provisioner --timeout=60s
```
timflannagan, line 118: :+1:

internal/provisioner/plain/main.go
```diff
@@ -165,15 +167,31 @@ func main() {
 		os.Exit(1)
 	}
 
+	const plainBundleProvisionerName = "plain"
```
timflannagan, line 170: does it make sense to move constant in the same package that contains the plain provisioner's unique ID constant variable?

internal/provisioner/registry/controllers/bundle_controller.go
```diff
@@ -0,0 +1,285 @@
...
+			)
+		}
+		return ctrl.Result{}, err
+	}
+
+	unpackResult, err := r.Unpacker.Unpack(ctx, bundle)
+	if err != nil {
+		return ctrl.Result{}, updateStatusUnpackFailing(&u, fmt.Errorf("source bundle content: %w", err))
+	}
+	switch unpackResult.State {
```
timflannagan, line 143: Do we need to account for a failing source state here?

internal/provisioner/registry/finalizer/deletecachedbundle.go
```diff
@@ -0,0 +1,22 @@
+package finalizer
```
timflannagan, line 1: Any idea on why we couldn't have this live as a top-level internal package that both provisioner implementations can use? I don't see any noticable distinction between the registry + plain finalizer logic unless I'm missing something obvious.

internal/source/git.go
```diff
@@ -0,0 +1,189 @@
...
+	PodNamespace    string
+	UnpackImage     string
+	GitClientImage  string
+}
+
+const gitBundleUnpackContainerName = "bundle"
+
+func (g *Git) Unpack(ctx context.Context, bundle *rukpakv1alpha1.Bundle) (*Result, error) {
+	if bundle.Spec.Source.Type != rukpakv1alpha1.SourceTypeGit {
+		return nil, fmt.Errorf("bundle source type %q not supported", bundle.Spec.Source.Type)
```
timflannagan, line 38: Any value in surfacing that we expect this source type to be git in this error message too?

internal/source/git.go
```diff
@@ -0,0 +1,189 @@
...
+	KubeClient      kubernetes.Interface
+	ProvisionerName string
+	PodNamespace    string
+	UnpackImage     string
+	GitClientImage  string
+}
+
+const gitBundleUnpackContainerName = "bundle"
+
+func (g *Git) Unpack(ctx context.Context, bundle *rukpakv1alpha1.Bundle) (*Result, error) {
```
timflannagan, line 36: What happens if I want my provisioner to unpack using a different mechanism than deploying and managing a Pod?

timflannagan:
I held off on commenting more because I think we'd need to push through the Unpack interface PR first, and functionality the implementation seems reasonable to me once we add tests + more cleanup. I think we can de-duplicate a lot of the registry/plain implementation as a follow-up anyways.

I mentioned this in slack too but we can update the [nested registry README.md](https://github.com/operator-framework/rukpak/tree/3ee1cd3408f806474a11323158f97169f20df385/internal/provisioner/registry) now that we have something more concrete in place. It would be nice to also showcase in user documentation an example of a BundleInstance resource that has differing unique provisioner IDs being specified.

timflannagan:
I briefly playing around with this locally and wasn't able to the plain provisioner working using the example you had pasted in your PR description. When inspecting the content URL contents, it looks like we have the following tree structure:

```console
# output tar contents to the extract directory
$ tree extract
tree extract/
extract/
└── manifests
    └── manifest.yaml
```

Which seems fine to me at a glance. It seems like the real problem is the manifests.yaml file contains a reference to an OperatorConditions custom resource, and my cluster doesn't have OLM installed. I guess my question here is does requiring OLM CRDs here in order to use a registry+v1 converter make sense? Is this interaction something we need to live with in the short term while we continue iterating?

internal/source/git.go
joelanford: That'd be fine, you'd pass a different implementation into `source.NewUnpacker` that does something else in this function.

joelanford:
> I guess my question here is does requiring OLM CRDs here in order to use a registry+v1 converter make sense? Is this interaction something we need to live with in the short term while we continue iterating?

Yeah - as is, this PR would require the `OperatorCondition` CRD to be installed in the cluster. I think its reasonable to remove this for now, but it will be interesting to see if there are OLM operators that completely fail if they can't interact with their `OperatorCondition` object.

internal/provisioner/registry/controllers/bundle_controller.go
```diff
@@ -0,0 +1,285 @@
...
+			Reason:  rukpakv1alpha1.ReasonUnpackFailed,
+			Message: err.Error(),
+		}),
+	)
+	return err
+}
+
+func getObjects(bundleFS fs.FS) ([]client.Object, error) {
+	var objects []client.Object
+	const manifestsDir = "manifests"
```
exdx, line 240: I had two thoughts here:
* Maybe we should export this constant somewhere, since both the registry and plain provisioners rely on this directory being the source of content for the given bundle? 
* In the registry+v1 format, technically the manifests can live in any directory (and that can be specified in the image labels or the annotations.yaml file). I know this is later on in the code, after the conversion, but do we ensure that whatever directory is specified actually contains the manifests? I feel like 99% of the time it's `/manifests` anyways but I guess we do need to support that use-case?

timflannagan:
> but it will be interesting to see if there are OLM operators that completely fail if they can't interact with their OperatorCondition object.

This is something I was tracking down yesterday and it wasn't clear to me how many existing OLM operators utilize this functionality. That said, if the high level goal with this POC is to avoid operator/bundle authors having to manual migrate from the registry+v1 bundle format, then we'll need to account for that OperatorConditions resource if there's existing logic in some of these operator's controller implementations.

Maybe we can have explicit OperatorConditions handlement as a follow-up to this POC, and keep the scope of these changes fairly tight given we already have the restrictions around install modes + webhook/APIService definitions?

internal/convert/registryv1.go
exdx: 🎉

internal/provisioner/registry/main.go
```diff
@@ -0,0 +1,214 @@
...
+	flag.StringVar(&httpExternalAddr, "http-external-address", "http://localhost:8080", "The external address at which the http server is reachable.")
+	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
+	flag.StringVar(&systemNamespace, "system-namespace", "rukpak-system", "Configures the namespace that gets used to deploy system resources.")
+	flag.StringVar(&unpackImage, "unpack-image", "quay.io/operator-framework/rukpak:latest", "Configures the container image that gets used to unpack Bundle contents.")
+	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
+		"Enable leader election for controller manager. "+
+			"Enabling this will ensure there is only one active controller manager.")
+	flag.BoolVar(&rukpakVersion, "version", false, "Displays rukpak version information")
+	flag.StringVar(&storageDirectory, "storage-dir", storage.DefaultBundleCacheDir, "Configures the directory that is used to store Bundle contents.")
+	flag.StringVar(&gitClientImage, "git-client-image", "alpine/git:v2.32.0", "Configures which git container image to use to clone bundle git repos.")
```
exdx, line 81: nit: do we need git support for the registry+v1 provisioner? Unless we are introducing supporting sourcing registry+v1 bundles from git as well? I don't think it's a bad idea, if we get it for free, but we are sort-of introducing a new format this way.

internal/util/util.go
```diff
@@ -122,8 +151,27 @@ func CheckDesiredBundleTemplate(existingBundle *rukpakv1alpha1.Bundle, desiredBu
 
 // GenerateBundleName is responsible for generating a unique
 // Bundle metadata.Name using the biName parameter as the base.
-func GenerateBundleName(biName string) string {
-	return fmt.Sprintf("%s-%s", biName, rand.String(5))
+func GenerateBundleName(bi *rukpakv1alpha1.BundleInstance) string {
```
exdx, line 154: This is consistent with the plain provisioner implementation in #377 which is cool. I don't suppose there is any risk of collisions either (between a plain and registry version of a bundle), since the registry bundles would have the registry provisioner ID specified in their template, which would change the generated hash.

exdx (commented):
Nice. I reviewed this a few times and had some thoughts.

timflannagan (commented):
Largely looks ready to go. Something quick I noticed: do we need to update the goreleaser template and add the registry binary to the list of builds?

internal/storage/localdir.go
joelanford: Solved via #401

internal/util/util.go
joelanford: Solved in #399

internal/convert/registryv1.go
joelanford: actually, no we don't. the `Annotations` field of `annotationsFile` is not a pointer.

internal/provisioner/registry/controllers/bundle_controller.go
joelanford: No, the valid states are `Pending`, `Unpacking`, and `Unpacked`. A failing state would be just returning an error from `Unpack`.

internal/provisioner/registry/finalizer/deletecachedbundle.go
joelanford: Resolved in #390

internal/provisioner/registry/controllers/bundle_controller.go
joelanford: This function is actually just doing the same plain+v0 bundle validation that we do in the plain manifest bundle (notice that we pass `plainFS` into it). So this actually has nothing to do with the registry+v1 manifests directory.

joelanford: That highlights to me that we might want a plain+v0 validation function that both provisioners can call.

internal/provisioner/registry/main.go
timflannagan: I think this is might be fine to keep in right now but I do wonder whether we'd ever add support for sourcing from git repositories using this provisioner.

internal/convert/registryv1.go
```diff
@@ -76,13 +202,18 @@ func Convert(in RegistryV1, installNamespace string, targetNamespaces []string)
 		return nil, err
 	}
 
+	if len(in.CSV.Spec.APIServiceDefinitions.Owned) > 0 {
+		return nil, fmt.Errorf("apiServiceDefintions are not supported")
+	}
+
+	if len(in.CSV.Spec.WebhookDefinitions) > 0 {
+		return nil, fmt.Errorf("webhookDefiniions are not supported")
+	}
```
timflannagan, line 211: Follow-up: enforce these constraints in tests?

timflannagan (approved):
Awesome - I'm excited to see this get in even if we're duplicating code in the shorter term. I think the only thing missing outside of the follow-ups you have noted down is updating the nested registry README.md now that we have something sort-of-concrete from a provisioner implementation despite piggy backing off the plain provisioner's BI controller.

timflannagan:
@joelanford Should we add a "Closes https://github.com/operator-framework/rukpak/issues/402" here?

internal/provisioner/registry/main.go
joelanford: In my mind, the method that we use to arrive at a `fs.FS` that represents the bundle is fully divorced from the contents and schema of that `fs.FS` (plain+v0, registry+v1, whatever)

IMO we should stop co-mingling the ideas as part of the rukpak migration. The conversation yesterday about sourcing from `ConfigMaps` would have the same question right? "Do we need ConfigMap support for the registry+v1 provisioner"

I can't think of a good reason to artificially limit how we unpack bundles because all of that unpack code is reused no matter what the format is.

internal/convert/registryv1.go
```diff
@@ -98,127 +229,101 @@ func Convert(in RegistryV1, installNamespace string, targetNamespaces []string)
...
-					},
-				},
-				RoleRef: rbacv1.RoleRef{
-					APIGroup: rbacv1.GroupName,
-					Kind:     "Role",
-					Name:     name,
-				},
+	}
+
+	//oc := &v2.OperatorCondition{
```
tylerslaton, line 262: nit: should we remove these comments?

internal/convert/registryv1.go
```diff
@@ -265,18 +370,111 @@ func generateName(base string, o interface{}) string {
...
+		Subjects: subjects,
+		RoleRef: rbacv1.RoleRef{
+			APIGroup: rbacv1.GroupName,
+			Kind:     "ClusterRole",
+			Name:     roleName,
+		},
+	}
+}
+
+//func keys[Key comparable, Value any](m map[Key]Value) []Key {
```
tylerslaton, line 466: nit: should we remove these comments?

tylerslaton (approved):
Nice, running locally the behavior was exactly how I'd expect! There's a lot here, but I think it looks really good. Something I would like to see in a follow-up is definitely some `registryv1` e2e tests and unit tests for the `registryv1.go` file. The other thing we should think through in a follow-up is (as mentioned earlier) if we want this provisioner to rely on OLM CRDs, which it may. That being said, super excited to see this get through 🚀

internal/convert/registryv1.go
```diff
@@ -1,34 +1,146 @@
...
+			Mode:    fs.ModeDir | 0755,
+			ModTime: now,
+		},
+		"manifests": &fstest.MapFile{
+			Data:    nil,
+			Mode:    fs.ModeDir | 0755,
+			ModTime: now,
+		},
+		"manifests/manifest.yaml": &fstest.MapFile{
+			Data:    manifest.Bytes(),
```
exdx, line 135: nit: how come we store the buffer data as bytes versus a string? My thinking was we would want to persist a string representation of the converted manifests to disk (in the tar.gz) for clients to access later.

internal/provisioner/registry/main.go
exdx: Yup, that's fair -- I think decoupling the two makes total sense.

I think my concern here was that down the line, in a downstream context, users may run into issues with sourcing registry+v1 bundles from git (not related to the implementation, but in general) and then we would be on the hook for supporting them. But I think the point about not limiting how we unpack bundles outweighs that concern.

exdx (approved):
This is great! LGTM 

As a follow-up, I think we will need some precise documentation around this provisioner, since it explicitly denies a lot of legitimate registry+v1 bundles from the current OLM world. Hopefully users would see the docs before trying out their existing bundles and experiencing an error.

internal/convert/registryv1.go
joelanford: This is just what the MapFS implementation wants. It's supports any files, including binary files, hence using byte slices.

internal/convert/registryv1.go
joelanford: This is all the stuff related to supporting OperatorCondition generation. There's a reasonable chance we'll need this, but I agree that keeping all these comments isn't the best.

I'll add another commit that drops the comments. That way these comments will be in the history so we can revive them if necessary.

PR #392, opened 2022-05-24
Makefile: Factor out kubectl wait commands into their own target

timflannagan:
Update the install target and factor out the wait commands into the own target, which allows CI to run `make wait` without having to redefine the same steps. See #320 for where these changes could be useful.


tylerslaton (approved):

Makefile
```diff
@@ -100,12 +100,16 @@ kind-cluster-cleanup: kind ## Delete the kind cluster
...
 ###################
-.PHONY: install run cert-mgr
+.PHONY: install install-manifests wait run cert-mgr uninstall
 
 ##@ install/run:
 
-install: generate cert-mgr ## Install rukpak
+install: generate cert-mgr install-manifests wait ## Install rukpak
+
+install-manifests:
```
timflannagan, line 109: Note: I didn't add a comment for this (or the wait) target as they seem like "internal" commands from the user's perspective.

tylerslaton: I agree with that.

exdx (approved):

PR #395, opened 2022-05-24
e2e: Add a test that ensures Bundles referencing the wrong provisioner ID don't get unpacked

timflannagan:


exdx (approved):
Makes sense to codify this behavior, even if it is a bit of a blind spot in the current implementation.

timflannagan:
cc @joelanford I had this test lying around from my registry+v1 converter POC branch when trying to debug why my Bundle that referenced the unique converter provisioner ID wasn't being stored correctly, but I think we may just care about enforcing that the plain provisioner's Pod watcher correctly get requeued for the provisioner that owns that unpack Pod?

joelanford (approved):

PR #396, opened 2022-05-24
Add package for handling bundle formats

ryantking:
This adds package `github.com/operator-framework/rukpak/pkg/bundle`, which is a public API for working with and converting between bundle formats. Support is limited for the `registry+v1` and `plain+v0` formats and converting from the former to the latter.


pkg/convert/registryv1.go
```diff
@@ -248,7 +217,78 @@ func Convert(in RegistryV1, installNamespace string, targetNamespaces []string)
...
+		singleNamespace       = supportedInstallModes.Has(string(v1alpha1.InstallModeTypeSingleNamespace))
+		ownNamespace          = supportedInstallModes.Has(string(v1alpha1.InstallModeTypeOwnNamespace))
+		multiNamespace        = supportedInstallModes.Has(string(v1alpha1.InstallModeTypeMultiNamespace))
+	)
+
+	if set.Len() == 0 && allNamespaces {
+		return nil
+	}
+	if set.Len() == 1 &&
+		((set.Has("") && allNamespaces) || singleNamespace || (ownNamespace && set.Has(installNamespace))) {
```
joelanford, line 261:
```suggestion
		((set.Has("") && allNamespaces) || (singleNamespace && !set.Has(installNamespace)) || (ownNamespace && set.Has(installNamespace))) {
```

timflannagan (commented):
I took a glance at these changes and they seem relatively reasonable to me. I'm a fan of using the options pattern here to help improve interacting with this library.

I have a couple of quick questions:
- Can we create a tracking issue for modularizing this library?
- Does it make sense to have a standalone commit for moving the convert library from the internal directory to the pkg directory?

I think my only concern right now is support guarantees with exported functions in a pkg/ directory that clients depend on while we continue to iterate on this project. Maybe it's sufficient in the short term to have a nested README that adds a warning that the exported functions/behaviors are subject to change? Thoughts?

joelanford (commented):
Nice! This is looking much cleaner than before.

I think the next step (could be in a follow-up) would be to have functions that can convert:
-  `fs.FS` => `RegistryV1`
-  `Plain` => `fs.FS`

That way, it would be simple to chain `fs.FS => RegistryV1 => Plain => fs.FS`

joelanford:
> I think my only concern right now is support guarantees with exported functions in a pkg/ directory that clients depend on while we continue to iterate on this project.

As long as we stay at v0, we're good to make breaking changes on minor version bumps of this repo.

timflannagan:
> That way, it would be simple to chain fs.FS => RegistryV1 => Plain => fs.FS

Yeah, this seems super reasonable to me. Does the registry code around FBCs also have an intermediate representation layer too?

pkg/bundle/plain/v0/bundle.go
```diff
@@ -0,0 +1,382 @@
...
+		roleBinding.Namespace = ns
+		roleBindings = append(roleBindings, &roleBinding)
+	}
+
+	b.StoreObjects(roleBinding.Name+"_generated.yaml", roleBinding) // TODO(ryantking): Better name
+	return nil
+}
+
+func (b Bundle) generateName(base string, o interface{}) string {
+	const maxNameLength = 63
```
exdx, line 358: nit: we enforce that the length of a rukpak Bundle type is maximum 40 characters long (see https://github.com/operator-framework/rukpak/blob/main/manifests/apis/crds/patches/bundle_validation.yaml#L1). Would it make sense to have a similar limit here?

pkg/bundle/plain/v0/bundle.go
```diff
@@ -0,0 +1,382 @@
...
+		base = base[:maxNameLength-len(hashStr)-1]
+	}
+
+	return fmt.Sprintf("%s-%s", base, hashStr)
+}
+
+// deepHashObject writes specified object to hash using the spew library
+// which follows pointers and prints actual values of the nested objects
+// ensuring the hash does not change when a pointer changes.
+func (Bundle) deepHashObject(hasher hash.Hash, objectToWrite interface{}) {
```
exdx, line 373: nit: there is a method in util that already does this https://github.com/operator-framework/rukpak/blob/main/internal/util/hash.go -- would it make sense to use that instead of a method to avoid code duplication?

exdx (commented):
This is a very cool PR, I like the use of generics. I had some high-level feedback.

pkg/bundle/plain/v0/bundle.go
```diff
@@ -0,0 +1,382 @@
...
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+// Bundle holds the contents of a plain+v0 bundle.
+type Bundle struct {
+	bundle.FS
+
+	// external settings
+	installNamespace string
+	targetNamespaces []string
```
tylerslaton, line 29: I got a little confused when I first started reading this package as the concept of namespacing, install modes and csv deployments are all `registry+v1` specific. I understand that all the functions relating to these concepts like `validateOptions()` and `supportedInstallModes()` are used to convert from a `registry+v1` bundle in `FromRegistryV1()`. However, I believe we can make this code a bit more readable for future contributors. I was thinking we could do one of three things:

1. Improve the comments around these functions describing why they are necessary and emphasizing that they are not a core part of what a `plain` bundle is (and are just for conversions)
2. In this `plain/v0` package add a new `bundle_conversion.go` file that contains these functions like normal in order to logically separate the concepts.
3. Move the conversion burden onto the `registry/v1` package and have something along the lines of `registryv1.ToPlainV0()` vs `plainv0.FromRegistryV1`.

pkg/bundle/plain/v0/bundle.go
```diff
@@ -0,0 +1,382 @@
...
+
+	// Keep track of service accounts to avoid duplicates
+	createdSvcAccs map[string]struct{}
+}
+
+// New creates a new plain+v0 bundle at the root of the given filesystem.
+//
+// If the file system containse another known bundle format, it will be
```
tylerslaton, line 37:
```suggestion
// If the file system contains another known bundle format, it will be
```

pkg/bundle/fs.go
```diff
@@ -0,0 +1,187 @@
...
+}
+
+// StoreObjects adds objects to the name filed.
+func (fsys FS) StoreObjects(name string, objs ...client.Object) {
+	objFile := fsys.storedObjs[name]
+	objFile.Objects = append(objFile.Objects, objs...)
+	fsys.storedObjs[name] = objFile
+}
+
+func (fsys FS) mergeFilterFns(filterFns []func(client.Object) bool) func(client.Object) bool {
```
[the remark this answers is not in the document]
timflannagan: Should we make this a variadic parameter here?

```go
type filterFn func(client.Object)bool

func (fsys FS) mergeFilterFns(filters ...filterFn) filterFn {
    ...
}
```

pkg/bundle/fs.go
```diff
@@ -0,0 +1,187 @@
...
+// it will return a `File` with the manifests parsed to their underlying types.
+func (fsys FS) Open(name string) (fs.File, error) {
+	// TODO(ryantking): When we figure out how to store objects, this will change
+	if objFile, ok := fsys.storedObjs[name]; ok {
+		return objFile, nil
+	}
+
+	f, err := fsys.baseFS.Open(name)
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
ryantking: No, if `err` is non nil here then `f` is not a valid file handler. Once we find `err` to be `nil` then we know that `f` is a valid file handler that must be closed.

pkg/bundle/plain/v0/bundle.go
ryantking: Pinging @joelanford and @timflannagan to weigh on this one.

pkg/bundle/fs.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+	if !fsys.isManifestFile(name) {
+		return f, nil
+	}
+
+	objFile, err := NewObjectFile[client.Object](f, fsys.scheme, fsys.strictTypes)
+	if err != nil {
+		f.Close() // close the file that's already open
```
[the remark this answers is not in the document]
ryantking: Yes, `objFile` is just a wrapper around `f` that passes most operations down to it and adds parsing to the `Open` function when applicable.

pkg/bundle/plain/v0/bundle.go
ryantking: Thanks for the suggestions. One of my major concerns, probably _the_ major concern, is designing a really simple and understandable API so this is really helpful.

pkg/bundle/fs.go
```diff
@@ -0,0 +1,187 @@
...
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// FS is a File System that contains a bundle.
+//
+// A bundle is a directory that contains files that may or may not be
```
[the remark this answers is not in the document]
ryantking: So do we want to enforce the semantics of a bundle here then? For example, should the `bundle.FS` know that the `manifests` directory has special meaning and enforce that it can't have sub directories? My thinking was that it would be up to the specific bundle types to actually implement the semantics and a `bundle.FS` is just a filesystem with kubernetes manifests.

pkg/bundle/fs.go
```diff
@@ -0,0 +1,187 @@
...
+	strictTypes  bool
+
+	// TODO(ryantking): This is a hack since there is no API for modifying files in an fs.FS
+	storedObjs map[string]ObjectFile[client.Object]
+}
+
+// New returns a new FS wrapped around the given base FS.
+func New(baseFS fs.FS, opts ...func(*FS)) FS {
+	scheme := runtime.NewScheme()
+	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
```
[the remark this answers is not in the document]
ryantking: I instead supported a list of `AddToScheme` functions so the user can add their own types to the scheme, let me know what you think about that.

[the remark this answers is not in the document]
tylerslaton: Nice, I like that.

pkg/bundle/fs_test.go
```diff
@@ -0,0 +1,110 @@
...
+			f, err = fsys.Open(name)
+		})
+
+		When("it is a normal file", func() {
+			BeforeEach(func() {
+				baseFS = testutil.NewPlainV0FS()
+				name = "Dockerfile"
+			})
+
+			It("should not return an error", func() {
```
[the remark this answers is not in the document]
tylerslaton: Did this ever get solved?

pkg/bundle/plain/v0/bundle.go
tylerslaton: Marking this as resolved. The comment + new way of converting Bundles feels very extensible and intuitive. Great work!

tylerslaton (approved):
Nice! This looks good to me. Great job, Ryan. 🚀

pkg/bundle/fs_test.go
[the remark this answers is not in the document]
ryantking: Yes it's in the objectfile test cases since all the FS does it pass it down to the constructor.

pkg/bundle/fs.go
```diff
@@ -0,0 +1,199 @@
...
+// FS is a File System that contains a bundle.
+//
+// A bundle is a directory that contains files that may or may not be
+// Kubernetes manifests. When opening a manifest file, it will return an
+// `ObjectFile` type that allows the consumer to access the typed objects
+// without first opening/parsing the manifest itself. It still fully supports
+// all file operations.
+//
+// TODO(ryantking): Should we support any other file system abstractions beyond `fs.FS`?
+// `fs.FS` doesn't support any write operations. They can modify the objects in an `ObjectFile`, but that won't change
```
joelanford, line 27: IMO, fs.FS is sufficient. We can use https://pkg.go.dev/testing/fstest#MapFS as a writable in-memory FS implementation, and if we want a writable disk implementation, then we can just use `os.*` functions to create the necessary files/directories and then use `os.DirFS` to read it.

pkg/bundle/plain/v0/convert.go
```diff
@@ -0,0 +1,358 @@
...
+		return
+	}
+	if ns, ok := csv.Annotations["operatorframework.io/suggested-namespace"]; ok {
+		b.installNamespace = ns
+		return
+	}
+
+	// TODO(ryantking): Does srcBundle.CSV().GetName() give us package name
+	// per @joelanford's install namespace detection suggestion?
+	b.installNamespace = fmt.Sprintf("%s-system", csv.GetName())
```
joelanford, line 86: No, `csv.GetName()` will return the CSV name (typically something like `foo.v0.1.0`). We want this to be the package name (typically something like `foo`).

The package name can be discovered from a registry+v1 bundle's metadata/annotations.yaml file.

test/testutil/fs.go
```diff
@@ -0,0 +1,62 @@
...
+	fsys, err := makeMapFS("testdata/plain-v0/")
+	if err != nil {
+		panic(err)
+	}
+
+	return fsys
+}
+
+// embed.FS is only walkable from the root directory so this function accepts a prefix to filter out unwanted files.
+// This looks like it could be a go bug if its not explicitly called out.
```
joelanford, line 34: Can you use https://pkg.go.dev/io/fs#Sub here? It returns an fs that is a subdirectory of the root FS.

pkg/bundle/fs.go
```diff
@@ -0,0 +1,199 @@
...
+	return objFile, nil
+}
+
+func (fsys FS) isManifestFile(name string) bool {
+	if filepath.Ext(name) != ".yaml" || filepath.Ext(name) == ".yml" {
+		return false
+	}
+	if len(fsys.manifestDirs) == 0 {
+		return true
+	}
```
joelanford, line 126: In the case of a theoretical helm chart bundle, I assume there would be no manifest dirs, which would mean every file _is_ a manifest file? Should this block actually return false?

pkg/bundle/fs.go
```diff
@@ -0,0 +1,199 @@
...
+	for _, dir := range fsys.manifestDirs {
+		if strings.HasPrefix(name, dir) {
+			return true
+		}
+	}
```
joelanford, line 131: This seems like something that should be left to the bundle specification. For explicitness and clarity, we said that plain manifest bundles:
- _must_ have a `manifests` directory in the root of the bundle
- _must not_ have any subdirectories in the `manifests` directory (i.e. the `manifests` directory must be a flat list of manifest files.

However seems like it would parse manifest files in subdirectories of a `manifests` dir. I think this gets back to my comment that perhaps the generic bundle code should not have any concept of manifests or kubernetes objects? 

How much would it blow up this PR if we said that we wouldn't parse k8s objects at all in this generic bundle.FS layer? Would there be anything valuable left?

PR #399, opened 2022-05-25
consider provisioner ID before enqueuing requests for bundles

### LINKED ISSUE #393: Update provisioner pod handler to correctly check provisioner class mapping
at the bundle.Spec.Source.Git.Ref before we unpack the Bundle contents

that are stored in the repository.

r.GitClientImage currently defaults to alpine/git:v2.32.0

addUnpackerInitContainer assumes the pod has an array of init containers initialized.

https://github.com/operator-framework/rukpak/blob/5343f290ded4d98beb9e368441d849b33799a47b/internal/provisioner/plain/controllers/bundle_controller.go#L271

```golang

		For(&rukpakv1alpha1.Bundle{}, builder.WithPredicates(
			util.BundleProvisionerFilter(plain.ProvisionerID),
		)).
		// The default unpacker creates Pod's ownerref'd to its bundle, so
		// we need to watch pods to ensure we reconcile events coming from these
		// pods.
		// TODO: Update provisioner pod handler to correctly check provisioner class mapping
		Owns(&corev1.Pod{}).
		Complete(r)
}

```

joelanford:
Add a controller event handler implementation that maps events to reconciliation requests for owned objects that have the expected provisionerClassName field. 

This is used in place of the generic `handler.EnqueueRequestForOwner` implementation because it additionally considers the owner's provisioner class name. This extra consideration is important when multiple provisioners exist in a cluster and it is necessary for them to ignore events for owned objects that reference a bundle that is assigned to another provisioner class.

Reviewer note: I implemented the `MapOwneeToOwnerProvisionerHandler` function such that it would be trivial to use it for BundleInstance in the future if that need ever arises.

@timflannagan mentioned that while this approach does work, it _is_ fairly complex to implement. It requires looking up the owner object and checking its provisioner class name. I think we want to keep the owner reference on the unpack pods so that we get garbage collection for free when Bundles are deleted, but I do think it's worth considering whether there might be a better way to implement this mapping.

Closes #393

### DIFF
--- a/api/v1alpha1/bundle_types.go
+++ b/api/v1alpha1/bundle_types.go
@@ -123,6 +123,10 @@ type Bundle struct {
 	Status BundleStatus `json:"status,omitempty"`
 }
 
+func (b *Bundle) ProvisionerClassName() string {
+	return b.Spec.ProvisionerClassName
+}
+
 //+kubebuilder:object:root=true
 
 // BundleList contains a list of Bundle
--- a/api/v1alpha1/bundle_webhook.go
+++ b/api/v1alpha1/bundle_webhook.go
@@ -30,9 +30,9 @@ import (
 // log is for logging in this package.
 var bundlelog = logf.Log.WithName("bundle-resource")
 
-func (r *Bundle) SetupWebhookWithManager(mgr ctrl.Manager) error {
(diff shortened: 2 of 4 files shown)


timflannagan (approved):
This seems like it would do the trick, and I don't want to block pulling work out from the POC, but it feels like this implementation introduces extra complexity we'll have to life with over time.

PR #400, opened 2022-05-25
docs: Update plain Bundle names to include new content based hash

exdx:
The names of embedded Bundles have recently been modified to include the
hash of the template as part of the name. This means the names of the
Bundles are now deterministic. This PR updates the documentation to
reflect the new content-addressable Bundle name in the examples.

### DIFF
--- a/internal/provisioner/plain/README.md
+++ b/internal/provisioner/plain/README.md
@@ -56,8 +56,8 @@ spec:
       provisionerClassName: core.rukpak.io/plain
 ```
 
-> Note: the generated Bundle will contain the BundleInstance's metadata.Name as a prefix, following
-> a randomized value to prevent collisions with other Bundle resources.
+> Note: the generated Bundle will contain the BundleInstance's metadata.Name as a prefix, followed by
+> the hash of the provided template.
 
 First, the Bundle will be in the Pending stage as the provisioner sees it and begins unpacking the referenced content:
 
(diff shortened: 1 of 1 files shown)


timflannagan (commented):
Do we expect that the Bundle's name will change when the desired template has been updated? It looks like we use a single Bundle metadata.Name for all the changes here?

exdx:
> Do we expect that the Bundle's name will change when the desired template has been updated? It looks like we use a single Bundle metadata.Name for all the changes here?

IIUC the names look similar but they are different -- the original is `combo-7cdc7d7d6d` and then the updated one is `combo-7ddfd9fcd5`

tylerslaton (approved):
LGTM, only thing I'd mention is in my comment above ^

exdx:
> > > Do we expect that the Bundle's name will change when the desired template has been updated? It looks like we use a single Bundle metadata.Name for all the changes here?
> > 
> > 
> > IIUC the names look similar but they are different -- the original is `combo-7cdc7d7d6d` and then the updated one is `combo-7ddfd9fcd5`
> 
> Maybe it makes sense to run again and generate a new name to get more obviously unique names?

The thing is, the hash is deterministic, so running it again would produce the same result (if we don't change the underlying BundleTemplate). Since the difference between the old version of the Bundle and the new version is small (just the image reference changes), I guess the hash ends up looking similar as well. This is surprising to me, since the [avalanche effect](https://en.wikipedia.org/wiki/Avalanche_effect) should result in very different hashes for a small change in input. But I guess the hashing algorithm we are using (FNV32a) doesn't seem to produce that effect. 

We could add more metadata (labels, annotations, etc) to produce a larger diff between the two Bundles, but that may take away from the original point in the example. I think I will just add an explicit callout in the doc between the difference in Bundle names between old and new. If in the future, we feel like we need the hashes to be very different, we could change the underlying hash algorithm and it would be pretty straightforward.

exdx:
Yeah, that makes more sense -- it could just be a coincidence 🤷

PR #401, opened 2022-05-25
Add HTTP storage Loader to enable loading of other provisioner's bundles

joelanford:
When multiple provisioners exist, it will be possible for a bundle instance with provisioner class name "foo" to embed a bundle with provisioner class name "bar". In this case, the "foo" bundle instance provisioner will need to be able to load the bundle content of the "bar" bundle, whose storage is handled by another provisioner.

This PR enables the "foo" provisioner to fallback to loading a bundle via its `status.contentURL` if it experiences an error loading that bundle from its storage.

### DIFF
--- a/internal/provisioner/plain/controllers/bundle_controller.go
+++ b/internal/provisioner/plain/controllers/bundle_controller.go
@@ -57,6 +57,7 @@ type BundleReconciler struct {
 //+kubebuilder:rbac:groups=core.rukpak.io,resources=bundles,verbs=list;watch
 //+kubebuilder:rbac:groups=core.rukpak.io,resources=bundles/status,verbs=update;patch
 //+kubebuilder:rbac:groups=core.rukpak.io,resources=bundles/finalizers,verbs=update
+//+kubebuilder:rbac:verbs=get,urls=/bundles/*
 //+kubebuilder:rbac:groups=core,resources=pods,verbs=list;watch;create;delete
 //+kubebuilder:rbac:groups=core,resources=pods/log,verbs=get
 //+kubebuilder:rbac:groups=authentication.k8s.io,resources=tokenreviews,verbs=create
--- a/internal/provisioner/plain/main.go
+++ b/internal/provisioner/plain/main.go
@@ -134,10 +134,18 @@ func main() {
 		setupLog.Error(err, "unable to parse bundle content server URL")
 		os.Exit(1)
 	}
-	bundleStorage := &storage.LocalDirectory{
+
+	localStorage := &storage.LocalDirectory{
 		RootDirectory: storageDirectory,
 		URL:           *storageURL,
 	}
+	httpLoader := storage.NewHTTP(
+		// TODO: Use a trusted CA to connect to bundle content URLs.
+		storage.WithInsecureSkipVerify(true),
+		storage.WithBearerToken(cfg.BearerToken),
+	)
+	bundleStorage := storage.WithFallbackLoader(localStorage, httpLoader)
+
 	// NOTE: AddMetricsExtraHandler isn't actually metrics-specific. We can run
 	// whatever handlers we want on the existing webserver that
 	// controller-runtime runs when MetricsBindAddress is configured on the
--- a/internal/storage/http.go
+++ b/internal/storage/http.go
@@ -1,0 +1,78 @@
+package storage
+
+import (
+	"compress/gzip"
+	"context"
+	"crypto/tls"
+	"fmt"
+	"io/fs"
+	"net/http"
+	"time"
+
+	"github.com/nlepage/go-tarfs"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	rukpakv1alpha1 "github.com/operator-framework/rukpak/api/v1alpha1"
+)
+
+type HTTP struct {
+	client      http.Client
+	requestOpts []func(*http.Request)
+}
+
+type HTTPOption func(*HTTP)
+
+func WithInsecureSkipVerify(v bool) HTTPOption {
+	return func(s *HTTP) {
+		tr := s.client.Transport.(*http.Transport)
+		if tr.TLSClientConfig == nil {
+			tr.TLSClientConfig = &tls.Config{}
+		}
+		tr.TLSClientConfig.InsecureSkipVerify = v
+	}
+}
+
+func WithBearerToken(token string) HTTPOption {
+	return func(s *HTTP) {
+		s.requestOpts = append(s.requestOpts, func(request *http.Request) {
+			request.Header.Set("Authorization", fmt.Sprintf("Bearer %s", token))
+		})
+	}
+}
+
+type HTTPRequestOption func(*http.Request)
+
+func NewHTTP(opts ...HTTPOption) *HTTP {
+	s := &HTTP{client: http.Client{
+		Timeout:   time.Minute,
+		Transport: http.DefaultTransport.(*http.Transport).Clone(),
+	}}
+	for _, f := range opts {
+		f(s)
+	}
+	return s
+}
+
+func (s *HTTP) Load(ctx context.Context, owner client.Object) (fs.FS, error) {
+	bundle := owner.(*rukpakv1alpha1.Bundle)
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, bundle.Status.ContentURL, nil)
+	if err != nil {
+		return nil, err
+	}
+	for _, f := range s.requestOpts {
+		f(req)
+	}
+	resp, err := s.client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("unexpected response status %q", resp.Status)
+	}
+	tarReader, err := gzip.NewReader(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+	return tarfs.New(tarReader)
+}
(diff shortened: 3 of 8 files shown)


internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -54,9 +54,10 @@ type BundleReconciler struct {
 	Unpacker   source.Unpacker
 }
 
-//+kubebuilder:rbac:groups=core.rukpak.io,resources=bundles,verbs=list;watch
+//+kubebuilder:rbac:groups=core.rukpak.io,resources=bundles,verbs=list;watch;update;patch
```
timflannagan, line 57: Any idea on why this is needed given we should only need to perform updates/patches on the status sub-resource?

joelanford: This is needed to add finalizers. We haven't needed it before because the plain provisioner has */*/* in support of the bundle instance side of things.

I noticed we needed it in #387 for the registry provisioner, but this is unrelated to this PR, so I'll pull this out and keep it in #387

timflannagan: ah right forgot about the finalizer aspect here. good catch.

timflannagan:
I think this is also reasonable to me but it would be nice to have some follow-up tests that mock this behavior.

joelanford:
Yeah +1 on tests. This PR is super forward looking and so doesn't have any way to do e2es easily.

I can probably whip up some unit tests for this tomorrow.

internal/provisioner/plain/main.go
```diff
@@ -134,10 +134,18 @@ func main() {
...
-	bundleStorage := &storage.LocalDirectory{
+
+	localStorage := &storage.LocalDirectory{
 		RootDirectory: storageDirectory,
 		URL:           *storageURL,
 	}
+	httpLoader := storage.NewHTTP(
+		// TODO: Use a trusted CA to connect to bundle content URLs.
+		storage.WithInsecureSkipVerify(true),
+		storage.WithBearerToken(cfg.BearerToken),
```
tylerslaton, line 145: How would we start making progress on this todo and how blocking should we consider it for this PR?

tylerslaton (commented):
The implementation here looks super clean. Like we said above getting some tests here would make me comfortable getting the code through.

internal/provisioner/plain/main.go
joelanford: I think it's non-blocking for this PR, and I have a follow-up that solves it :)

joelanford:
Tests added!

internal/storage/http.go
```diff
@@ -0,0 +1,78 @@
...
+	"net/http"
+	"time"
+
+	"github.com/nlepage/go-tarfs"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	rukpakv1alpha1 "github.com/operator-framework/rukpak/api/v1alpha1"
+)
+
+type HTTP struct {
```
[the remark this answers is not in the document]
joelanford: This is `storage.HTTP`, so I think it's fine? It would only collide with the http package if you did something like `http := storage.HTTP{}`

internal/storage/storage.go
```diff
@@ -9,10 +9,38 @@ import (
...
+func WithFallbackLoader(s Storage, fallback Loader) Storage {
+	return &fallbackLoaderStorage{
+		Storage:        s,
+		fallbackLoader: fallback,
+	}
+}
+
+func (s *fallbackLoaderStorage) Load(ctx context.Context, owner client.Object) (fs.FS, error) {
+	fsys, err := s.Storage.Load(ctx, owner)
+	if err != nil {
```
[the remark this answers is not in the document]
joelanford: I figured it was best to fallback if _any_ error happens. That way we maximize the chance of success.

It would be up to the primary storage loader to retry, try alternate methods or whatever. Basically, a loader returning an error is essentially saying "I give up", in which case I think it makes sense for the fallback to give it a shot.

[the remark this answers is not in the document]
joelanford: I suppose we could make a `ErrSkipFallback` or something that loaders could use to tell any potential fallback loader not to bother. Seems unnecessary though.

internal/storage/http.go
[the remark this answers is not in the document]
exdx: oh yeah, I overlooked it's in the storage package. OK, sounds good then.

internal/storage/storage.go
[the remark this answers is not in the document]
exdx: Yes, I think if we're saying if the primary storage loader fails for any reason, then use the fallback loader then this implementation makes sense without introducing some more refined error handling.

internal/storage/http.go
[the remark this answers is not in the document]
timflannagan: nit: godocs because it's a net-new struct :upside_down_face:

internal/storage/http.go
```diff
@@ -0,0 +1,78 @@
...
+		Transport: http.DefaultTransport.(*http.Transport).Clone(),
+	}}
+	for _, f := range opts {
+		f(s)
+	}
+	return s
+}
+
+func (s *HTTP) Load(ctx context.Context, owner client.Object) (fs.FS, error) {
+	bundle := owner.(*rukpakv1alpha1.Bundle)
```
timflannagan, line 57: nit: add a cast check in case the owner isn't a Bundle struct?

internal/storage/http_test.go
```diff
@@ -0,0 +1,135 @@
...
+			localStore *LocalDirectory
+			server     http.Server
+			wg         sync.WaitGroup
+		)
+
+		BeforeEach(func() {
+			ctx = context.Background()
+			owner = &rukpakv1alpha1.Bundle{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: fmt.Sprintf("test-bundle-%s", rand.String(5)),
```
timflannagan, line 57: nit: Use https://github.com/operator-framework/rukpak/blob/79d30ffa32a9df9b8dd2f47438a4e62137b79a25/internal/util/util.go#L137-L139 here?

internal/storage/http_test.go
```diff
@@ -0,0 +1,135 @@
...
+			Expect(err).To(BeNil())
+
+			// Create the local store to store a bundle and to run the server that
+			// the HTTP loader will connect to.
+			port := l.Addr().(*net.TCPAddr).Port
+			baseURL, err := url.Parse(fmt.Sprintf("http://127.0.0.1:%d/bundles/", port))
+			Expect(err).To(BeNil())
+			localStore = &LocalDirectory{
+				URL:           *baseURL,
+				RootDirectory: GinkgoT().TempDir(),
```
timflannagan, line 74: TIL :eyes:

internal/storage/http_test.go
```diff
@@ -0,0 +1,135 @@
...
+			}
+
+			// Set the bundle's content URL (regardless of whether there's actually a file there)
+			owner.Status.ContentURL, err = localStore.URLFor(context.Background(), owner)
```
timflannagan, line 78: nit: you have a ctx variable in scope already?

internal/provisioner/plain/main.go
```diff
@@ -134,10 +134,18 @@ func main() {
...
 		os.Exit(1)
 	}
-	bundleStorage := &storage.LocalDirectory{
+
+	localStorage := &storage.LocalDirectory{
 		RootDirectory: storageDirectory,
 		URL:           *storageURL,
 	}
+	httpLoader := storage.NewHTTP(
+		// TODO: Use a trusted CA to connect to bundle content URLs.
```
timflannagan, line 143: Create tracking issue?

timflannagan (approved):
Nice work!

internal/storage/http.go
joelanford: I'm thinking we might just want to change the storage interface to accept rukpakv1alpha1.Bundle directly.  I don't think we have a use case for storage working with anything other than that. WDYT? Follow-up?

For now, I feel like panicing is better than returning an error because there's nothing an end user can do if this cast fails.

internal/provisioner/plain/main.go
joelanford: This should happen automatically when this PR merges, right?

internal/storage/http.go
tylerslaton: I'm +1 to this function taking a `rukpakv1alpha1.Bundle` directly since we're using its properties directly on line 58.

internal/storage/storage.go
```diff
@@ -9,10 +9,38 @@ import (
...
+		return s.fallbackLoader.Load(ctx, owner)
```
tylerslaton, line 43: I replaced this line locally with:

```go
return nil, fmt.Errorf("BREAKING")
```

And noticed that the unit test suite was still passing. Could we get coverage on this line to verify its functionality?

tylerslaton (commented):
One thing I wanted to confirm, everything looks great though!

internal/storage/storage.go
joelanford: Oh good catch. I don't have tests for the fallback loader.

internal/storage/storage_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+		store = WithFallbackLoader(primaryStore, fallbackStore)
+	})
+
+	It("should find primary bundle", func() {
+		loadedTestFS, err := store.Load(ctx, primaryBundle)
+		Expect(err).To(BeNil())
+		Expect(fsEqual(primaryFS, loadedTestFS))
+	})
+	It("should find fallback bundle", func() {
```
tylerslaton, line 66: 🎉

tylerslaton (approved):

PR #405, opened 2022-05-27
plain provisioner: use a CA-signed certificate for kube-rbac-proxy

### LINKED ISSUE #404: Use a trusted CA to connect to bundle content URLs.
whatever handlers we want on the existing webserver that

controller-runtime runs when MetricsBindAddress is configured on the

https://github.com/operator-framework/rukpak/blob/f50dcc1ad84fc27958fc3bc1b5434135997a872d/internal/provisioner/plain/main.go#L143

```golang

		setupLog.Error(err, "unable to parse bundle content server URL")
		os.Exit(1)
	}

	localStorage := &storage.LocalDirectory{
		RootDirectory: storageDirectory,
		URL:           *storageURL,
	}
	httpLoader := storage.NewHTTP(
		// TODO: Use a trusted CA to connect to bundle content URLs.
		storage.WithInsecureSkipVerify(true),
		storage.WithBearerToken(cfg.BearerToken),
	)
	bundleStorage := storage.WithFallbackLoader(localStorage, httpLoader)

	// NOTE: AddMetricsExtraHandler isn't actually metrics-specific. We can run
	// whatever handlers we want on the existing webserver that
	// controller-runtime runs when MetricsBindAddress is configured on the

```

joelanford:
Closes #404


internal/provisioner/plain/main.go
```diff
@@ -139,9 +142,18 @@ func main() {
...
+		if rootCAs, err = util.LoadCertPool(bundleCAFile); err != nil {
+			setupLog.Error(err, "unable to parse bundle certificate authority file")
+			os.Exit(1)
+		}
+	}
+
 	httpLoader := storage.NewHTTP(
-		// TODO: Use a trusted CA to connect to bundle content URLs.
-		storage.WithInsecureSkipVerify(true),
+		storage.WithRootCAs(rootCAs),
```
timflannagan, line 156: Should we still expose a knob for using insecureSkipVerify, e.g. --insecure-skip-tls-verify?

joelanford: I'm not opposed, but do we have a use case where we need that flag now that this PR boostraps certs and CAs?

timflannagan: I'd guess the use case would be dev environments, but if we're already leaning on cert-mgr to give us certificates for free we might be all set?

timflannagan (approved):

PR #407, opened 2022-05-28
add private git repository support

### LINKED ISSUE #285: Support private git source bundle

akihikokuroda:
This is private git source support refactored for go-git library #336 .

Closes #285


internal/source/git.go
```diff
@@ -96,6 +113,18 @@ func (r *Git) Unpack(ctx context.Context, bundle *rukpakv1alpha1.Bundle) (*Resul
 	return &Result{Bundle: bundleFS, ResolvedSource: resolvedSource, State: StateUnpacked}, nil
 }
 
+func (r *Git) getCredential(ctx context.Context, bundle *rukpakv1alpha1.Bundle) (string, string, error) {
+	if bundle.Spec.Source.Git.SecretName == "" {
```
exdx, line 117: Would it potentially make the logic more straightforward if this function were only called in the case where the `SecretName` is not blank? We could have something like the following:

```
if bundle.Spec.Source.Git.SecretName != "" {
    auth, err := r.getCredentials(ctx, bundle)
    if err != nil {
        return nil, err
    }
    cloneOpts.Auth = auth
}

File: internal/provisioner/plain/main.go [hunk 2]
```diff
@@ -119,6 +120,7 @@ func main() {
 			SelectorsByObject: cache.SelectorsByObject{
 				&rukpakv1alpha1.BundleInstance{}: {},
 				&rukpakv1alpha1.Bundle{}:         {},
+				&corev1.Secret{}:                 {},
```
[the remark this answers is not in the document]
akihikokuroda: @exdx The Get (sigs.k8s.io/controller-runtime/pkg/client) didn't work without it. Am I missing something to make it work?
```
       secret := &corev1.Secret{}
	err := r.Get(ctx, client.ObjectKey{Namespace: r.PodNamespace, Name: bundle.Spec.Source.Git.SecretName}, secret)
```
Should I change it to the client in `k8s.io/client-go/kubernetes`?
Thanks!

internal/source/git.go
```diff
@@ -96,6 +113,18 @@ func (r *Git) Unpack(ctx context.Context, bundle *rukpakv1alpha1.Bundle) (*Resul
...
+		return "", "", nil
+	}
+	secret := &corev1.Secret{}
+	err := r.Get(ctx, client.ObjectKey{Namespace: r.PodNamespace, Name: bundle.Spec.Source.Git.SecretName}, secret)
```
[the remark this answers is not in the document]
akihikokuroda: I thought so but it didn't work for me.

internal/source/git.go
akihikokuroda: It's better.  I'll update it. Thanks!  I moved the `SecretName` check in the function for some reason in the PR for pre-golib refactor.  I don't remember the reason.  It may not be a valid reason any more after the refactor.

internal/provisioner/plain/main.go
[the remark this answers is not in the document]
exdx: Hmm that's interesting, I didn't expect that, but admittedly I don't know that much about the c/r client. I suppose maybe the c/r client only talks to its cache directly by design? 

I'm not sure about introducing the client-go client to the controller here. Maybe @timflannagan could give us a suggestion?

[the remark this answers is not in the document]
joelanford: Ideally, what we end up with is our cache contains secrets such that:
    - We can get/list/watch ALL secrets in the `rukpak-system` namespace
    - We can only get/list watch secrets with the rukpak selector labels in other namespaces.

Off the top of my head, I don't know whether there's a simple way to accomplish that with a single cache. We basically need the informer to have:
`metadata.namespace == "rukpak-system" || rukpakSelector.Match(metadata.labels)`

In the meantime, we could use `mgr.GetAPIReader` instead of `mgr.GetClient`. We'd bypass the cache with that which is:
- Good: we don't have to watch all secrets in the cluster
- Bad: we have to hit the API server during every reconcile of a bundle that contains a secret ref.

I'd propose that we go with the "Bad" alternative for now. And make a TODO to look into optimizing the way our cache is constructed.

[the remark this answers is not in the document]
exdx: That would be better -- I agree that the "bad" alternative is good enough for now and we can potentially revisit later if we run into scaling issues.

internal/source/git.go
```diff
@@ -13,12 +13,17 @@ import (
...
 	"github.com/go-git/go-git/v5/storage/memory"
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	rukpakv1alpha1 "github.com/operator-framework/rukpak/api/v1alpha1"
 )
 
 type Git struct {
+	client.Client
+	PodNamespace string
```
joelanford, line 26: This is the namespace we're going to read secrets from right? In that case, I'd suggest:
```suggestion
	SecretNamespace string
```

internal/source/git.go
```diff
@@ -13,12 +13,17 @@ import (
...
+	"github.com/go-git/go-git/v5/plumbing/transport/http"
 	"github.com/go-git/go-git/v5/storage/memory"
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	rukpakv1alpha1 "github.com/operator-framework/rukpak/api/v1alpha1"
 )
 
 type Git struct {
+	client.Client
```
joelanford, line 25: Use `client.Reader` here instead. I assume the git source has to need to update/patch/delete?

internal/source/git.go
```diff
@@ -36,9 +41,24 @@ func (r *Git) Unpack(ctx context.Context, bundle *rukpakv1alpha1.Bundle) (*Resul
...
-		Progress: os.Stdout,
-		Tags:     git.NoTags,
+		URL:             gitsource.Repository,
+		Progress:        os.Stdout,
+		Tags:            git.NoTags,
+		InsecureSkipTLS: bundle.Spec.Source.Git.TLSNoVerify,
+	}
+
+	if bundle.Spec.Source.Git.SecretName != "" {
+		userName, accessToken, err := r.getCredential(ctx, bundle)
```
joelanford, line 51: Rather than `accessToken` terminology, can we use `password` everywhere (here, and as the key in the `secret.data` map). I think `password` is the generally accepted terminology that the git project uses in docs about authentication.

Some git hosts (e.g. github) support access tokens as passwords, but I think access token is a more specific term than password, and not one we should enshrine in the API.

joelanford: In fact, there's a specific secret type in kubernetes that enforces these fields: https://kubernetes.io/docs/concepts/configuration/secret/#basic-authentication-secret

Let's align with that.

api/v1alpha1/bundle_types.go
```diff
@@ -81,6 +81,11 @@ type GitSource struct {
 	// is required. Setting more than one field or zero fields will result in an
 	// error.
 	Ref GitRef `json:"ref"`
+	// SecretName contains the secret name that has authorization information for HTTPS protocol and is in the namespace that the provisioner is deployed.
+	// The secret is expected to contain `data.username` and `data.accesstoken` for the username and personal access token, respectively.
+	SecretName string `json:"secretName,omitempty"`
+	// If TslNoVerify is set true, the server certificate is ignored.
```
joelanford, line 87: Nit: GoDoc should start with the Identifier: `// TLSNoVerify ...` There's also a typo in the spelling of `TslNoVerify` in the comment.

Secondly, another nit: What should we call this field?
- `wget` uses `--no-check-certificate`
- `curl` uses `--insecure`
- Go Stdlib uses `InsecureSkipVerify`

Anyone else have examples of how different things refer to this concept?

I'm partial to using `Insecure` as the prefix, at least. That makes it clearer that this is a generally frowned-upon field to use (and we should state that in our GoDoc comment).

api/v1alpha1/bundle_types.go
```diff
@@ -81,6 +81,11 @@ type GitSource struct {
...
+	TLSNoVerify bool `json:"tlsNoVerify,omitempty"`
```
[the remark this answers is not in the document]
akihikokuroda: One secret that has different contents for different auth types probably can cover most of the cases.  Is it enough to add the auth type field and HTTP auth value for now?  We can add the certificate auth in the follow up. 

```
	AuthorizationTypeCertificate = "certificate"
	AuthorizationTypeAccessToken = "accessToken"

	// AuthorizatonType defines the kind of authorizaton method being used.
	AuthorizationType string `json:"authorizationType,omitempty"`
```

internal/source/git.go
```diff
@@ -38,9 +43,24 @@ func (r *Git) Unpack(ctx context.Context, bundle *rukpakv1alpha1.Bundle) (*Resul
...
+		userName, password, err := r.getCredential(ctx, bundle)
+		if err != nil {
+			return nil, err
+		}
+		if password == "" {
+			return nil, errors.New("data.accessToken is missing")
+		}
+		if userName == "" {
+			userName = "user"
+		}
```
exdx, line 62: nit: could we put this logic inside `getCredential` so it returns either the auth or an error?

exdx (approved):

api/v1alpha1/bundle_types.go
```diff
@@ -95,6 +100,16 @@ type GitRef struct {
 	Commit string `json:"commit,omitempty"`
 }
 
+type Authorization struct {
+	// Type defines the kind of authorization method being used.
+	AuthorizationType string `json:"authorizationType,omitempty"`
```
joelanford, line 105: I think we only want `AuthorizationType` if we actually have a union (something like):
```go
type Authorization struct {
	AuthorizationType string `json:"authorizationType,omitempty"`
	Secret corev1.LocalObjectReference `json:"secret,omitempty"`
	HTTP *HTTPAuthorization `json:"http,omitempty"`
	SSH *SSHAuthorization `json:"ssh,omitempty"`
}

type HTTPAuthorization struct {
	InsecureSkipVerify bool `json:"insecureSkipVerify,omitempty"`
}
...
```

I'm torn though. Another union type seems like more complexity for us to manage. And even if we're doing https, it's only necessary to include the HTTPAuthorization struct if you need to enable InsecureSkipVerify.

joelanford: Maybe for now, we keep things simple and:
1. Drop `AuthorizationType`
2. Document in GoDoc that `InsecureSkipVerify` is only used when cloning repos using the `https` scheme.

api/v1alpha1/bundle_types.go
```diff
@@ -95,6 +100,16 @@ type GitRef struct {
...
+	// Secret contains reference to the secret that has authorization information for HTTPS protocol and is in the namespace that the provisioner is deployed.
+	// The secret is expected to contain `data.username` and `data.password` for the username and password, respectively.
+	Secret corev1.LocalObjectReference `json:"secret,omitempty"`
+	// InsecureSkipVerify is set true, the server certificate is ignored. It is not safe to set this true for production environment use.
```
joelanford, line 109:
```suggestion
	// InsecureSkipVerify controls whether a client verifies the server's certificate chain and host name. This
	// configuration is used only when cloning git repositories that use the "https" scheme. If InsecureSkipVerify
	// is true, the clone operation will accept any certificate presented by the server and any host name in that
	// certificate. In this mode, TLS is susceptible to machine-in-the-middle attacks unless custom verification is
	// used. This should be used only for testing.
```

(I re-used a bunch from the upstream GoDoc about in the `crypto/tls` package.)

internal/source/git.go
```diff
@@ -98,6 +112,23 @@ func (r *Git) Unpack(ctx context.Context, bundle *rukpakv1alpha1.Bundle) (*Resul
...
+func (r *Git) getCredential(ctx context.Context, bundle *rukpakv1alpha1.Bundle) (string, string, error) {
+	secret := &corev1.Secret{}
+	err := r.Get(ctx, client.ObjectKey{Namespace: r.SecretNamespace, Name: bundle.Spec.Source.Git.Auth.Secret.Name}, secret)
+	if err != nil {
+		return "", "", err
+	}
+	userName := string(secret.Data["username"])
+	password := string(secret.Data["password"])
+	if password == "" {
+		return "", "", errors.New("data.accessToken is missing")
```
joelanford, line 124:
```suggestion
		return "", "", errors.New("data.password is missing")
```

internal/source/git.go
```diff
@@ -98,6 +112,23 @@ func (r *Git) Unpack(ctx context.Context, bundle *rukpakv1alpha1.Bundle) (*Resul
...
+	}
+	if userName == "" {
+		userName = "user"
+	}
```
joelanford, line 128: What's the idea here? Is it that we can use a dummy value when password is an access token? I wonder if we should just enforce that this secret is specifically a `basic-auth` secret, which means we get the guarantee from the kubernetes API server that `username` and `password` are always set.

internal/source/git.go
akihikokuroda: I missed it.  Thanks!

tylerslaton (approved):
Small nit suggestion, but looks good!

internal/source/git.go
```diff
@@ -98,6 +112,23 @@ func (r *Git) Unpack(ctx context.Context, bundle *rukpakv1alpha1.Bundle) (*Resul
 	return &Result{Bundle: bundleFS, ResolvedSource: resolvedSource, State: StateUnpacked}, nil
 }
 
+func (r *Git) getCredential(ctx context.Context, bundle *rukpakv1alpha1.Bundle) (string, string, error) {
```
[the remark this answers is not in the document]
akihikokuroda: Thanks!  I updated the function header.  It made easy to read.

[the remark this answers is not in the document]
joelanford: Just for future reference, putting names in the function signatures return values has a few side effects:
1. When the function scope is entered, those names are instantiated as variables with the associated return type
2. It allows [naked returns](https://go.dev/tour/basics/7), which IMO are a _hugely_ bad practice because it makes it much harder to mentally understand what's being returned at the return site.

I basically never use named return values for this reason (and instead, I include GoDoc that describe the return values).

I'd honestly be in favor of a linter that outright blocked this kind of thing in the future.

internal/source/git.go
```diff
@@ -98,6 +112,20 @@ func (r *Git) Unpack(ctx context.Context, bundle *rukpakv1alpha1.Bundle) (*Resul
...
+	secret := &corev1.Secret{}
+	err = r.Get(ctx, client.ObjectKey{Namespace: r.SecretNamespace, Name: bundle.Spec.Source.Git.Auth.Secret.Name}, secret)
+	if err != nil {
+		return
+	}
+	userName = string(secret.Data["username"])
+	password = string(secret.Data["password"])
+	if password == "" {
+		err = errors.New("data.password is missing")
+	}
```
joelanford, line 125: Should we remove this check as well?

joelanford:
For follow-up: I think we'll want an e2e that covers this. I know its a can of worms because we need to setup a git repo with a git server that requires basic auth, but something we need to do nonetheless.

internal/source/git.go
akihikokuroda: OK.  I take it out.

internal/source/git.go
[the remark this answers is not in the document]
tylerslaton: GoDoc is another option for sure. I'd be more comfortable with updating the comments with this to describe the return values based on what you're saying Joe. Unresolving.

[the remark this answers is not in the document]
akihikokuroda: OK.  I put it back and adding the comment for the function.

joelanford (approved):

Issue #408, opened 2022-05-30
Follow up git source refactor 


akihikokuroda:
This is follow up of RP #366 

There are 3 follow up items 
https://github.com/operator-framework/rukpak/pull/366#issuecomment-1140086432

PR #409, opened 2022-05-30
add gitsource directory check in validatingwebhook

akihikokuroda:

This a part of #408 

Adding "/" and "../" checks for the directory of gitsource.


api/v1alpha1/bundle_webhook.go
```diff
@@ -76,6 +76,9 @@ func checkBundleSource(r *Bundle) error {
 		if r.Spec.Source.Git == nil {
 			return fmt.Errorf("bundle.spec.source.git must be set for source type \"git\"")
 		}
+		if r.Spec.Source.Git.Directory != "" && (r.Spec.Source.Git.Directory[0] == '/' || r.Spec.Source.Git.Directory[:3] == "../") {
```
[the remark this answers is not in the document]
akihikokuroda: Thanks!  I'll add check.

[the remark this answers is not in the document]
akihikokuroda: Thanks for review.  I'll make changes.

api/v1alpha1/bundle_webhook.go
```diff
@@ -76,6 +76,14 @@ func checkBundleSource(r *Bundle) error {
 		if r.Spec.Source.Git == nil {
 			return fmt.Errorf("bundle.spec.source.git must be set for source type \"git\"")
 		}
+		if r.Spec.Source.Git.Directory != "" {
+			if r.Spec.Source.Git.Directory[0] == '/' {
+				return fmt.Errorf("bundle.spec.source.git.directory can be \"/\"")
+			}
+			if len(r.Spec.Source.Git.Directory) > 2 && r.Spec.Source.Git.Directory[:3] == "../" {
```
joelanford, line 83: 1 nit, 1 suggestion:
- nit: use `strings.HasPrefix` to simplify existing check.
- suggestion: we need to use `filepath.Clean` here as well 
    - `foo/../../` is the same as `../`. This code would pass that directory, but that should fail.
```suggestion
			if strings.HasPrefix(filepath.Clean(r.Spec.Source.Git.Directory), "../") {
```

api/v1alpha1/bundle_webhook.go
```diff
@@ -76,6 +76,14 @@ func checkBundleSource(r *Bundle) error {
 		if r.Spec.Source.Git == nil {
 			return fmt.Errorf("bundle.spec.source.git must be set for source type \"git\"")
 		}
+		if r.Spec.Source.Git.Directory != "" {
+			if r.Spec.Source.Git.Directory[0] == '/' {
+				return fmt.Errorf("bundle.spec.source.git.directory can be \"/\"")
+			}
```
joelanford, line 82: I think this should be allowed, and we should interpret it as the root of the repo. What would also mean that we'd treat `/foo` to mean that we will unpack `<repoRoot>/foo` as the bundle directory.

api/v1alpha1/bundle_webhook.go
```diff
@@ -76,6 +76,14 @@ func checkBundleSource(r *Bundle) error {
...
+			if len(r.Spec.Source.Git.Directory) > 2 && r.Spec.Source.Git.Directory[:3] == "../" {
+				return fmt.Errorf("bundle.spec.source.git.directory can not start with \"../\"")
```
joelanford, line 84: Nit: add a bit more context about why "../" is not allowed.

Also a tip that backticks can be used to define string literals to avoid needing to escape quotation marks within the string.
```suggestion
				return fmt.Errorf(`bundle.spec.source.git.directory begins with "../": directory must define path within the repository`)
```

api/v1alpha1/bundle_webhook.go
akihikokuroda: I see.  I'll take it out.  Thanks.

api/v1alpha1/bundle_webhook.go
akihikokuroda: I'll take your suggestion.  Thanks!

api/v1alpha1/bundle_webhook.go
akihikokuroda: I take your suggestion.  Thanks!

PR #410, opened 2022-05-30
put clone progress into log only on error

akihikokuroda:

This is a part of #408 

The clone progress of gitsource is put into log only on error

### DIFF
--- a/internal/source/git.go
+++ b/internal/source/git.go
@@ -1,6 +1,7 @@
 package source
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
@@ -35,9 +36,10 @@ func (r *Git) Unpack(ctx context.Context, bundle *rukpakv1alpha1.Bundle) (*Resul
 	}
 
 	// Set options for clone
+	progress := bytes.Buffer{}
 	cloneOpts := git.CloneOptions{
 		URL:      gitsource.Repository,
-		Progress: os.Stdout,
+		Progress: &progress,
 		Tags:     git.NoTags,
 	}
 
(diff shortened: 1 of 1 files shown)


internal/source/git.go
```diff
@@ -54,7 +56,7 @@ func (r *Git) Unpack(ctx context.Context, bundle *rukpakv1alpha1.Bundle) (*Resul
 	// Clone
 	repo, err := git.CloneContext(ctx, memory.NewStorage(), memfs.New(), &cloneOpts)
 	if err != nil {
-		return nil, fmt.Errorf("bundle unpack git clone error: %w", err)
+		return nil, fmt.Errorf("bundle unpack git clone error: %w - %s", err, progress.String())
```
joelanford, line 59: Would you mind including the resulting output of a clone failure log message so we can see how this looks?

Without having looked at it myself, I wonder if it would make sense to use a new line to allow the git clone output to begin on its own line?

```suggestion
		return nil, fmt.Errorf("bundle unpack git clone error: %w\n%s", err, progress.String())
```

But maybe my suggestion looks worse because the default logger encodes this error string such that the log actually ends up with "\n" in it?

exdx (approved):
LGTM, this is a good follow-up. I think implementing Joe's suggestion here would make this even better.

internal/source/git.go
akihikokuroda: @joelanford You are right.  The log message gets "\n" in it.  I leave it as-is.

tylerslaton (approved):
Thanks for the contribution!

PR #411, opened 2022-06-01
consolidate unpacker initialization; other minor provisioner main.go cleanups

joelanford:
The changes from #387 and #366 had some minor conflicts at the time they we're being authored because of code duplication around the setup of the unpacker.

This PR consolidates the unpacker initialization into a separate function, which will help us de-deduplicate code, avoid a similar situation in the future.

I've also been playing with a PVC source type, and I noticed that the addition of more sources will lead to even more duplication unless we consolidate.

### DIFF
--- a/internal/provisioner/plain/main.go
+++ b/internal/provisioner/plain/main.go
@@ -29,7 +29,6 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/selection"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
-	"k8s.io/client-go/kubernetes"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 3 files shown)


internal/provisioner/plain/main.go
```diff
@@ -95,25 +94,21 @@ func main() {
...
+
+	cfg := ctrl.GetConfigOrDie()
 	mgr, err := ctrl.NewManager(cfg, ctrl.Options{
 		Scheme:                 scheme,
 		MetricsBindAddress:     httpBindAddr,
 		Port:                   9443,
 		HealthProbeBindAddress: probeAddr,
 		LeaderElection:         enableLeaderElection,
-		LeaderElectionID:       "510f803c.olm.operatorframework.io",
+		LeaderElectionID:       "plain.core.rukpak.io",
```
tylerslaton, line 111: 🎉 Good catch.

internal/source/default.go
```diff
@@ -0,0 +1,26 @@
+package source
```
tylerslaton, line 1: nit: Would this function have a good home inside of the `unpacker.go` file?

PR #414, opened 2022-06-02
Git source private repository support via ssh protocol

### LINKED ISSUE #294: Support client certificate authorization for git source
The git source bundle supports only the personal access token authorization in #287.  The github also has the client certificate authorization.  It's good for the git source bundle to support it.

akihikokuroda:

This PR is a follow up of #407 (supports http(s) protocol for private repository).  This PR adds the support of ssh protocol with the certificate. 

Closes #294

### DIFF
--- a/api/v1alpha1/bundle_types.go
+++ b/api/v1alpha1/bundle_types.go
@@ -99,11 +99,13 @@ type GitRef struct {
 }
 
 type Authorization struct {
-	// Secret contains reference to the secret that has authorization information for HTTPS protocol and is in the namespace that the provisioner is deployed.
-	// The secret is expected to contain `data.username` and `data.password` for the username and password, respectively.
+	// Secret contains reference to the secret that has authorization information and is in the namespace that the provisioner is deployed.
+	// The secret is expected to contain `data.username` and `data.password` for the username and password, respectively for http(s) scheme.
+	// Refer to https://kubernetes.io/docs/concepts/configuration/secret/#basic-authentication-secret
+	// The secret is expected to contain `data.ssh-privatekey` and `data.ssh-knownhosts` for the ssh privatekey and the host entry in the known_hosts file respectively for ssh authorization.
+	// Refer to https://kubernetes.io/docs/concepts/configuration/secret/#ssh-authentication-secrets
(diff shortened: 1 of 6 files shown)


api/v1alpha1/bundle_types.go
```diff
@@ -100,10 +100,10 @@ type GitRef struct {
 
 type Authorization struct {
 	// Secret contains reference to the secret that has authorization information for HTTPS protocol and is in the namespace that the provisioner is deployed.
-	// The secret is expected to contain `data.username` and `data.password` for the username and password, respectively.
+	// The secret is expected to contain `data.username` and `data.password` for the username and password, respectively for http(s) scheme.
+	// The secret is expected to contain `data.certificate`, `data.passphrase` and `data.host` for the certificate, passphrase and the host entry in the known_hosts file respectively .
```
joelanford, line 104: Can we align the expected secret fields with the official `kubernetes.io/ssh-auth` secret type? I think it would be nice if we include links in the GoDoc for the http basic-auth and ssh-auth Kubernetes Secret types, so folks are using secrets that they may already be maintaining. 

I think we should only accept the non-passphrased key directly. The passphrase loses its effectiveness if you have to store it in plaintext right next to the encrypted private key.

I like the `data.host` idea of allowing a known_hosts file to be included (are you allowed to include _extra_ fields in a `kubernetes.io/ssh-auth` secret?). WDYT about naming with the same sort of pattern as the `ssh-privatekey` key. Perhaps: `data.ssh-knownhosts` or `data.ssh-known_hosts`

Here are the links I think we may want to consider including in the GoDoc.
- https://kubernetes.io/docs/concepts/configuration/secret/#basic-authentication-secret
- https://kubernetes.io/docs/concepts/configuration/secret/#ssh-authentication-secrets

akihikokuroda: I didn't know that kubernetes has these types defined.  For the basic-auth, the current code must refactored to use `stringData` instead of `data` of the secret. 
I'll take out the passphrase.  I saw functions in the library supporting both w/ and w/o passphrase so I implemented.  I've never set the passphrase myself.
For the ssh-auth, additional items can be added.  I'll use `data.ssh-knownhosts`.

exdx (approved):

PR #417, opened 2022-06-03
Introduce `upload` source and `rukpakctl run`

joelanford:
This PR introduces an `upload` source and a `rukpakctl run` command that makes use of the new `upload` source.

Closes https://github.com/operator-framework/rukpak/issues/472

The `local` source is a great way to help us get up and running quickly with a mechanism that enables direct usage bundle contents to a kubernetes cluster without requiring those bundles to be pushed to a git remote or an image registry. However, the `local` source's use of configmaps could only get us so far. Configmap size is restricted by etcd configuration (typically 2MB), so the `local` source could only support bundles up to 2MB in size. Some improvements for the `local` source are discussed in #464. 

The `upload` source does not place this restriction on bundle size because it does not rely on etcd for storage of local bundle content.

An upload source enables bundle developers or other clients to directly upload bundle tar.gz files to inject them into bundles.

This introduces a upload manager service that:
- accepts uploads for bundles with the `upload` source type
- serves those bundles to provisioners that request them
- garbage collects its local storage when bundle uploads are no longer needed

This also introduces client tooling necessary to actually exercise this functionality in a user-friendly way. It:
- Adds `rukpakctl run` to create a BD with an upload source bundle, get the name of the generated bundle, and then upload the bundle contents to the upload manager, which then makes that bundle content available to the bundle provisioners.
- Adds a simple library for running a port-forward to a specific service in the cluster (in this case, we use it to connect to the upload manager from outside the cluster)
- A `BundleUploader` type that is essentially a small API for actually doing the upload. This is used in `rukpakctl run` and also in the e2e tests.


joelanford:
Hmm.... I wonder if the inclusion of the `binary: {}` struct in the Bundle API is causing go-spew (and thus the hashing for name generation) to change the output. I ran the e2e twice and I'm pretty sure this PR's code produced the same (incorrect, according to the tests) hash suffix for the generated bundle.

joelanford:
> About the client tool, can it create the bundle instead of the bundle instance embedding a bundle? So it doesn't need to wait for the bundle name is determined. we can have another command to create the bundle instance that specify the bundle.

To actually run the bundle, we have to create a bundle instance that embeds the binary-type bundle. So at least in that case, I don't think we have any choice but to:
1. Create the BI
2. Wait for the bundle to exist
3. Upload the tar.gz for that bundle name

akihikokuroda:
@joelanford Thanks!  I didn't realize that the bundleinstance has to embed the bundle.  I thought it could still reference a bundle.

api/v1alpha1/bundle_types.go
```diff
@@ -62,8 +62,11 @@ type BundleSource struct {
 	Image *ImageSource `json:"image,omitempty"`
 	// Git is the git repository that backs the content of this Bundle.
 	Git *GitSource `json:"git,omitempty"`
-	// Local is a reference to a local object in the cluster.
-	Local *LocalSource `json:"local,omitempty"`
+	// Binary is a source that enables this Bundle's content to be uploaded
+	// via Rukpak's bundle upload service. This source type is primarily useful
+	// with bundle development workflows because it enabled bundle developers
```
exdx, line 67:
```suggestion
	// with bundle development workflows because it enables bundle developers
```

cmd/rukpakctl/cmd/content.go
```diff
@@ -28,9 +28,10 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/tools/clientcmd"
 	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+	config2 "sigs.k8s.io/controller-runtime/pkg/client/config"
```
exdx, line 34: nit: could we rename to config? I don't see a collision with another package here.

exdx (approved):

akihikokuroda (approved):

akihikokuroda:
LGTM.  The [design](https://hackmd.io/nprZyBwPQqmix94abAiMsA?view) refers to the configmap source.  Are we going to use the binary source instead of the comfigmap source there?

joelanford:
> The [design](https://hackmd.io/nprZyBwPQqmix94abAiMsA?view) refers to the configmap source. Are we going to use the binary source instead of the comfigmap source there?

That's a great question. We probably need to re-open that design to discuss the best way to accomplish what it proposes while avoiding the concerns raised in #464.

I think using the binary source is definitely an option. But that _is_ slightly more complex because the thing creating the `ResolveSet` bundle would have to perform the same upload function that `rukpakctl` uses.

cmd/rukpakctl/cmd/content.go
```diff
@@ -41,61 +42,51 @@ type options struct {
...
-	rootCmd.AddCommand(contentCmd)
+func newContentCmd() *cobra.Command {
+	var opt options
+
+	contentCmd := &cobra.Command{
+		Use:   "content <bundle name>",
+		Short: "display contents of the specified bundle.",
+		Long:  `display contents of the specified bundle.`,
+		Args:  cobra.ExactArgs(1),
+		Run: func(cmd *cobra.Command, args []string) {
```
timflannagan, line 53: I know this is just repurposing the existing cmd implementation but use RunE here instead to avoid os.Exit(1)?

cmd/rukpakctl/cmd/run.go
```diff
@@ -0,0 +1,149 @@
...
+		bundleDeploymentProvisionerClassName string
+		bundleProvisionerClassName           string
+	)
+
+	cmd := &cobra.Command{
+		Use:   "run <bundleDeploymentName> <bundleDir>",
+		Short: "Run a bundle from an upload of a local bundle directory.",
+		Long:  "Run a bundle from an upload of a local bundle directory.",
+		Run: func(cmd *cobra.Command, args []string) {
+			bundleDeploymentName, bundleDir := args[0], args[1]
```
timflannagan, line 47: Add a cobra.ExactArgs(2) to avoid any panics?

cmd/rukpakctl/cmd/run.go
```diff
@@ -0,0 +1,149 @@
...
+func newRunCmd() *cobra.Command {
+	var (
+		bundleDeploymentProvisionerClassName string
+		bundleProvisionerClassName           string
+	)
+
+	cmd := &cobra.Command{
+		Use:   "run <bundleDeploymentName> <bundleDir>",
+		Short: "Run a bundle from an upload of a local bundle directory.",
+		Long:  "Run a bundle from an upload of a local bundle directory.",
```
timflannagan, line 45: TODO: Add some examples in the long description?

cmd/rukpakctl/cmd/run.go
timflannagan: Looks like this was defined further down.

cmd/rukpakctl/cmd/run.go
```diff
@@ -0,0 +1,149 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
+
+	rukpakv1alpha1 "github.com/operator-framework/rukpak/api/v1alpha1"
+	"github.com/operator-framework/rukpak/internal/rukpakctl"
+	"github.com/operator-framework/rukpak/internal/util"
+)
+
+const (
+	rukpakSystemNamespace = "rukpak-system"
```
timflannagan, line 31: Make this configurable?

internal/provisioner/registry/main.go
```diff
@@ -179,7 +181,7 @@ func main() {
 	}
 
 	const registryBundleProvisionerName = "registry"
-	unpacker, err := source.NewDefaultUnpacker(mgr, ns, registryBundleProvisionerName, unpackImage)
+	unpacker, err := source.NewDefaultUnpacker(mgr, ns, registryBundleProvisionerName, unpackImage, baseBinaryDownloadURL, rootCAs)
```
timflannagan, line 184: This may be a candidate for refactoring given the number of function parameters now.

internal/rukpakctl/binary.go
```diff
@@ -0,0 +1,97 @@
...
+		if err != nil {
+			return ctx.Err()
+		}
+
+		bundleTgz := &bytes.Buffer{}
+		if err := util.FSToTarGZ(bundleTgz, bundleFS); err != nil {
+			return err
+		}
+
+		req, err := http.NewRequestWithContext(ctx, http.MethodPut, fmt.Sprintf("https://localhost:%d/bundles/%s", localPort, fmt.Sprintf("%s.tgz", bundleName)), bundleTgz)
```
timflannagan, line 52: Have a dedicated function for building the bundles URL?

internal/rukpakctl/ca.go
```diff
@@ -0,0 +1,23 @@
...
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func GetRukpakCA(ctx context.Context, cl client.Reader) (*x509.CertPool, error) {
+	rukpakCASecret := &corev1.Secret{}
+	if err := cl.Get(ctx, types.NamespacedName{Namespace: "rukpak-system", Name: "rukpak-ca"}, rukpakCASecret); err != nil {
```
timflannagan, line 15: Do we need to avoid hardcoding the namespaced name parameters here?

timflannagan (commented):
I gave this a quick sweep through without deep diving into the server implementations. I'm still on the fence for why we need to eliminate the local source type implementation, and why we need this binary upload service implementation, but I won't block progress as this improves local development (and rukpak users) experience quite a bit.

The only major things I'd like to see:
- Do we need to update the core webhook to also accommodate this new source type?
- Adding godocs for some of the exported utility packages that were added.

And we can tackle cleaning up the CLI packages over time.

internal/provisioner/registry/main.go
joelanford: Adding a TODO comment. Seems like something better left to a dedicate PR. WDYT?

timflannagan: Yeah, that's fine with me. I'm a big fan of smaller PRs anyways.

cmd/rukpakctl/cmd/content.go
joelanford: I agree this could be improved. `RunE` can be a little awkward to work with as well because it always spits out the usage if an error is returned.

You can disable that with `SilenceUsage`, but that can also be annoying if you have some failures where usage makes sense and some failures where it doesn't.

My general approach is to use a combination of `RunE` and `log.Fatal`. Any errors that are related to validating the CLI args would just be returned. Any runtime errors would get the `log.Fatal` treatment so that the usage is not printed.

joelanford: And if I know cobra can handle all the validation I need, I just use `Run` with `log.Fatal`.

joelanford:
> Do we need to update the core webhook to also accommodate this new source type?

Is there anything I'm missing about what we'd need to check? The bundle `spec` can't change (and there wouldn't be anything really worth changing since `binary` isn't configurable)

And the upload manager http server forbids re-uploads of already unpacked binary bundles, so I think that covers the immutability from that perspective.

> Adding godocs for some of the exported utility packages that were added.
Will do. They're all under the internal tree so I somewhat glossed over that, but I agree nonetheless. docs > no docs regardless of what/where.

timflannagan:
> Is there anything I'm missing about what we'd need to check? The bundle spec can't change (and there wouldn't be anything really worth changing since binary isn't configurable)

This is my bad: I forgot the core webhook just ensures Bundle spec immutability, and we moved the union type validation logic as oneOf CRD schema validations. I was talking about the latter when making that comment.

cmd/rukpakctl/cmd/run.go
```diff
@@ -0,0 +1,182 @@
...
+	var (
+		systemNamespace                      string
+		uploadServiceName                    string
+		caSecretName                         string
+		bundleDeploymentProvisionerClassName string
+		bundleProvisionerClassName           string
+	)
+
+	cmd := &cobra.Command{
+		Use:   "run <bundleDeploymentName> <bundleDir>",
```
exdx, line 41: WDYT about making the `bundleDeploymentName` optional, and just using the name of the bundle directory? It makes the UX even more easy

cmd/rukpakctl/cmd/run.go
```diff
@@ -0,0 +1,182 @@
...
+			bundleLabels := map[string]string{
+				"app":          bundleDeploymentName,
+				"bundleDigest": fmt.Sprintf("%x", digest.Sum(nil)),
+			}
+
+			bd := buildBundleDeployment(bundleDeploymentName, bundleLabels, bundleDeploymentProvisionerClassName, bundleProvisionerClassName)
+			if err := cl.Patch(ctx, bd, client.Apply, client.ForceOwnership, client.FieldOwner("rukpakctl")); err != nil {
+				log.Fatalf("apply bundle deployment: %v", err)
+			}
+			fmt.Printf("bundledeployment.core.rukpak.io %q applied\n", bundleDeploymentName)
```
exdx, line 102: Should we add a real logger and respect logging levels set by the client? Could also be done in a follow-up.

cmd/rukpakctl/cmd/run.go
```diff
@@ -0,0 +1,182 @@
...
+	cmd.Flags().StringVar(&systemNamespace, "system-namespace", "rukpak-system", "the namespace in which the rukpak controllers are deployed.")
+	cmd.Flags().StringVar(&uploadServiceName, "upload-service-name", "upload-manager", "the name of the service of the upload manager.")
+	cmd.Flags().StringVar(&caSecretName, "ca-secret-name", "rukpak-ca", "the name of the secret in the system namespace containing the root CAs used to authenticate the upload service.")
+	cmd.Flags().StringVar(&bundleDeploymentProvisionerClassName, "bundle-deployment-provisioner-class", "core.rukpak.io/plain", "Provisioner class name to set on bundle deployment.")
+	cmd.Flags().StringVar(&bundleProvisionerClassName, "bundle-provisioner-class", "core.rukpak.io/plain", "Provisioner class name to set on bundle.")
+	return cmd
+}
+
+func buildBundleDeployment(bdName string, bundleLabels map[string]string, biPCN, bPNC string) *unstructured.Unstructured {
+	return &unstructured.Unstructured{Object: map[string]interface{}{
```
exdx, line 140: What's the benefit of not using a typed BD object here?

internal/util/tar.go
```diff
@@ -0,0 +1,66 @@
...
+	"compress/gzip"
+	"fmt"
+	"io"
+	"io/fs"
+	"os"
+)
+
+// FSToTarGZ writes the filesystem represented by fsys to w as a gzipped tar archive.
+// This function unsets user and group information in the tar archive so that readers
+// of archives produced by this function do not need to account for differences in
```
exdx, line 14: This should help alleviate some of the issues we see when attempting to prune red hat catalog images that include root-permission directories. Maybe this is worth highlighting in the user-facing documentation? If it's not overly technical

exdx (approved):
LGTM. Nice job!

cmd/rukpakctl/cmd/run.go
joelanford: The main reason I chose that is to ensure that the pivot workflow works regardless of whether you use the same directory name. If you had `./my-bundle-v0.1.0` and `./my-bundle-v0.2.0`, it would be really easy to create 2 different BDs when what you really want is to re-use the same BD, and pivot it from v0.1.0 to v0.2.0.

cmd/rukpakctl/cmd/run.go
joelanford: We could - I was trying to match `kubectl` style output as closely as I could. But definitely agree that enabling more verbosity would be a nice improvement.

cmd/rukpakctl/cmd/run.go
joelanford: If you use a typed object, any default values from that struct get serialized into the JSON patch, which could cause unrelated fields to be patched back to the default value even though that isn't the intention. Using an unstructured ensures that the patch contains _only_ what is specified.

timflannagan: That seems like a reasonable explanation to me. Does it make sense to capture that as a comment there?

cmd/rukpakctl/cmd/run.go
```diff
@@ -0,0 +1,187 @@
...
+			},
+		},
+	}}
+}
+
+func getBundleName(ctx context.Context, cfg *rest.Config, bundleLabels map[string]string) (string, error) {
+	dynCl, err := dynamic.NewForConfig(cfg)
+	if err != nil {
+		return "", fmt.Errorf("build dynamic client: %v", err)
+	}
```
timflannagan, line 173: Not blocking: create the dynamic client at the driver level, and pass it as a paramter vs. the rest config?

cmd/rukpakctl/cmd/run.go
```diff
@@ -0,0 +1,187 @@
...
+
+	watch, err := dynCl.Resource(rukpakv1alpha1.GroupVersion.WithResource("bundles")).Watch(ctx, metav1.ListOptions{Watch: true, LabelSelector: labels.FormatLabels(bundleLabels)})
+	if err != nil {
+		return "", fmt.Errorf("watch bundles: %v", err)
+	}
+	defer watch.Stop()
+
+	select {
+	case evt := <-watch.ResultChan():
+		return evt.Object.(client.Object).GetName(), nil
```
timflannagan, line 183: Do you think we need to type assert here to avoid a panic? I feel like this should be largely fine.

cmd/uploadmgr/main.go
```diff
@@ -0,0 +1,280 @@
...
+	})
+	if err != nil {
+		setupLog.Error(err, "unable to create manager")
+		os.Exit(1)
+	}
+
+	// NOTE: AddMetricsExtraHandler isn't actually metrics-specific. We can run
+	// whatever handlers we want on the existing webserver that
+	// controller-runtime runs when MetricsBindAddress is configured on the
+	// manager.
```
timflannagan, line 95: :+1:

internal/rukpakctl/ca.go
```diff
@@ -0,0 +1,26 @@
...
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// GetClusterCA returns an x509.CertPool by reading the contents of a Kubernetes Secret. It uses the provided
+// client to get the requested secret and then loads the contents of the secret's "ca.crt" key into the cert pool.
+func GetClusterCA(ctx context.Context, cl client.Reader, ns, secretName string) (*x509.CertPool, error) {
```
timflannagan, line 16: Would it be easier to pass a NamespacedName instead of individual ns/secretName string fields?

internal/rukpakctl/portforward.go
```diff
@@ -0,0 +1,137 @@
...
+			}
+		}
+	}
+	if port == 0 {
+		return fmt.Errorf("could not find port %q for service %q", pf.port.String(), pf.serviceName)
+	}
+
+	path := fmt.Sprintf("/api/v1/namespaces/%s/pods/%s/portforward", pf.serviceNamespace, podName)
+	host := strings.TrimLeft(pf.cfg.Host, "htps:/")
+	serverURL := url.URL{Scheme: "https", Path: path, Host: host}
```
timflannagan, line 82: I think we could modularize this logic, but this is fine for now.

internal/rukpakctl/upload.go
```diff
@@ -0,0 +1,122 @@
...
+// Uploads of content that is identical to the existing bundle's content will not result in an error,
+// which means this function is idempotent. Running this function multiple times with the same input
+// does not result in any change to the cluster state after the initial upload.
+func (bu *BundleUploader) Upload(ctx context.Context, bundleName string, bundleFS fs.FS) (bool, error) {
+	pf, err := NewServicePortForwarder(bu.Cfg, types.NamespacedName{Namespace: bu.UploadServiceNamespace, Name: bu.UploadServiceName}, intstr.FromString("https"))
+	if err != nil {
+		return false, err
+	}
+
+	ctx, cancel := context.WithCancel(ctx)
```
timflannagan, line 49: Add a comment that we're explicitly canceling in the goroutine further down below?

internal/rukpakctl/upload.go
```diff
@@ -0,0 +1,122 @@
...
+// uploads the bundleFS as the content for the bundle named by the provided bundleName.
+//
+// Upload returns a boolean value indicating if the bundle content was modified on the server and
+// an error value that will convey any errors that occurred during the upload.
+//
+// Uploads of content that is identical to the existing bundle's content will not result in an error,
+// which means this function is idempotent. Running this function multiple times with the same input
+// does not result in any change to the cluster state after the initial upload.
+func (bu *BundleUploader) Upload(ctx context.Context, bundleName string, bundleFS fs.FS) (bool, error) {
+	pf, err := NewServicePortForwarder(bu.Cfg, types.NamespacedName{Namespace: bu.UploadServiceNamespace, Name: bu.UploadServiceName}, intstr.FromString("https"))
```
timflannagan, line 44: Any value in making the protocol a constant variable?

internal/rukpakctl/upload.go
```diff
@@ -0,0 +1,122 @@
...
+		return false, err
+	}
+
+	ctx, cancel := context.WithCancel(ctx)
+	eg, ctx := errgroup.WithContext(ctx)
+	eg.Go(func() error {
+		return pf.Start(ctx)
+	})
+
+	bundleReader, bundleWriter := io.Pipe()
```
timflannagan, line 55: Do we need to explicitly close this writer?

timflannagan: Andddd it looks like I'm blind and this is defined right below. TIL about CloseWithError. Does it make sense to have another comment here that we're closing the writer in another goroutine?

internal/source/unpacker.go
```diff
@@ -90,12 +94,19 @@ func (s *unpacker) Unpack(ctx context.Context, bundle *rukpakv1alpha1.Bundle) (*
 // NewDefaultUnpacker returns a new composite Source that unpacks bundles using
 // a default source mapping with built-in implementations of all of the supported
 // source types.
-func NewDefaultUnpacker(mgr ctrl.Manager, namespace, provisionerName, unpackImage string) (Unpacker, error) {
+//
+// TODO: refactor NewDefaultUnpacker due to growing parameter list
```
timflannagan, line 98: :+1:

internal/source/unpacker.go
```diff
@@ -112,5 +123,10 @@ func NewDefaultUnpacker(mgr ctrl.Manager, namespace, provisionerName, unpackImag
 			Client: mgr.GetClient(),
 			reader: mgr.GetAPIReader(),
 		},
+		rukpakv1alpha1.SourceTypeUpload: &Upload{
+			baseDownloadURL: baseUploadManagerURL,
+			bearerToken:     mgr.GetConfig().BearerToken,
+			client:          http.Client{Timeout: 10 * time.Second, Transport: httpTransport},
```
timflannagan, line 129: Any value in a constant variable for the timeout period?

internal/util/tar.go
timflannagan: Yep, +1 on that recommendation.

test/e2e/plain_provisioner_test.go
```diff
@@ -958,6 +960,124 @@ var _ = Describe("plain provisioner bundle", func() {
...
+
+		It("can unpack the bundle successfully", func() {
+			Eventually(func() error {
+				if err := c.Get(ctx, client.ObjectKeyFromObject(bundle), bundle); err != nil {
+					return err
+				}
+				if bundle.Status.Phase != rukpakv1alpha1.PhaseUnpacked {
+					return errors.New("bundle is not unpacked")
+				}
+				return nil
```
timflannagan, line 1014: Handle this through gomega matchers?

timflannagan (approved):
This is really nice work. Lots of complexity here, and I didn't play around with it locally so YMMV, but the packages that were introduced were pretty clean. Overall, I didn't have any blocking feedback, and I'm excited to see this finally get close to the finish line. I tried to avoid making a mass amount of comments, so I stopped midway through in hopes we can merge this before closing out this current release milestone.

Do we need any follow-up tickets around benchmarking this service? Does it make sense to call out in the user-facing documentation (if we haven't already) and/or the CRD schema documentation that this service can be used to handle arbitrarily sized bundles to avoid the etcd size limitation?

exdx:
One additional thought I had -- do we ensure that the kube client proxy to the service tears down gracefully after ending the upload? I have seen issues with `kubectl port-forward` where it would close the stream but the localhost port would still be in use. This would prevent the command from being idempotent.

joelanford:
> do we ensure that the kube client proxy to the service tears down gracefully after ending the upload?

If I understand your question, I think the answer is yes. We start a goroutine in the error group whose sole purpose is to wait for the context to be canceled and then close the stop channel.

https://github.com/joelanford/rukpak/blob/01f7e6f85dec68610bf9d0aa390fb991d9c28418/internal/rukpakctl/portforward.go#L104-L108

Then at the very end of that function we call `eg.Wait()` to block returning from `Start()` until all error group routines have completed.

cmd/rukpakctl/cmd/run.go
joelanford: If this starts panic-ing, something pretty drastic would have changed, I think we'd have more serious issues. Essentially, the bet we're making is that the event Object contains something that also implements `metav1.Object`, which seems like a safe bet until at least k8s 2.0.

internal/rukpakctl/upload.go
joelanford: You mean `const portName = "https"`? I'm not sure there's much value there. The only other place it needs to align is in the actual service definition which is in a yaml that can't reuse that const.

cmd/rukpakctl/cmd/run.go
timflannagan: Works for me :+1:

internal/rukpakctl/upload.go
timflannagan: And I imagine this isn't something we'd need to be able to configure down the line, right? I'm fine with the current implementation.

internal/source/unpacker.go
joelanford: Sure, why not!

internal/util/tar.go
joelanford: It might be a little bit technical given the existing way most users would interact with this (`rukpakctl run`). That subcommand does all the necessary tar.gz stuff under the hood and it is not exposed at all to users.

Regardless, I could definitely see value in documenting the upload manager service as a separate thing because there may be in-cluster processes that can directly interact with the upload manager without requiring the extra port-forward stuff that `rukpakctl run` provides.

joelanford:
> CRD schema documentation that this service can be used to handle arbitrarily sized bundles to avoid the etcd size limitation?

I didn't call this out specifically in the "upload" source struct GoDoc because the "image" and "git" sources also support arbitrarily sized bundles. Maybe we should instead call out the `local` source's limitations?

timflannagan:
Yeah, that sounds reasonable to me. That can be done as a follow-up anyways from my perspective.

PR #421, opened 2022-06-07
deps: bump k8s dependencies to 1.24

### LINKED ISSUE #397: bump k8s dependencies to 1.24
Update rukpak's k8s dependencies to the next upstream version of Kubernetes, 1.24. Currently the project is on 1.23.  

* ensure versions in go.mod point to 0.24.z
```
	k8s.io/api v0.23.1
	k8s.io/apiextensions-apiserver v0.23.1
	k8s.io/apimachinery v0.23.1
	k8s.io/client-go v0.23.1
```
* ensure kind cluster used by CI is also running 1.24
* ensure kubectl and any other testing dependencies are also based on 1.24 (envtest)

exdx:
Bumps all k8s-* dependencies to 1.24. 

Closes #397

### DIFF
--- a/test/e2e/plain_provisioner_test.go
+++ b/test/e2e/plain_provisioner_test.go
@@ -723,11 +723,6 @@ var _ = Describe("plain provisioner bundleinstance", func() {
 			)
 		})
 		It("should generate a Bundle that contains the correct labels", func() {
-			expectedLabels := map[string]string{
-				util.CoreOwnerKindKey:          rukpakv1alpha1.BundleInstanceKind,
-				util.CoreOwnerNameKey:          bi.GetName(),
-				util.CoreBundleTemplateHashKey: "54dd864b69",
-			}
 			Eventually(func() (map[string]string, error) {
 				if err := c.Get(ctx, client.ObjectKeyFromObject(bi), bi); err != nil {
 					return nil, err
@@ -739,8 +734,9 @@ var _ = Describe("plain provisioner bundleinstance", func() {
 				return b.Labels, nil
 			}).Should(And(
 				Not(BeNil()),
-				Equal(expectedLabels)),
-			)
+				WithTransform(func(s map[string]string) string { return s[util.CoreOwnerKindKey] }, Equal(rukpakv1alpha1.BundleInstanceKind)),
+				WithTransform(func(s map[string]string) string { return s[util.CoreOwnerNameKey] }, Equal(bi.GetName())),
+			))
 		})
 		Describe("template is unsuccessfully updated", func() {
 			var (


exdx:
Unfortunately, `go mod tidy` fails with the following: 
```
github.com/operator-framework/rukpak/internal/convert imports
	github.com/operator-framework/operator-registry/pkg/registry tested by
	github.com/operator-framework/operator-registry/pkg/registry.test imports
	github.com/operator-framework/operator-registry/pkg/lib/bundle imports
	github.com/operator-framework/api/pkg/validation imports
	github.com/operator-framework/api/pkg/validation/internal imports
	k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/validation imports
	k8s.io/apiserver/pkg/util/webhook imports
	k8s.io/component-base/traces imports
	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp imports
	go.opentelemetry.io/otel/semconv: module go.opentelemetry.io/otel@latest found (v1.7.0), but does not contain package go.opentelemetry.io/otel/semconv
github.com/operator-framework/rukpak/internal/convert imports
	github.com/operator-framework/operator-registry/pkg/registry tested by
	github.com/operator-framework/operator-registry/pkg/registry.test imports
	github.com/operator-framework/operator-registry/pkg/lib/bundle imports
	github.com/operator-framework/api/pkg/validation imports
	github.com/operator-framework/api/pkg/validation/internal imports
	k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/validation imports
	k8s.io/apiserver/pkg/util/webhook imports
	k8s.io/component-base/traces imports
	go.opentelemetry.io/otel/exporters/otlp imports
	go.opentelemetry.io/otel/sdk/metric/controller/basic imports
	go.opentelemetry.io/otel/metric/registry: module go.opentelemetry.io/otel/metric@latest found (v0.30.0), but does not contain package go.opentelemetry.io/otel/metric/registry
```

I believe this is due to the operator-registry relying on an old version of the [open-telemetry](https://github.com/open-telemetry/opentelemetry-go) project

timflannagan (commented):
Can we manually vendor anything we use from the registry here? It looks like it's a single type definition that we can inline into this repository, and cut out that dependency entirely.

exdx:
> Can we manually vendor anything we use from the registry here? It looks like it's a single type definition that we can inline into this repository, and cut out that dependency entirely.

Yes agreed -- that sounds much nicer, and will avoid a lot of pain going forward.

exdx:
> We should also bump the `kind` tool dependency to make sure we pick up a the kind version that deploys a 1.24 cluster by default. Looks like that's 0.14.0 at the moment.

Yes, definitely. I wanted to split the tooling bumps in a separate commit. Since the current issue with the e2e (#423)  is likely to be resolved in a standalone PR and this PR will then be rebased, I think it's fine to add that change now.

exdx:
Moving this out of WIP since the PR now does what is required: all direct dependencies and tooling are moved to 1.24 versions of kubernetes. 

Seeing two e2e failures, both seemingly unrelated:
```
[FAIL]plain provisioner bundleinstance embedded bundle template [It] should generate a Bundle that contains the correct labels
/home/runner/work/rukpak/rukpak/test/e2e/plain_provisioner_test.go:740
```
This is due to the non-determinism in the hash output per #423 
```
[FAIL]bundle api validating webhook the bundle name is too long [It] should fail the long name bundle creation
/home/runner/work/rukpak/rukpak/test/e2e/webhook_test.go:85
```
I believe this is due to the webhook getting bypassed for some reason? I may have seen a failure like this before in prior e2e runs. cc @akihikokuroda

akihikokuroda:
I believe that the validation error message must be changed.  The test needs to be changed to match to the new error message.

The webhook_test.go:85 must be changed to:
```
Expect(err).To(MatchError(ContainSubstring(fmt.Sprintf("Bundle.core.rukpak.io %q is invalid: metadata.name: Too long: may not be longer than 40", bundle.GetName()))))
```

exdx:
> I believe that the validation error message must be changed. The test needs to be changed to match to the new error message.
> 
> The webhook_test.go:85 must be changed to:
> 
> ```
> Expect(err).To(MatchError(ContainSubstring(fmt.Sprintf("Bundle.core.rukpak.io %q is invalid: metadata.name: Too long: may not be longer than 40", bundle.GetName()))))
> ```

We can update the string in the error message, but I'm having a hard time seeing why updating from 1.23 to 1.24 would cause the error message to change? In both cases, the Bundle creation fails due to the Bundle CRD schema validation right? The core webhook is no longer doing Bundle name length checks. 

Looking at the existing e2e test, we expect that something like `should be at most 40 chars long` is emitted, but I can't tell where that's coming from?

akihikokuroda:
The validation has been moved from the webhook to the [CRD definition patch](https://github.com/operator-framework/rukpak/blob/7b5c237860db77a6f7b68f293033671a63c520a3/manifests/apis/crds/patches/bundle_validation.yaml#L2). The message is not from the rukpak code.  It's from kubernetes (I guess).

test/e2e/webhook_test.go
```diff
@@ -82,7 +82,7 @@ var _ = Describe("bundle api validating webhook", func() {
 			Expect(err).To(WithTransform(apierrors.IsNotFound, BeTrue()))
 		})
 		It("should fail the long name bundle creation", func() {
-			Expect(err).To(MatchError(ContainSubstring(fmt.Sprintf("Invalid value: %q: metadata.name in body should be at most 40 chars long", bundle.GetName()))))
+			Expect(err).To(MatchError(ContainSubstring(fmt.Sprintf("Bundle.core.rukpak.io %q is invalid: metadata.name: Too long: may not be longer than 40", bundle.GetName()))))
```
timflannagan, line 85: Does this point to test brittleness? Should we update this to only check whether the error message is complaining about character length as a follow-up?

timflannagan (approved):

test/e2e/webhook_test.go
exdx: We could -- I'm honestly a little confounded on how this test is currently passing in CI:
* Initially, the bundle name length check was part of the core webhook, and it would return the old error (40 chars long)
* At some point, we moved the name validation to the CRD schema, and so the error returned changed to what we're seeing now (may not be longer than 40). 

I'm not sure why the test wasn't failing as part of that change. It does seem to be brittle, but as I said I'm not sure why we didn't run into issues with this earlier. I can open a follow-up.

exdx: #435

PR #422, opened 2022-06-07
*: Remove the operator-registry as a direct dependency

timflannagan:
Add an internal/registry package that manually vendors the
operator-registry type definitions we care about. Update that convert
package to reference that internal type definition, and remove the only
registry package we relied on.


internal/registry/registry.go
```diff
@@ -0,0 +1,21 @@
+package registry
```
timflannagan, line 1: Maybe I can rename this directory from "registry" to "operator-registry" to make it even more transparent? Or add a package godoc comment that explains these types were manually vendored from that repository if it makes it easier for our future selves?

exdx (approved):

internal/registry/registry.go
timflannagan: Updated in the last force push.

PR #424, opened 2022-06-07
docs: update provisioner docs to contain information about registry+v1 and aggregate concepts

tylerslaton:
## Summary
Doing two things here:
1. Aggregating some document concepts that can be shared in the root `provisioner` README.
2. Filling out the `registry` provisioner's README.

### DIFF
--- a/internal/provisioner/README.md
+++ b/internal/provisioner/README.md
@@ -1,6 +1,85 @@
-# provisioner
+# Provisioners
 
 The following subdirectories contain concrete Provisioner implementations.
 
 - [plain](plain/README.md) - provisions `plain+v0` k8s bundles
 - [registry](registry/README.md) - provisions `registry+v1` OLM bundles
+
+## Global Provisioner Concepts
+
+### Using provisioners to install bundle content
+RukPak comes built-in with the concept of provisioners. Provisioners are pluggable controllers that understand how to unpack and install
+`Bundle`s and `BundleInstance`s that specify they are in the provisioner's expected format.
+
+```mermaid
+graph TD
+    C[Provisioner]
+    C -->|Watches| D[Bundle]
+    C -->|Watches| E[BundleInstance]
+    D -->|References| F[Content]
+    E -->|Creates/Manages| F[Content]
+```
+
+Each provisioner has its own custom logic for how it goes about handling content. Using a specific provisioner is done by specifying the `provisionerClassName` property in the spec of either the `Bundle` or `BundleInstance` api. Here is an example of this concept in an embedded `BundleInstance`.
+
+```yaml
+apiVersion: core.rukpak.io/v1alpha1
+kind: BundleInstance
+metadata:
+  name: my-bundle-instance
+spec:
+  provisionerClassName: core.rukpak.io/my-provisioner # <-- Provisioner for the BundleInstance
+  template:
+    metadata:
+      labels:
+        app: my-bundle
+    spec:
+      source:
+        type: image
+        image:
+          ref: my-bundle@sha256:xyz123
+      provisionerClassName: core.rukpak.io/my-provisioner # <-- Provisioner for the Bundle
+```
+
+### Pivoting between bundle versions
+
+The `BundleInstance` API is meant to indicate the version of the bundle that should be active within the cluster.
+
+Given an existing BundleInstance resource in the cluster, which contains an embedded Bundle template for the
+`my-bundle-v0.0.1` bundle, you can modify the desired specification and a provisioner will automatically generate
+a new `my-bundle-v0.0.2` Bundle matching that template.
+
+When the new Bundle resource has been rolled out successfully, the old `my-bundle-v0.0.1` Bundle will be deleted from the cluster.
+
+Provisioners also continually reconcile the created content via dynamic watches to ensure that all
+resources referenced by the bundle are present on the cluster.
+
+### Make bundle content available but do not install it
+
+There is a natural separation between sourcing of the content and application of that content via two separate RukPak
+APIs, `Bundle` and `BundleInstance`. A user can specify a particular `Bundle` to be available in the cluster for
+inspection before any application of the resources. Given a `Bundle` resource named `my-bundle`, a provisioner
+will pull down and unpack the bundle to a tar.gz file that is saved using the provisioner's internal storage mechanism.
+
+By default, RukPak's built-in provisioners and other core resources are deployed into the `rukpak-system` namespace.
+
+The content of a bundle can be queried using the `status.contentURL`, assuming you have the necessary
+RBAC permissions to access bundle content.
+
+As an example, a client outside the cluster can view the file contents from a bundle named `my-bundle` by running
+the following script:
+
+```bash
+BUNDLE_NAME=my-bundle
+
+kubectl create sa fetch-bundle -n default
+kubectl create clusterrolebinding fetch-bundle --clusterrole=bundle-reader --serviceaccount=default:fetch-bundle
+export TOKEN=$(kubectl get secret -n default $(kubectl get sa -n default fetch-bundle -o jsonpath='{.secrets[0].name}') -o jsonpath='{.data.token}' | base64 -d)
+export URL=$(kubectl get bundle $BUNDLE_NAME -o jsonpath='{.status.contentURL}')
+kubectl run -qit --rm -n default --restart=Never fetch-bundle --image=curlimages/curl --overrides='{ "spec": { "serviceAccount": "fetch-bundle" }  }' --command -- curl -sSLk -H "Authorization: Bearer $TOKEN" -o - $URL | tar ztv
+kubectl delete clusterrolebinding fetch-bundle
+kubectl delete sa fetch-bundle -n default
+```
+
+Simplifying the process of fetching this bundle content (e.g. via a plugin) is on the RukPak roadmap.
+
--- a/internal/provisioner/registry/README.md
+++ b/internal/provisioner/registry/README.md
@@ -1,3 +1,282 @@
-# registry
+# Registry Provisioner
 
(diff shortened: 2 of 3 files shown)


internal/provisioner/README.md
```diff
@@ -1,6 +1,60 @@
...
+
+```mermaid
+graph TD
+    C[Provisioner]
+    C -->|Watches| D[Bundle]
+    C -->|Watches| E[BundleInstance]
+    D -->|References| F[Content]
+    E -->|Creates/Manages| F[Content]
+```
+### Make bundle content available but do not install it
```
exdx, line 19: I think starting with this section is a bit confusing to users -- is inspecting bundle content without applying it the first provisioner concept we want to show off? The plain doc had the use case of installing a Bundle via a BI as the primary use case, I think it's more intuitive.

internal/provisioner/README.md
```diff
@@ -1,6 +1,60 @@
...
+
+There is a natural separation between sourcing of the content and application of that content via two separate RukPak
+APIs, `Bundle` and `BundleInstance`. A user can specify a particular `Bundle` to be available in the cluster for
+inspection before any application of the resources. Given a `Bundle` resource named `my-bundle`, the plain provisioner
```
exdx, line 23: nit: since this is the top-level provisioner README, should we have references to the plain provisioner specifically here? I would expect we could treat this doc as more generic, and defer to the individual provisioner README docs for specifics.

internal/provisioner/README.md
```diff
@@ -1,6 +1,60 @@
...
+will pull down and unpack the bundle to a tar.gz file that is saved into a bundle cache directory mounted in the
+provisioner pods.
```
joelanford, line 25: Bundle storage is an implementation detail of the each provisioner. At this level, it might be better to say:
```suggestion
will pull down and unpack the bundle to a tar.gz file that is saved using the provisioner's internal storage mechanism.
```

internal/provisioner/README.md
```diff
@@ -1,6 +1,60 @@
...
+
+By default, `rukpak-system` is the configured namespace for deploying `plain` provisioner-related system resources.
```
joelanford, line 27:
```suggestion
By default, RukPak's built-in provisioners and other core resources are deployed into the `rukpak-system` namespace.
```

internal/provisioner/README.md
```diff
@@ -1,6 +1,60 @@
...
+```
+
+Simplifying the process of fetching this bundle content (e.g. via a plugin) is on the RukPak roadmap.
+
+### Pivoting between bundle versions
+
+The `BundleInstance` API is meant to indicate the version of the bundle that should be active within the cluster.
+
+Given an existing BundleInstance resource in the cluster, which contains an embedded Bundle template for the
+`my-bundle-v0.0.1` bundle, you can modify the desired specification and the plain provisioner will automatically generate
```
joelanford, line 54:
```suggestion
`my-bundle-v0.0.1` bundle, you can modify the desired specification and a provisioner will automatically generate
```

internal/provisioner/README.md
```diff
@@ -1,6 +1,60 @@
...
+a new `my-bundle-v0.0.2` Bundle matching that template.
+
+When the new Bundle resource has been rolled out successfully, the old `my-bundle-v0.0.1` Bundle will be deleted from the cluster.
+
+The provisioner also continually reconciles the created content via dynamic watches to ensure that all
+resources referenced by the bundle are present on the cluster.
```
joelanford, line 60:
```suggestion
Provisioners also continually reconcile the created content via dynamic watches to ensure that all
resources referenced by the bundle are present on the cluster.
```

internal/provisioner/README.md
joelanford, line 23:
```suggestion
inspection before any application of the resources. Given a `Bundle` resource named `my-bundle`, a provisioner
```

internal/provisioner/registry/README.md
```diff
@@ -1,3 +1,278 @@
-# registry
+# Registry Provisioner
 
-This provisioner was removed in commit 76fefcfcd1408d04b4eb7eb368d3eaab49565237.
+## Summary
+
+The `registry` provisioner is a core Rukpak provisioner that knows how to interact with bundles of a particular format.
+These `registry+v1` bundles, or `registry` bundles, are container images containing a set of static Kubernetes YAML
+manifests organized in the legacy Operator Lifecycle Manger (OLM) format. For more information on the `registry+v1` format, see
+the [OLM packaging doc](https://docs.openshift.com/container-platform/4.10/operators/understanding/olm-packaging-format.html).
```
joelanford, line 8: We should link to upstream docs here, if possible.

internal/provisioner/registry/README.md
```diff
@@ -1,3 +1,278 @@
...
+The `plain` provisioner will unpack the provided Bundle onto the cluster, and eventually make the content
+available on the cluster.
+
+```yaml
+apiVersion: core.rukpak.io/v1alpha1
+kind: BundleInstance
+metadata:
+  name: my-bundle-instance
+spec:
+  provisionerClassName: core.rukpak.io/registry
```
joelanford, line 36:
```suggestion
  provisionerClassName: core.rukpak.io/plain
```

internal/provisioner/registry/README.md
```diff
@@ -1,3 +1,278 @@
...
+  template:
+    metadata:
+      labels:
+        app: my-bundle
+    spec:
+      source:
+        type: image
+        image:
+          ref: my-bundle@sha256:xyz123
+      provisionerClassName: core.rukpak.io/plain
```
joelanford, line 46:
```suggestion
      provisionerClassName: core.rukpak.io/registry
```

internal/provisioner/registry/README.md
exdx, line 8: nit: let's avoid any downstream documentation references in the upstream project. Could we point to https://olm.operatorframework.io/docs/tasks/creating-operator-manifests/ instead?

internal/provisioner/registry/README.md
```diff
@@ -1,3 +1,278 @@
...
+### Install and apply a specific version of a `registry+v1` bundle
+
+> :warning: Anyone with the ability to create or update BundleInstance objects can become cluster admin. It's important
+> to limit access to this API via RBAC to only those that explicitly require access, as well as audit your bundles to
+> ensure the content being installed on-cluster is as-expected and secure.
+
+The `registry` provisioner can convert and make available a specific `registry+v1` bundle as a `plain+v0` in the cluster.
+
+Simply create a `BundleInstance` resource that contains the desired specification of a Bundle resource.
+The `plain` provisioner will unpack the provided Bundle onto the cluster, and eventually make the content
```
joelanford, line 27:
```suggestion
The `registry` provisioner will unpack the provided Bundle onto the cluster, and the `plain` provisioner will eventually make the content
```

internal/provisioner/registry/README.md
```diff
@@ -1,3 +1,278 @@
...
 
-This provisioner was removed in commit 76fefcfcd1408d04b4eb7eb368d3eaab49565237.
+## Summary
+
+The `registry` provisioner is a core Rukpak provisioner that knows how to interact with bundles of a particular format.
+These `registry+v1` bundles, or `registry` bundles, are container images containing a set of static Kubernetes YAML
+manifests organized in the legacy Operator Lifecycle Manger (OLM) format. For more information on the `registry+v1` format, see
+the [OLM packaging doc](https://docs.openshift.com/container-platform/4.10/operators/understanding/olm-packaging-format.html).
+
+The `registry` provisioner is able to convert a given `registry+v1` bundle onto a cluster in the `plain+v0` format. Instantiation of the
```
joelanford, line 10: Somewhere in here we might want to list the limitations of the conversion. We don't support _all_ registry+v1 bundles.

internal/provisioner/registry/README.md
```diff
@@ -1,3 +1,278 @@
...
+bundle is then handled by the `plain` provisioner in order to make the content available in the cluster. It does so by reconciling `Bundle` 
+types that have the `spec.provisionerClassName` field set to `core.rukpak.io/registry`. This field must be set to the correct provisioner 
+name in order for the `registry` provisioner to see and interact with the bundle. Creating a `BundleInstance` of that `Bundle` would then 
+require that you have set the `BundleInstance` property `spec.provisionerClassName` field to `core.rukpak.io/plain`.
+
```
exdx, line 15: Might be helpful to drop an example BI spec right here? The explanation above is hard to visualize. Or maybe we just could add a diagram of some kind?

joelanford (commented):
Awesome. Thanks for taking the time to do this!

exdx (approved):
LGTM, just some minor suggestions

internal/provisioner/README.md
tylerslaton: The reference to plain here should definitely not be there. Fixing all of the reference to it in this top-level document.

internal/provisioner/registry/README.md
tylerslaton: I put a link to the section of the document where we start defining this since it felt alike we could reuse some text here.

internal/provisioner/README.md
exdx: That section looks really good. I would argue that we should move `Make bundle content available but do not install it` down even further and move `Pivoting between bundle versions` after the new section, since that concept is more important. This was just an oversight in the original doc I wrote.

tylerslaton: Gotcha, I swapped those two now!

internal/provisioner/registry/README.md
tylerslaton: I left a new note that outlines this particular issue.

PR #426, opened 2022-06-08
add private git repository access doc

akihikokuroda:

Doc updates for #414 and #407


docs/git-bundles.md
```diff
@@ -21,8 +21,86 @@ When creating a Bundle from a git source, a reference to a particular commit, ta
...
+
+1. Create the secret
+
+```sh
+echo -n 'user name' > username.txt
+echo -n 'password' > password.txt
+kubectl create secret generic gitsecret --type "kubernetes.io/basic-auth" --from-file=username=./username.txt --from-file=password=./password.txt -n rukpak-system
+```
+
+2. Create a private repository (private-registry/combo) and copy operator-framework/combo contents into it
```
exdx, line 45: nit: I'm not sure the user should have to create a new private repository and copy combo in order to show how private git repo access works. I would maybe just say something like
`Find an existing private git repository or create a new one that is private by default`

docs/git-bundles.md
```diff
@@ -21,8 +21,86 @@ When creating a Bundle from a git source, a reference to a particular commit, ta
...
+kind: Bundle
+metadata:
+  name: combo-private
+spec:
+  source:
+    type: git
+    git:
+      ref:
+        branch: main
+      repository: https://github.com/private-registry/combo
```
exdx, line 61: nit: registry is for container images, in this case it's github so I would maybe use something like <username> instead.

docs/git-bundles.md
```diff
@@ -21,8 +21,86 @@ When creating a Bundle from a git source, a reference to a particular commit, ta
...
+### Example steps for ssh URL
+
+1. Create the secret
+
+```sh
+cat ~/.ssh/known_hosts | grep 'github' > ssh_knownhosts.txt
+kubectl create secret generic gitsecret --type "kubernetes.io/ssh-auth" --from-file=ssh-privatekey=~/.ssh/id_rsa --from-file=ssh-knownhosts=./ssh_konwnhosts.txt -n rukpak-system
+```
+
+2. Create a private repository (private-registry/combo) and copy operator-framework/combo contents into it
```
exdx, line 79: Same as above.

docs/git-bundles.md
```diff
@@ -21,8 +21,86 @@ When creating a Bundle from a git source, a reference to a particular commit, ta
...
+kind: Bundle
+metadata:
+  name: combo-private
+spec:
+  source:
+    type: git
+    git:
+      ref:
+        branch: main
+      repository: git@github.com:private-registry/combo
```
exdx, line 95: Same as above.

exdx (approved):
Nice, well written. Just some small suggestions and then this is good to go.

docs/git-bundles.md
akihikokuroda: It seems better.  I'll change it.  Thanks!

docs/git-bundles.md
akihikokuroda: I'll change it.  Thanks!

docs/git-bundles.md
akihikokuroda: I'll change it.  Thanks!

Issue #430, opened 2022-06-09
Add E2E tests for `registry+v1` bundle


tylerslaton:
An e2e test exists that verifies that a registry+v1 bundle can be unpacked by the registry+v1 provisioner, converted to a plain bundle, and then installed by the plain provisioner

exdx:
I think that #432 does fulfill the AC of this ticket -- will close this for now.

PR #432, opened 2022-06-14
e2e add registry provisioner bundleinstance test

### LINKED ISSUE #427: Add e2e tests for the content http server of the plain provisioner
Add e2e tests that exercise the content http server of the plain provisioner (including authorization via kube-rbac-proxy).

Follow-up work from #311

akihikokuroda:
Adding one bundle with the registry provisioner and one bundleinstane with the plain provisioner.


test/e2e/plain_provisioner_test.go
```diff
@@ -1186,6 +1187,66 @@ var _ = Describe("plain provisioner bundleinstance", func() {
...
+			}
+			err := c.Create(ctx, bi)
+			Expect(err).To(BeNil())
+		})
+		AfterEach(func() {
+			By("deleting the testing BI resource")
+			Expect(c.Delete(ctx, bi)).To(BeNil())
+		})
+
+		It("should rollout the bundle contents successfully", func() {
```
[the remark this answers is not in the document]
akihikokuroda: I thought this might be enough for #427.  The granular tests may be coved in #430.  WDYT?

test/e2e/plain_provisioner_test.go
```diff
@@ -26,6 +26,7 @@ import (
 
 	rukpakv1alpha1 "github.com/operator-framework/rukpak/api/v1alpha1"
 	plain "github.com/operator-framework/rukpak/internal/provisioner/plain/types"
+	registry "github.com/operator-framework/rukpak/internal/provisioner/registry/types"
```
tylerslaton, line 29: Should this be inside of a new file? Something like:

`registry_provisioner_test.go`

tylerslaton (commented):
This looks good to me and I'm fine to merge this as is. However, I will echo that we should expand the cases that we're covering for the registry provisioner before closing the two issues linked in this PR.

Side-note: I don't think that the two issues are direct duplicates but are certainly closely related.

test/e2e/plain_provisioner_test.go
akihikokuroda: I make the new file.  It was just one test so I thought a new file might not necessary but we probably add more tests for the registry bundle so I create the new file.  Thanks!

test/e2e/retistry_provisioner_test.go
```diff
@@ -0,0 +1,78 @@
+package e2e
```
exdx, line 1: nit: there is a typo in the name of the test file. It should be `registry_provisioner_test`

test/e2e/plain_provisioner_test.go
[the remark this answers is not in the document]
exdx: I'm not sure this closes #427 but I think it's fine for #430 (for now). For #427 I think we want to do some more indepth checks around auth (where if you don't have the right permissions, writing to that endpoint would fail?)

test/e2e/retistry_provisioner_test.go
akihikokuroda: Good catch!  Thanks!

akihikokuroda:
I merge this without closing related issues #427 and #430.

Issue #331, opened 2022-05-06
Revisit the BundleInstance API's UX after the latest API overhaul


timflannagan:
Follow-up to #158 which overhauled the BundleInstance API to support the dynamic creation of Bundle contents.

Goal: revisit the status.installedBundleName and the custom printer columns.

Currently, the UX around the BI's printer column output can be confusing, especially in the case where a generated BundleA was successfully created/unpacked/installed, and the BI was updated with a different desired configuration, and the generated BundleB fails to unpack/install. In this case, the printer columns signal that there's still a successful installation as BundleA is still active, which can be confusing.

One implementation may be centered around using the Deployment/ReplicaSet status fields to derive printer columns (e.g. Ready/Available/etc.), and refactor the BI's printer columns to use boolean types to help communicate the current state of a resource vs. relying on string values which are difficult to accurately portray the full picture of the current state.

exdx:
UX may include a numeric status: 0/1 or 1/2 available during a pivot that fails. 

The UX may be informed by a higher level component (like deppy) handling this status and keeping the rukpak layer straightforward.

timflannagan:
Related to https://github.com/operator-framework/rukpak/issues/403 which can be a short term UX win.

PR #437, opened 2022-06-24
api: Change BundleInstance status.installedBundleName to status.activeBundle

### LINKED ISSUE #403: Rename BundleInstance status.installedBundleName field
The `status.installedBundleName` field on the BI feels overly verbose, and now that dynamic Bundles exist, it may be sufficient to have this field be something more straightforward. There was broad agreement about potentially changing this field.

Possible candidates:
* status.bundle
* status.bundleName

exdx:
This change simplifies the BI status and improves the UX in the world
of dynamic bundles, when the bundle is known when creating the BI.

Closes #403 -- see #331 for more context.

### DIFF
--- a/api/v1alpha1/bundledeployment_types.go
+++ b/api/v1alpha1/bundledeployment_types.go
@@ -64,14 +64,14 @@ type BundleTemplate struct {
 
 // BundleDeploymentStatus defines the observed state of BundleDeployment
 type BundleDeploymentStatus struct {
-	Conditions          []metav1.Condition `json:"conditions,omitempty"`
-	InstalledBundleName string             `json:"installedBundleName,omitempty"`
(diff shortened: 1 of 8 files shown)


tylerslaton (approved):

akihikokuroda (approved):
LGTM

exdx:
> I think I'm still in the "Active Bundle" camp, but we may just need fresh eyes here. I think the "Bundle" printer column is a bit ambiguous. We're basically just outputting the Bundle reference now, right?

Yes, we're outputting the name of the bundle under `status.bundle` in this proposed change. It associates the particular bundle with the BD. I agree it's a bit ambiguous, but the UI is nice. 

I think "Active Bundle" is probably nicer, since pivoting is a key feature of rukpak, and it helps illustrate that more

PR #443, opened 2022-06-29
feat: Add local source type

### LINKED ISSUE #92: Support referencing and mounting local Bundle content

exdx:
Add a local SourceType with support for a configmap. This enables users
to create a Bundle that points to a configmap for its underlying
resources.

Closes #92


internal/source/local.go
```diff
@@ -0,0 +1,79 @@
...
+			return nil, fmt.Errorf("creating filesystem from configmap: %s", err)
+		}
+		_, err = file.Write([]byte(contents))
+		if err != nil {
+			return nil, fmt.Errorf("creating filesystem from configmap: %s", err)
+		}
+	}
+
+	// Add an ownerref to the configmap based on the Bundle
+	var ownerref = metav1.NewControllerRef(bundle, bundle.GroupVersionKind())
```
tylerslaton, line 57: Could you explain for me why we want Rukpak to now own the ConfigMap? In essence this means that if the Bundle is deleted then the ConfigMap gets deleted as well. Should that always be the case or would we prefer that to be configurable somehow? I can imagine a scenario where a user would not expect the ConfigMap they created to be deleted by deleting the Bundle.

Another problem that may arise from this is if two Bundles share the same ConfigMap. What happens in that scenario if we delete that ConfigMap?

test/e2e/plain_provisioner_test.go
```diff
@@ -735,7 +737,107 @@ var _ = Describe("plain provisioner bundle", func() {
...
+				},
+			}
+			err = c.Create(ctx, bundle)
+			Expect(err).To(BeNil())
+		})
 
+		AfterEach(func() {
+			err := c.Delete(ctx, bundle)
+			Expect(client.IgnoreNotFound(err)).To(BeNil())
+			err = c.Delete(ctx, configmap)
```
tylerslaton, line 802: Would this `configmap` not get deleted via the cascade anyway?

tylerslaton (commented):
This looks good, great job getting this out so quickly. Had a few questions I was hoping you could answer before I send out an approval.

test/e2e/plain_provisioner_test.go
exdx: That's a good point though -- if we expect the configmap to be deleted, maybe we shouldn't delete it explicitly in the AfterEach?

internal/source/local.go
exdx: I'm looking at this for the perspective of "this Bundle binds this configmap, so tie the configmap's lifecycle to that of the Bundle". It helps keep the cluster state the same before/after a BD is applied. I think if we document the fact that the configmap is owned by the Bundle users will expect this behavior. We could make it configurable further on if we find out that users would like their configmaps to stay around. 

Good question about two Bundles sharing the same configmap -- currently that's not really supported since its implied each Bundle is backed by its own unique content source. I'm not sure I see the value of having two cluster-scoped Bundles that both reference the same backing content.

tylerslaton: > I'm not sure I see the value of having two cluster-scoped Bundles that both reference the same backing content.
I don't either, to be sure. I just want to make sure we've thought of it and have an official stance before putting this out.

Thanks for the answer, I feel better about this to the point I'll go ahead and mark this as resolved. Its good to document this so that it is expected.

docs/local-bundles.md
```diff
@@ -0,0 +1,51 @@
+# Local Bundles
+
+## Summary
+
+A Bundle can reference content from an in-cluster resource instead of a remote container image or a git repository by
+using the`local` source type in the Bundle manifest. This enables one to easily source content locally without external
+repositories/registries. Sourcing from a local configmap is currently supported, with support for additional content
+sources like a PV is on the roadmap.
```
tylerslaton, line 8:
```suggestion
sources, like a PV, on the roadmap.
```

docs/local-bundles.md
```diff
@@ -0,0 +1,51 @@
...
+
+The configmap can be created with the command:
+
+```bash
+kubectl create configmap <configmap name> --from-file=<manifests directory>
+```
+
+> Note: Once a configmap is referenced by a Bundle, the configmap will have an owner reference placed on it that points
+> to the Bundle. As a result, when the Bundle is removed, the configmap will also be removed. This ensures that the cluster is
+> in the same state before and after installing content.
```
tylerslaton, line 24: 🙌🏻

test/e2e/plain_provisioner_test.go
```diff
@@ -735,7 +737,94 @@ var _ = Describe("plain provisioner bundle", func() {
...
+			err = c.Create(ctx, bundle)
+			Expect(err).To(BeNil())
+		})
+
+		AfterEach(func() {
+			err := c.Delete(ctx, bundle)
+			Expect(client.IgnoreNotFound(err)).To(BeNil())
+		})
 
+		It("Can create and unpack the bundle successfully", func() {
```
tylerslaton, line 803: Would it make sense to add a test case for defining what should happen if the `Bundle` inside of the `ConfigMap` is invalid?

tylerslaton (approved):
Nice, this looks great. Had a few small nits but nothing blocking.

test/e2e/plain_provisioner_test.go
exdx: Actually, since I'm BDC this week, I'd rather go ahead and merge this as-is and track follow-up e2e tests as a separate issue. This would avoid any potential merge conflicts.

exdx: https://github.com/operator-framework/rukpak/issues/451

PR #447, opened 2022-07-05
refactor: deduplicate provisioner code

### LINKED ISSUE #431: Deduplicate code used across provisioners
- `main.go` files
    - plain/main.go 
    - registry/main.go
- `bundle_controller` files 
    - plain/controllers/bundle_controller.go 
    - registry/controllers/bundle_controller.go

tylerslaton:
## Summary
Doing a few things here:
1. Adding a new "common" sub package into the provisioner package for shared code
2. Creating a singular main for the provisioner package
3. Updating the manifests accordingly

These changes have been logically broken up in into commits for ease of review.

## Explanation
The intention here is to refactor the code-base in such a way that provisioners share their duplicate code. To do this, I removed each individual provisioner package's `main` in favor of a singular unified `main` which is found in the root of the `provisioner` package. This main binary is configurable via flags and arguments to specify which `provisioner` we should run. This results in calls to appropriate start-up function for the specific `provisioner` which enables its unique features.

By doing this, we should be able to bootstrap up new provisioner very quickly in the future while also deduplicating our code.

*Note*: I am currently working through any issues that may arise through the E2E suite but this reflects my overall approach to this problem.

Closes #431


exdx (commented):
Quick feedback

internal/provisioner/main.go
```diff
@@ -68,18 +77,19 @@ func main() {
...
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.StringVar(&systemNamespace, "system-namespace", "rukpak-system", "Configures the namespace that gets used to deploy system resources.")
 	flag.StringVar(&unpackImage, "unpack-image", "quay.io/operator-framework/rukpak:latest", "Configures the container image that gets used to unpack Bundle contents.")
-	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
-		"Enable leader election for controller manager. "+
-			"Enabling this will ensure there is only one active controller manager.")
+	flag.BoolVar(&enableLeaderElection, "leader-elect", false, "Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.")
 	flag.BoolVar(&rukpakVersion, "version", false, "Displays rukpak version information")
 	flag.StringVar(&storageDirectory, "storage-dir", storage.DefaultBundleCacheDir, "Configures the directory that is used to store Bundle contents.")
+	flag.StringVar(&httpBindAddr, "provisioner", "plain", "The provisioner to start.")
```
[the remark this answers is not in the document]
tylerslaton: I don't think that it would, right? This just defaults to `plain` if the flag is not defined. Otherwise, like in the case of the `registry` provisioner, you'll just pass which provisioner you want.

manifests/provisioners/registry/resources/deployment.yaml
```diff
@@ -38,13 +38,14 @@ spec:
...
           imagePullPolicy: IfNotPresent
-          command: ["/registry"]
+          command: ["/provisioner"]
           args:
             - "--unpack-image=quay.io/operator-framework/rukpak:latest"
             - "--storage-dir=/var/cache/bundles"
             - "--http-bind-address=127.0.0.1:8080"
             - "--http-external-address=https://$(REGISTRY_PROVISIONER_SERVICE_NAME).$(REGISTRY_PROVISIONER_SERVICE_NAMESPACE).svc"
             - "--bundle-ca-file=/etc/pki/tls/ca.crt"
+            - "--provisioner=registry"
```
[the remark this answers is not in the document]
tylerslaton: That's fair. I think that I could go either way on this. Having some default feels somewhat natural to me but I could see how there may not be a logical default that matches everyone's use case.

internal/provisioner/registry/controllers/bundle_controller.go
```diff
@@ -137,136 +132,46 @@ func (r *BundleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 
 	unpackResult, err := r.Unpacker.Unpack(ctx, bundle)
 	if err != nil {
-		return ctrl.Result{}, updateStatusUnpackFailing(&u, fmt.Errorf("source bundle content: %v", err))
+		return ctrl.Result{}, base.UpdateStatusUnpackFailing(&u, fmt.Errorf("source bundle content: %v", err))
```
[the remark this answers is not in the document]
tylerslaton: That sounds like a good call to me.

akihikokuroda:
Nice refactor!  By seeing this,  I wonder why there is a separated registry provisioner and the plain provisioner with 2 different sources (git and image)?  What would our recommendation be to have multiple provisioners vs one provisioner with different sources?

manifests/provisioners/registry/resources/deployment.yaml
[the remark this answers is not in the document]
tylerslaton: Updated in the most recent commit to take an arg for the provisioner to start up

internal/provisioner/plain/plain.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/operator-framework/rukpak/internal/provisioner/plain/controllers"
+	"github.com/operator-framework/rukpak/internal/source"
+	"github.com/operator-framework/rukpak/internal/storage"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+
+	crfinalizer "sigs.k8s.io/controller-runtime/pkg/finalizer"
+)
+
+// TODO: Minimize signature
+func SetupProvisionerWithManager(ns string, mgr manager.Manager, bundleStorage storage.Storage, bundleFinalizers crfinalizer.Finalizers, unpacker source.Unpacker) error {
```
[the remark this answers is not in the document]
tylerslaton: So this gets into some `Go` intricacies. Since `manager.Manager` is an interface, a pointer to that interface isn't accepted here. We can't even call the interface's methods. IIRC, an interface is, under the hood, a pointer to its implementation so we are doing what you're asking already. 

Referenced: https://stackoverflow.com/questions/44370277/type-is-pointer-to-interface-not-interface-confusion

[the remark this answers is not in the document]
exdx: Interesting, thanks for clearing that up.

internal/provisioner/common/bundle.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	unpackResult, err := r.Unpacker.Unpack(ctx, bundle)
+	if err != nil {
+		return ctrl.Result{}, UpdateStatusUnpackFailing(&u, fmt.Errorf("source bundle content: %v", err))
+	}
+
+	if r.ParseUnpackState == nil {
+		r.ParseUnpackState = defaultParseUnpackStateFunc
+	}
```
joelanford, line 143: It would probably be better not to mutate the state of `r` during reconcile (this function might be called concurrently). Perhaps we could have a function that initializes `ProvisionerBundleReconciler` (e.g. `NewProvisionerBundleReconciler`) where we could do this check/set the default there, prior to the reconciler actually being used?

internal/provisioner/common/bundle.go
```diff
@@ -0,0 +1,160 @@
...
+func (r *ProvisionerBundleReconciler) SetupWithManager(mgr ctrl.Manager, predicates ...predicate.Predicate) error {
+	l := mgr.GetLogger().WithName("controller.bundle")
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&rukpakv1alpha1.Bundle{}, builder.WithPredicates(
+			append(predicates, util.BundleProvisionerFilter(r.ProvisionerID))...,
+		)).
+		// The default unpacker creates Pod's ownerref'd to its bundle, so
+		// we need to watch pods to ensure we reconcile events coming from these
+		// pods.
+		Watches(&crsource.Kind{Type: &corev1.Pod{}}, util.MapOwneeToOwnerProvisionerHandler(context.TODO(), mgr.GetClient(), l, r.ProvisionerID, &rukpakv1alpha1.Bundle{})).
```
joelanford, line 158: The watches required by a reconciler are at least partially dependent on the unpacker implementation, which is configurable on the reconciler. With that in mind, I'd lean toward making `SetupWithManager` implementation specific.

internal/provisioner/main.go
```diff
@@ -14,16 +16,13 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package main
-
```
joelanford, line 18: Hmm. Don't we usually keep the copyright notice at the very top of the file? I'm curious if there's a specific motivation for this change.

joelanford (commented):
So following up from my first review, I think I'm still on the "let's have separate controller files" train.

What if we:
1. Have a few initializer functions `plain/controllers.NewBundleReconciler` and `registry/controllers.NewBundleReconciler` that know how to initialize a specific provisioner (e.g. set the provisionerID and the parseUnpackState function). It looks like you mostly have this with `SetupProvisionerWithManager` functions, but I think I'd move things around such that `plain/controllers/bundle_controller.go` is put back, contains those initializers, and then just delegates to the common functions where necessary.
2. The provisioner-specific `Reconciler` implementations embed the common `ProvisionerBundleReconciler` and inherit its fields. The provisioner-specific `Reconcile` function could just directly call the `Reconcile` implementation of the embedded common `ProvisionerBundleReconciler`.
3. The provisioner-specific `Reconciler` implementations provide their own `SetupWithManager` implementations to account for any setup differences that involve the manager (e.g. the watches I mentioned)

internal/provisioner/common/bundle.go
tylerslaton: Leaving this as unresolved so you can see how you feel about this most recent implementation coming through now.

tylerslaton:
@joelanford I believe that I addressed your over-arching concerns. Mind giving this PR another look?

internal/provisioner/common/bundle.go
```diff
@@ -0,0 +1,137 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/finalizer"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	rukpakv1alpha1 "github.com/operator-framework/rukpak/api/v1alpha1"
+	"github.com/operator-framework/rukpak/internal/source"
+	"github.com/operator-framework/rukpak/internal/storage"
+	updater "github.com/operator-framework/rukpak/internal/updater/bundle"
+)
+
+type ParseUnpackStateFunc func(context.Context, *source.Result, *updater.Updater, *rukpakv1alpha1.Bundle, storage.Storage) (ctrl.Result, error)
```
exdx, line 22: nit: newline after this type definition

internal/provisioner/main.go
```diff
@@ -91,6 +96,19 @@ func main() {
 		os.Exit(0)
 	}
 
+	args := flag.Args()
+	if len(args) != 1 {
```
exdx, line 100: TODO: this probably should be overhauled to use cobra, so we can enforce these types of semantics before runtime

internal/provisioner/main.go
```diff
@@ -179,36 +197,17 @@ func main() {
...
-	if err = (&controllers.BundleDeploymentReconciler{
-		Client:             mgr.GetClient(),
-		Scheme:             mgr.GetScheme(),
-		BundleStorage:      bundleStorage,
-		ReleaseNamespace:   ns,
-		ActionClientGetter: helmclient.NewActionClientGetter(cfgGetter),
-	}).SetupWithManager(mgr); err != nil {
-		setupLog.Error(err, "unable to create controller", "controller", rukpakv1alpha1.BundleDeploymentKind)
+	// Call the relevant provisioner setup function
+	if err := setupProvisioner[provisioner](ns, mgr, bundleStorage, bundleFinalizers, unpacker); err != nil {
```
exdx, line 207: I wonder if, for the sake of clarity, we can attach the reconciler attributes in-line, as before, since it makes it really clear what's happening. This method does save space, but at the expense of being able to see what's happening.

exdx (approved):

timflannagan (commented):
Do we want a single executable for the plain, registry, etc. provisioners? Does that imply those are core provisioners long term? What happens if an admin/user doesn't need a registry+v1 provisioner, and would just like to install the plain provisioner?

exdx:
I think the approach we have now with separate deployments is scalable enough in the short-term. I believe both registry and plain are core provisioners that will be pre-packaged with rukpak as part of a 1.0 release. If a user wouldn't like one of the provisioners to run, they can always scale down the deployment replicas (we could even have a `rukpakctl disable <provisioner>` that makes this simple.

timflannagan:
Another question then: what happens if we need to provide a UX around which provisioners are present in the cluster? Would that influence individual executables vs. a single executable that has flags for toggling each provisioner implementation?

exdx:
I think if we needed something like that, we could introduce a new API, something like `RukPakConfig` that had that information encoded in it?

The reason I hesitate regarding the single binary approach is because the plain and registry provisioners are actually intertwined right now right? We parse the registry+v1 bundle, transform it into a plain bundle under the hood, and then deploy it via the plain BD controller? Since there's not a clean separation, it could be surprising UX for users to disable the plain provisioner and see their registry+v1 bundles failing to deploy. 

The other alternative, disabling the registry+v1 provisioner, seems logical, but there is no ecosystem of plain bundles at the moment, so the motivation behind using only plain is less clear.

timflannagan:
> The reason I hesitate regarding the single binary approach is because the plain and registry provisioners are actually intertwined right now right? We parse the registry+v1 bundle, transform it into a plain bundle under the hood, and then deploy it via the plain BD controller? Since there's not a clean separation,

I think from my perspective, we have that clean separation today, it's just when rukpak users want to source plain bundles. Where we see that separation start to leak is that situation you described above, where our registry+v1 provisioner is half-baked and relies on the plain provisioner for the application and management of bundle contents. It's unclear whether this behavior is something we'll continue to support in the intermediate term though.

>  it could be surprising UX for users to disable the plain provisioner and see their registry+v1 bundles failing to deploy.

I definitely agree this is a surprising UX, but we'd still have this problem today regardless of a single executable vs. multiple, right? I think we've landed on needing an auxiliary visibility API in the meantime though.

exdx:
Yup, sounds good -- opened #468

joelanford:
Every now and then, I do get the urge to throw all of the core controllers/provisioners into a single process. That would likely have a relatively small impact on CPU (just very marginal improvements), but it would be a bigger win for memory because we could:
1. reduce the overhead of another go process
2. use a shared cache, which would mean significant savings of stored bundles, bundle deployments and (eventually) provisioner class objects

It would also have the benefit of being slightly easier to manage (less kustomize stuff, fewer main.go files, shorter Makefile, etc.)

internal/provisioner/main.go
tylerslaton: That is fair, Joe mentioned that and I'll work on adding it in.

tylerslaton:
So some clarifying statements:

1. There are still separate Deployments for each provisioner, they just pass a different flag for which provisioner they want out of our binary. This is nice because we can add a batch of "standard" provisioners with a single image and binary that is easily configurable. In essence this is how the repo's Dockerfile operates today; one image, several binaries. 
2. The main intent behind having a single unified binary is that all provisioners we define in this repo can share code that they will all use. Going back to the original intent of this PR, we're attempting to templates the code that is shared and allow extensibility with code that isn't. In the current implementation, a large portion of code is duplicated which can make onboarding provisioners a bit frustrating moving forward.

> @timflannagan I definitely agree this is a surprising UX, but we'd still have this problem today regardless of a single executable vs. multiple, right? I think we've landed on needing an auxiliary visibility API in the meantime though.

Yes, this PR doesn't really create, help, or fix that issue. That issue is mainly just with the current implementation of the `registry` provisioner.

timflannagan:
Yeah, I brought this conversation off track. I'm fine with a single executable after reading the points that Joe outlined. I think we should be careful about scope creep here though, so maybe we should tackle consolidating all the provisioners into a single process first, then work towards deduplicating the shared code. Thoughts?

internal/provisioner/main.go
exdx: I think it's a good candidate for a follow-up

akihikokuroda (approved):

tylerslaton:
> Yeah, I brought this conversation off track. I'm fine with a single executable after reading the points that Joe outlined. I think we should be careful about scope creep here though, so maybe we should tackle consolidating all the provisioners into a single process first, then work towards deduplicating the shared code. Thoughts?

@timflannagan I would be fine with this, I could just separate out this PR into two commits if it would make this more digestible. That being said, I think that the changes here do fit under the deduplication of code umbrella.

tylerslaton:
As per a recent refinement meeting, I'm going to be overhauling this PR to remove the `common` for now and instead focus on combining the `main.go`'s like it is now. In addition to this, we are going to make it so only a single process runs on cluster and contains all provisioners. Putting on hold for now.

timflannagan:
@tylerslaton It might be worth moving that to it's own branch/PR and letting this once sit so we have a historical record.

tylerslaton:
> As per a recent refinement meeting, I'm going to be overhauling this PR to remove the common for now and instead focus on combining the main.go's like it is now. In addition to this, we are going to make it so only a single process runs on cluster and contains all provisioners. Putting on hold for now.

I'm actually going to go ahead and close this. We can reopen it at a later date if it becomes pertinent. The relevant changes from this branch will be carried over into a new PR.

PR #449, opened 2022-07-05
API changes for bundle conversion CLI

ryantking:
This PR responds to the comments left by @joelanford  on https://github.com/operator-framework/rukpak/pull/396.

The big unknown is where to put the manifest directory restriction. For example, in a `registry+v1` bundle, manifests must be in a flat top level directory called "manifests." Since `plain+v0` bundles have a similar restriction, should the `manifest.FS` enforce that (in a configurable way) or should it be up to the bundle implementations entirely. I'm leaning towards moving it entirely to bundle implementations. Even though it introduces some mild duplication (nothing a common lib can't solve), it seems more correct to have the bundles totally own their specs. I just want confirmation before doing the work.

The other issue is that converting to a new `fstest.MapFS` is problematic since `fs.WalkDir` does not pickup any new files. I've tried adding a dummy directory, but that did not solve it. I see a few options here:

1. Find an off the shelf`fstest.MapFS` that works the way needed.
2. Try to code around it using `ReadDir` and the like, but that gets dicey if we remove the "single flat directory" restriction from the `manifest.FS` type.
3. Completely replace `fs.FS` with a more feature-filled off the shelf one like `afero`. Not a fan of this option, but putting it here for purposes of exhausting all our options.

### DIFF
--- a/pkg/bundle/plain/v0/bundle.go
+++ b/pkg/bundle/plain/v0/bundle.go
@@ -3,12 +3,12 @@ package v0
 import (
 	"io/fs"
 
-	"github.com/operator-framework/rukpak/pkg/bundle"
+	"github.com/operator-framework/rukpak/pkg/manifest"
 )
 
 // Bundle holds the contents of a plain+v0 bundle.
 type Bundle struct {
-	bundle.FS
+	manifest.FS
 
 	// external settings
 	installNamespace string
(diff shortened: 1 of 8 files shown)


joelanford:
> should it be up to the bundle implementations entirely

This is where I'm at.

PR #456, opened 2022-07-08
include bundle unpack failure msg in the bundledeployment condition msg 

### LINKED ISSUE #444: When a bundle fails to unpack, the reason should be surfaced in bundleDeployment
Goal: Re-visit the UX around the HasValidBundle and Installed types, and the printer columns.

The BundleDeployment controllers first try to unpack the desired bundle source before applying those unpacked source manifests. In the case that the desired bundle fails to unpack, the provisioner implementations surface a HasValidBundle status conditions type. When querying for BundleDeployment status locally, the "INSTALL STATE" printer column UX is empty which doesn't properly relay the current state.

akihikokuroda:
Closes #444 

BundleDeployment condition has this with this PR.
```
    conditions:
    - lastTransitionTime: "2022-07-08T16:37:24Z"
      message: 'Failed to unpack the Bundle: fail-6f96958c8b, reason: get objects
        from bundle manifests: read ".gitignore": error unmarshaling JSON: while decoding
        JSON: json: cannot unmarshal string into Go value of type map[string]interface
        {}'
      reason: UnpackFailed
      status: "False"
      type: HasValidBundle
```

for the Bundle condition
```
    conditions:
    - lastTransitionTime: "2022-07-08T16:37:24Z"
      message: 'get objects from bundle manifests: read ".gitignore": error unmarshaling
        JSON: while decoding JSON: json: cannot unmarshal string into Go value of
        type map[string]interface {}'
      reason: UnpackFailed
      status: "False"
      type: Unpacked
```

### DIFF
--- a/internal/provisioner/plain/controllers/bundledeployment_controller.go
+++ b/internal/provisioner/plain/controllers/bundledeployment_controller.go
@@ -133,6 +133,9 @@ func (r *BundleDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Req
 			reason = rukpakv1alpha1.ReasonUnpackFailed
 			status = metav1.ConditionFalse
 			message = fmt.Sprintf("Failed to unpack the %s Bundle", bundle.GetName())
+			if c := meta.FindStatusCondition(bundle.Status.Conditions, rukpakv1alpha1.TypeUnpacked); c != nil {
+				message = fmt.Sprintf("%s: %s", message, c.Message)
+			}
 		}
 		u.UpdateStatus(
 			updater.EnsureCondition(metav1.Condition{


internal/provisioner/plain/controllers/bundledeployment_controller.go
```diff
@@ -132,7 +132,12 @@ func (r *BundleDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		if bundle.Status.Phase == rukpakv1alpha1.PhaseFailing {
 			reason = rukpakv1alpha1.ReasonUnpackFailed
 			status = metav1.ConditionFalse
-			message = fmt.Sprintf("Failed to unpack the %s Bundle", bundle.GetName())
+			c := meta.FindStatusCondition(bundle.Status.Conditions, rukpakv1alpha1.TypeUnpacked)
+			bundleMessage := "No unpacked condition"
+			if c != nil {
+				bundleMessage = c.Message
+			}
+			message = fmt.Sprintf("Failed to unpack the Bundle: %s, reason: %s", bundle.GetName(), bundleMessage)
```
timflannagan, line 140: nit: I think we can drop the ", reason: ..." here and just refactor the message to `fmt.Sprintf("Failed to unpack the %s Bundle: %s", bundle.GetName(), bundleMessage)`. WDYT?

akihikokuroda: It's good.  I'll make change.  Thanks!

internal/provisioner/plain/controllers/bundledeployment_controller.go
```diff
@@ -132,7 +132,12 @@ func (r *BundleDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		if bundle.Status.Phase == rukpakv1alpha1.PhaseFailing {
 			reason = rukpakv1alpha1.ReasonUnpackFailed
 			status = metav1.ConditionFalse
-			message = fmt.Sprintf("Failed to unpack the %s Bundle", bundle.GetName())
+			c := meta.FindStatusCondition(bundle.Status.Conditions, rukpakv1alpha1.TypeUnpacked)
+			bundleMessage := "No unpacked condition"
+			if c != nil {
+				bundleMessage = c.Message
+			}
```
timflannagan, line 139:
```suggestion
			bundleMessage := "No unpacked condition"
			if c := meta.FindStatusCondition(bundle.Status.Conditions, rukpakv1alpha1.TypeUnpacked); c != nil {
				bundleMessage = c.Message
			}
```

akihikokuroda: The new condition is now:
```
  status:
    conditions:
    - lastTransitionTime: "2022-07-08T17:17:22Z"
      message: 'Failed to unpack the fail1-7d8b4f45b4 Bundle: get objects from bundle
        manifests: read ".gitignore": error unmarshaling JSON: while decoding JSON:
        json: cannot unmarshal string into Go value of type map[string]interface {}'
      reason: UnpackFailed
      status: "False"
      type: HasValidBundle
```

internal/provisioner/plain/controllers/bundledeployment_controller.go
```diff
@@ -132,7 +132,11 @@ func (r *BundleDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		if bundle.Status.Phase == rukpakv1alpha1.PhaseFailing {
 			reason = rukpakv1alpha1.ReasonUnpackFailed
 			status = metav1.ConditionFalse
-			message = fmt.Sprintf("Failed to unpack the %s Bundle", bundle.GetName())
+			bundleMessage := "No unpacked condition"
+			if c := meta.FindStatusCondition(bundle.Status.Conditions, rukpakv1alpha1.TypeUnpacked); c != nil {
+				bundleMessage = c.Message
+			}
+			message = fmt.Sprintf("Failed to unpack the %s Bundle: %s", bundle.GetName(), bundleMessage)
```
timflannagan, line 139: I think we're almost there. Given we know where handling a failing Bundle unpack phase, tacking on a "No unpacked condition" might not be the greatest UX here.

```
message := fmt.Sprintf("Failed to unpack the %s Bundle", bundle.GetName())
if c := meta.FindStatusCondition(bundle.Status.Conditions, rukpakv1alpha1.TypeUnpacked); c != nil {
    message = fmt.Sprintf("%s: %s", message, c.Message)
}
...
```

We should just be able to override the default message variable and append that condition's message after the original value.

akihikokuroda: I see.  That's better.

akihikokuroda: It's clean 3 line change now :-)

timflannagan (approved):
I didn't test this locally but these changes look fine to me. Do you think we need a test here?

akihikokuroda:
Probably not.  I'll check how the current tests handle the error cases.  I'll create another issue if the tests are not enough.

timflannagan:
Yeah, sounds good to me.

akihikokuroda:
Issue for e2e test #457

Issue #425, opened 2022-06-07
Support out-of-tree provisioner implementations


exdx:
Rukpak currently supports multiple provisioner implementations that are all housed in-tree. The tooling and documentation are geared towards this design. Furthermore, all the packages involved in building a provisioner are behind the `internal` directory at the root level at the moment. 

As the APIs stabilize, it becomes more desirable to support different provisioner implementations that don't necessarily live in the rukpak codebase. This can be thought of similar to the Rust crate system, where the standard library is small and provides a few necessary crates, and many other crates are then developed by the community outside of the core payload. For example, a helm provisioner would be useful, but does not necessarily have to be a part of core rukpak.  

Support for out of tree provisioners can come in several stages.

Phase 1:
* Move commonly useful provisioner packages out of `internal` and under `pkg`
* Add documentation on developing your own provisioner

Phase 2:
* Develop (or highlight) an external provisioner implementation 
* Add docs and tooling around integrating an external provisioner with the existing rukpak stack on-cluster

Phase 3:
* Support an ecosystem of different provisioner implementations 

This issue can be in reference to Phase 1, and follow-up issues can arise to track the subsequent phases.

exdx:
note: this may be more relevant in the near term based on how the deppy rukpak provisioner is implemented and packaged. 

We can look to the [Gateway API](https://gateway-api.sigs.k8s.io/) as an example of extensible kuberenetes APIs that can be used by several different providers.

akihikokuroda:
I would like to refactor the helm provisioner to out-of-tree provisioner.  So I want to start the phase 1.  @operator-framework/rukpak-maintainers WDYT?

tylerslaton:
> I would like to refactor the helm provisioner to out-of-tree provisioner. So I want to start the phase 1. @operator-framework/rukpak-maintainers WDYT?

That sounds like a good call to me. I think we'll want to get some of the common packages (like what you listed above) into a `pkg`. That'll feed into getting an out-of-tree provisioner for helm going without much blockage from our (current) internal packages.

joelanford:
> we'll want to get some of the common packages (like what you listed above) into pkg first so out-of-tree repositories can utilize them

My ask here is that we do this pretty selectively and with a critical eye on the package names and exported types/functions.
- What's the minimal public API required to support everything we'd want an out-of-tree plugin to be able to do
- Of the functionality that we decide needs to be public, should any of it be refactored such that we minimize the chance of any of it breaking in the future?

[...]

Issue #460, opened 2022-07-13
Introduce a helm chart provisioner


cmoulliard:
## Requirement

That should be great to have a provisioner able to install/uninstall/upgrade a helm chart and that we have also the possibility to pass the values.yaml (key/values pairs to customize the helm template resource files). That will certainly imply that we design something like (Helm tiller) to perform the rendering of the chart into k8s resources during the installation

timflannagan:
Yep, a helm provisioner seems like a solid candidate for the next provisioner implementation, but I imagine the priority is low for the directly OLM team in the short term while we continue to iterate on layers that build on top of rukpak primitives.

We've briefly floated around the idea of prototyping a helm provisioner, which has it's own class of potential problems:

- How do we expose the values.yaml in a way that rukpak's BundleDeployment API can consume them?
- Would it be possible to have a provisioner implementation that sources helm charts, and "converts" (e.g. renders those charts + values.yaml) that bundle format into a "plain bundle" format?

That conversion logic is something we have already pursued w.r.t the registry+v1 provisioner, which translates OLM's legacy registry+v1 bundle format into a "plain bundle" format, so there's already precedence for that runtime design. And the benefit there is that we only need to bootstrap a controller that watches the rukpak Bundle resource, converts the bundle format into a plain bundle format, and stores that bundle format in it's in-memory store.

Thoughts @joelanford?

timflannagan:
Moving this to the backlog for now, and we can re-groom later. Another related issue may be #425 as a helm provisioner may not be a core rukpak provisioner (if there's even core rukpak provisioners that ship from this repository longer term), and may live out-of-tree going forward.

cmoulliard:
> * How do we expose the values.yaml in a way that rukpak's BundleDeployment API can consume them?

They could be passed simply using a ConfigMap.

cmoulliard:
> * Would it be possible to have a provisioner implementation that sources helm charts, and "converts" (e.g. renders those charts + values.yaml) that bundle format into a "plain bundle" format?

I would like a lot that the provisioner is able to render the templates using a template engine + values passed. This is what carvel kapp controller do - https://carvel.dev/kapp-controller/docs/v0.38.0/app-examples/#gitops-with-a-helm-chart

timflannagan:
> They could be passed simply using a ConfigMap.

[...]

PR #462, opened 2022-07-15
move some internal code to pkg

akihikokuroda:
as a part of phase 1 of #425, this PR moves 

`source`, `storage`, `updater` `version` `finalizer` and some methods in `util` to `pkg`


joelanford:
@akihikokuroda did you link the right parent issue? #452 seems to be unrelated.

akihikokuroda:
@joelanford Sorry. it is #425

joelanford:
Ah! Thanks. Do we have an out-of-tree provisioner we're working on? That might help drive prioritize where to improve our various "provisioner-runtime" sorts of packages and also drive what is made public at all.

Not sure what others think, but my preference would be to have a specific motivation for each thing we make public. Reason being that once we call rukpak 1.0, we'd be making the commitment that those public packages would not have any breaking changes.

At that point, we could always decide to move those library bits to a separate repo to separate the versions of the APIs/controllers from the underlying libraries.

akihikokuroda:
I see.  I saw #460 referencing #425.  If it doesn't get priority, this work should wait then.

timflannagan:
Yep, my preference would be to keep everything internal. I think we've discussed this before, and someone (Joe?) made the argument that we're an alpha level project, and therefore external repositories importing packages defined in this repository under pkg/ don't have the same API surface guarantees we live to make in other OLM projects.

timflannagan:
With that said, we may want to start thinking about what needs to live in core rukpak over the next couple of releases if the maintenance burden of additional provisioner requirements is too substantial for it to live in-tree.

akihikokuroda:
I close this PR.

Issue #464, opened 2022-07-19
The `local` source implementation has a few drawbacks


joelanford:
Getting a simple source implementation in place to support local bundle content is a major quality of life improvement from the current `local` source that uses a configmap to store bundle contents. However, this implementation has two major drawbacks:
1. It limits the size of a bundle to the maximum etcd value size (minus overhead), which can be as little as 1MB
2. It currently only supports plain manifest bundles because configmaps don't directly support directory hierarchies (this _could_ be overcome by switching to a `BinaryData` tar archive value rather than the current key/value per file format).

To level-up our support for allowing operator authors to reference local bundles on-cluster, I propose a different mechanism: Run an RBAC-protected webservice on-cluster that:
1. Accepts PUT requests for non-unpacked bundles with source type `binary` (name TBD).
2. Accepts GET requests so that bundle provisioners can fetch the uploaded content.

This mechanism would solve both of the main drawbacks of the existing `local` source type.

exdx:
Added to current milestone

timflannagan:
Are we putting a line in the sand that we don't want to support local (e.g. ConfigMap) bundle sources, or is this something we can re-evaluate over time? Do we have an idea of any explicit use cases for why users want/need to be able to use ConfigMaps to house bundle contents?

exdx:
I imagine both can be useful -- some users will be more comfortable with configmaps then tar.gz files as a source type. I don't see why adding binary necessarily involves ripping out local. 

Furthermore, from a planning perspective, local bundle support was on the radar for a while, triaged, groomed, added to a milestone, and delivered. The binary feature sort of came from nowhere and is removing that local feature in the very next release. I'm not convinced this is optimal.

joelanford:
> some users will be more comfortable with configmaps then tar.gz files as a source type

Even if we kept the configmap-based source in, I think we'd want to solve problem #2 that I mention in the issue description (configmaps don't natively support directory hierarchy). That's very limiting since most bundle formats will very likely end up evolving to use a directory somewhere.

My perspective is that we're currently trying to solve the problem of "how does a developer work with an arbitrary local bundle?" and I feel like the `binary` source is an improvement on the `local` source that solves the same problem.

> From a GitOps perspective though

[...]

PR #471, opened 2022-07-26
rukpakctl: remove use of globals and exported types; use contexts consistently

joelanford:
This is step 1 of making some improvements discussed in #464, which is a minor refactoring of the rukpakctl commands to remove global vars and unnecessarily exported types.

I also uncovered a few places we should be passing contexts down, so I've made those fixes as well.

Step 2 will follow up with further changes, to include:
- renaming `local` to `configmaps`
- supporting multiple configmap references which include their directory paths within bundle
- forcing the configmap references to be in the rukpak-system namespace (that way bundle controllers don't have to cache all configmaps on the cluster)

### DIFF
--- a/cmd/rukpakctl/cmd/bundle.go
+++ b/cmd/rukpakctl/cmd/bundle.go
@@ -19,87 +19,80 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"log"
 
 	"github.com/spf13/cobra"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/kubernetes/scheme"
 	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
 
 	rukpakv1alpha1 "github.com/operator-framework/rukpak/api/v1alpha1"
 	"github.com/operator-framework/rukpak/cmd/rukpakctl/utils"
 )
 
-type BundleOptions struct {
+type bundleOptions struct {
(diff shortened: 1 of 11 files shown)


cmd/rukpakctl/cmd/root.go
```diff
@@ -5,55 +5,35 @@ Copyright © 2022 NAME HERE <EMAIL ADDRESS>
...
 Cobra is a CLI library for Go that empowers applications.
 This application is a tool to generate the needed files
 to quickly create a Cobra application.`,
-	// Uncomment the following line if your bare application
-	// has an action associated with it:
-	// Run: func(cmd *cobra.Command, args []string) { },
-}
+		// Uncomment the following line if your bare application
+		// has an action associated with it:
+		// Run: func(cmd *cobra.Command, args []string) { },
```
tylerslaton, line 27: nit: Should we just go ahead and remove this comment?

tylerslaton (approved):
These changes all look reasonable to me! Had one very minor nit.

akihikokuroda (approved):

cmd/rukpakctl/cmd/bundle.go
```diff
@@ -19,87 +19,80 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"log"
 
 	"github.com/spf13/cobra"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/tools/clientcmd"
```
timflannagan, line 26: :+1:

timflannagan (approved):

cmd/rukpakctl/cmd/root.go
joelanford: Sure!

PR #473, opened 2022-07-29
helm provisioner initial drop

akihikokuroda:

This is a part of #460

# Helm provisioner

### BundleDeployment

`Config runtime.RawExtension json:"config,omitempty"`

is added to hold provisioner specific configurations.

```
// BundleDeploymentSpec defines the desired state of BundleDeployment
type BundleDeploymentSpec struct {
	// ProvisionerClassName sets the name of the provisioner that should reconcile this BundleDeployment.
	// Template describes the generated Bundle that this deployment will manage.
	Template *BundleTemplate `json:"template"`
	// Config is provisioner specific configurations
}
```
For the Helm provisioner, the `Config` is used for providing the values.yaml contents in an embedded string.

```
    values: |
      # Default values for hello-world.
      # This is a YAML-formatted file.
      # Declare variables to be passed into your templates.
      replicaCount: 1
      image:
        repository: nginx
        pullPolicy: IfNotPresent
        # Overrides the image tag whose default is the chart appVersion.
        tag: ""
      nameOverride: "formvalues"
      fullnameOverride: ""
      serviceAccount:
        # Specifies whether a service account should be created
        create: true
        # Annotations to add to the service account
        annotations: {}
        # The name of the service account to use.
        # If not set and create is true, a name is generated using the fullname template
        name: ""
```

### Bundle
The `HTTPSource` is added to the `BundleSource`
```
type HTTPSource struct {
	// URL is where the bundle contents is.
	URL string `json:"url"`
        // Auth configures the authorization method if necessary.
        Auth Authorization `json:"auth,omitempty"`
}
```

## Implementation

Helm chars are downloaded from repository by the bundle controller. The downloaded chart is stored in the `storage` in the helm chart unpacked.  An example of the stored chart is 

```
hello-world/
hello-world/README.md
hello-world/Chart.yaml
hello-world/.helmignore
hello-world/templates
hello-world/templates/_helpers.tpl
hello-world/templates/service.yaml
hello-world/templates/serviceaccount.yaml
hello-world/templates/NOTES.txt
hello-world/templates/deployment.yaml
hello-world/values.yaml
```
from a [sample helm chart](https://github.com/helm/examples)

This can be obtained via 
```
      source:
        http:
          url: https://github.com/helm/examples/releases/download/hello-world-0.1.0/hello-world-0.1.0.tgz  
        type: http
```
or 
```
      source:
        git:
          ref:
            branch: main
          repository: https://github.com/helm/examples
          directory: ./charts
        type: git
```
The bundledeployment controller installs the chart using`github.com/operator-framework/helm-operator-plugins/pkg/client` which is using `helm.sh/helm/v3` internally.  It applies the values from `config` at install.


internal/provisioner/helm/main.go
```diff
@@ -0,0 +1,226 @@
...
+
+	cfg := ctrl.GetConfigOrDie()
+	mgr, err := ctrl.NewManager(cfg, ctrl.Options{
+		Scheme:                 scheme,
+		MetricsBindAddress:     httpBindAddr,
+		Port:                   9443,
+		HealthProbeBindAddress: probeAddr,
+		LeaderElection:         enableLeaderElection,
+		LeaderElectionID:       "helm.core.rukpak.io",
+		NewCache: cache.BuilderWithOptions(cache.Options{
```
exdx, line 110: Can we setup this cache to only look for Bundles that have the provisioner ID set to `helm`? I believe we do this for the other provisioner controllers.

internal/source/http.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	httpClient := http.Client{Timeout: 10 * time.Second}
+	resp, err := httpClient.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("%s: http request for bundle content failed: %v", action, err)
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode == http.StatusNotFound {
+		return &Result{State: StatePending, Message: "waiting for bundle to be uploaded"}, nil
```
exdx, line 54: Is it possible that this 404 error occurs when the helm chart provided cannot be found on artifact hub? If so, this error message would be misleading, as the bundle would never get uploaded. Maybe we should make the error more generic?

api/v1alpha1/bundledeployment_types.go
```diff
@@ -49,6 +50,9 @@ type BundleDeploymentSpec struct {
 	ProvisionerClassName string `json:"provisionerClassName"`
 	// Template describes the generated Bundle that this deployment will manage.
 	Template *BundleTemplate `json:"template"`
+	// Config is provisioner specific configurations
+	// +kubebuilder:pruning:PreserveUnknownFields
+	Config runtime.RawExtension `json:"config,omitempty"`
```
exdx, line 55: Is this specifically for the helm provisioner? If so, I think we should name it to reflect that?

exdx (commented):
This looks good for WIP. There is a lot of code duplication, so #447 should help a lot here when it lands potentially. 

I think it would be helpful to have a small doc that outlines this new provisioner, or even just an example under testcases.

internal/source/http.go
akihikokuroda: I'll update.  Thanks!

internal/provisioner/helm/main.go
akihikokuroda: I looked at the other provisioners.  I couldn't find they limited the cache contents.  I found they control the watch in `SetupWithManager`.  Anyway, it's reasonable to limit the cache, I'll look into it.  Thanks!

akihikokuroda: It seems that the field selector for arbitrary fields of CRD are not supported.    https://github.com/kubernetes/kubernetes/issues/53459

api/v1alpha1/bundle_types.go
```diff
@@ -64,6 +65,8 @@ type BundleSource struct {
 	Git *GitSource `json:"git,omitempty"`
 	// Local is a reference to a local object in the cluster.
 	Local *LocalSource `json:"local,omitempty"`
+	//  HTTP is the remote location that backs the content of this Bundle.
+	HTTP *HTTPSource `json:"http,omitempty"`
```
exdx, line 69: Maybe this should be `HTTPS` if most charts being referenced are from a domain secured by a certificate? And we want to encourage people to use https in general?

internal/provisioner/helm/README.md
```diff
@@ -0,0 +1,272 @@
+# Plain Provisioner
+
+## Summary
+
+The `helm` provisioner is able to instantiate a given `helm+v0` bundle with a specified helm chart onto a cluster and then install the helm chart,
```
exdx, line 5: This brings up the interesting point of whether we need the `+v0` -- since helm is an external format, I'm not sure we are going to have the same metadata questions we have regarding our own, internal, `plain` format. I'm personally for dropping the v0 and just calling them helm bundles in the short term. 

This would create less confusion on what helm bundles are and how to use them.

internal/provisioner/helm/README.md
```diff
@@ -0,0 +1,272 @@
...
+### Install and apply a specific version of a `helm+v0` bundle
+
+> :warning: Anyone with the ability to create or update BundleDeployment objects can become cluster admin. It's important
+> to limit access to this API via RBAC to only those that explicitly require access, as well as audit your bundles to
+> ensure the content being installed on-cluster is as-expected and secure.
+
+The `helm` provisioner can install and make available a specific `helm+v0` bundle in the cluster.
+
+Simply create a `BundleDeployment` resource that contains the desired specification of a Bundle resource.
+The `helm` provisioner will retrive the specifie helm chart and instantiate Bundle onto the cluster, and eventually install the helm chart
```
exdx, line 19:
```suggestion
The `helm` provisioner will retrieve the specific helm chart and instantiate the Bundle onto the cluster, and eventually install the helm chart
```

internal/provisioner/helm/README.md
```diff
@@ -0,0 +1,272 @@
...
+        http:
+          url: https://github.com/helm/examples/releases/download/hello-world-0.1.0/hello-world-0.1.0.tgz  
+        type: http
+```
+
+For the helm chart, the values file can be applied during the installtation of the chart.  The values.yaml has to be in a ConfigMap in the namespace where the helm provisioner is running.
+The ConfigMap can be created by the kubectl
+
+```console
+kubectl create configmap my-ahoy-values --from-file=values.yaml
```
exdx, line 48: Could we show what is actually in the `values.yaml` file? Might be helpful for users to see.

internal/provisioner/helm/README.md
```diff
@@ -0,0 +1,272 @@
...
+```yaml
+  config:
+    valuesFrom:
+      configMapName: my-ahoy-values
+```
+
+> Note: the generated Bundle will contain the BundleDeployment's metadata.Name as a prefix, followed by
+> the hash of the provided template.
+
+As the bundle content is retrived and stored onto the cluster via the defined storage mechanism, the bundle status
```
exdx, line 62:
```suggestion
As the bundle content is retrieved and stored onto the cluster via the defined storage mechanism, the bundle status
```

internal/provisioner/helm/README.md
```diff
@@ -0,0 +1,272 @@
...
+will be updated to Unpacked, indicating that all its contents have been stored on-cluster.
+
+```console
+$ kubectl get bundle -l app=my-ahoy
+NAME                 TYPE   PHASE      AGE
+my-ahoy-5764594dc8   http   Unpacked   33s
```
exdx, line 68: should this be `TYPE` helm?

internal/provisioner/helm/README.md
```diff
@@ -0,0 +1,272 @@
...
+```console
+$ kubectl delete bundledeployments.core.rukpak.io my-ahoy
+bundledeployment.core.rukpak.io "my-ahoy" deleted
+```
+
+To stop and clean up the kind cluster, delete it:
+
+```bash
+# Clean up kind cluster
+make kind-cluster-cleanup
```
exdx, line 271: I think `kind delete cluster` also makes sense, if we want to leave the host in the same exact state as before running the cluster.

api/v1alpha1/bundle_types.go
akihikokuroda: OK. I'll change HTTP to HTTPS everywhere.

internal/provisioner/helm/README.md
akihikokuroda: OK.  That makes sense.

internal/provisioner/helm/README.md
akihikokuroda: Thanks!

internal/provisioner/helm/README.md
akihikokuroda: OK.  I'll add it.  Here is the values.yaml file from a sample.  I'll change the sample from apache to helm/examples to make it easy.

```
# Default values for hello-world.
# This is a YAML-formatted file.
# Declare variables to be passed into your templates.

replicaCount: 1

image:
  repository: nginx
  pullPolicy: IfNotPresent
  # Overrides the image tag whose default is the chart appVersion.
  tag: ""

nameOverride: "formvalues"
fullnameOverride: ""

serviceAccount:
  # Specifies whether a service account should be created
  create: true
  # Annotations to add to the service account
  annotations: {}
  # The name of the service account to use.
  # If not set and create is true, a name is generated using the fullname template
  name: ""

service:
  type: ClusterIP
  port: 80

```

internal/provisioner/helm/README.md
akihikokuroda: Thanks!

internal/provisioner/helm/README.md
akihikokuroda: This column shows the source type.  We can add provisioner class name.

internal/provisioner/helm/README.md
akihikokuroda: OK.  The `make run` creates the rukpak cluster.  So I'll change the line to `kind delete clusters rukpak`

api/v1alpha1/bundle_types.go
```diff
@@ -93,6 +96,13 @@ type LocalSource struct {
 	ConfigMapRef *ConfigMapRef `json:"configMap"`
 }
 
+type HTTPSource struct {
```
tylerslaton, line 99: +1 To Dan's comment. Let's make sure to update all references to HTTP to be HTTPS. I can't imagine a world where we would want to support HTTP.

akihikokuroda:
@exdx Thanks for review.  I'll make changes.  I'll also change the sample helm chart from the apache to the helm/examples and add another test with the values.yaml file.

internal/provisioner/helm/README.md
```diff
@@ -0,0 +1,272 @@
...
+## Summary
+
+The `helm` provisioner is able to instantiate a given `helm+v0` bundle with a specified helm chart onto a cluster and then install the helm chart,
+in the cluster. It does so by reconciling `Bundle` and `BundleDeployment` types that have
+the `spec.provisionerClassName` field set to `core.rukpak.io/helm`. This field must be set to the correct provisioner
+name in order for the `helm` provisioner to see and interact with the bundle.
+
+### Install and apply a specific version of a `helm+v0` bundle
+
+> :warning: Anyone with the ability to create or update BundleDeployment objects can become cluster admin. It's important
```
tylerslaton, line 12: nit:
```suggestion
> :warning: Anyone with the ability to create or update BundleDeployment objects can become a cluster admin. It's important
```

internal/provisioner/helm/README.md
```diff
@@ -0,0 +1,272 @@
...
+        app: my-ahoy
+    spec:
+      provisionerClassName: core.rukpak.io/helm
+      source:
+        http:
+          url: https://github.com/helm/examples/releases/download/hello-world-0.1.0/hello-world-0.1.0.tgz  
+        type: http
+```
+
+For the helm chart, the values file can be applied during the installtation of the chart.  The values.yaml has to be in a ConfigMap in the namespace where the helm provisioner is running.
```
tylerslaton, line 44: nit:
```suggestion
For the helm chart, the values file can be applied during the installtation of the chart. The values.yaml has to be in a ConfigMap in the namespace where the helm provisioner is running.
```

internal/provisioner/helm/types/types.go
```diff
@@ -0,0 +1,6 @@
+package plain
```
tylerslaton, line 1:
```suggestion
package helm
```

internal/source/http.go
```diff
@@ -0,0 +1,87 @@
...
+
+	"github.com/go-git/go-billy/v5/memfs"
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	rukpakv1alpha1 "github.com/operator-framework/rukpak/api/v1alpha1"
+)
+
+// http is a bundle source that sources bundles from the specified url.
+type HTTP struct {
```
tylerslaton, line 20: Just a reminder to update to use `https` instead of `http` across the implementation.

test/e2e/helm_provisioner_test.go
```diff
@@ -0,0 +1,131 @@
...
+			}
+			err := c.Create(ctx, bd)
+			Expect(err).To(BeNil())
+		})
+		AfterEach(func() {
+			By("deleting the testing BD resource")
+			Expect(c.Delete(ctx, bd)).To(BeNil())
+		})
+
+		It("should rollout the bundle contents successfully", func() {
```
tylerslaton, line 61: I would really like to see a failure path test implemented as well.

tylerslaton (commented):
Alright, initial round of review from me. All in all this looks great. As far as what @exdx was mentioning about the deduplication PR helping this out, I would be fine with getting this through as is and then handling whatever deduplication we deem necessary at that time. 

Right now our main goal is to consolidate all of core Rukpak provisioners into a single process which I'm working on now. After that we will revisit the deduplication.

internal/provisioner/helm/types/types.go
akihikokuroda: Thanks!

internal/provisioner/helm/README.md
akihikokuroda: Thanks!

internal/provisioner/helm/README.md
akihikokuroda: Thanks!

internal/source/http.go
akihikokuroda: I will.  Thanks!

test/e2e/helm_provisioner_test.go
akihikokuroda: I'll add some failure paths.  Thanks!

api/v1alpha1/bundle_types.go
joelanford: The scheme will ultimately be determined by the URL though right? Will we accept `http://` URLs? I could imagine a use case where the bundle is hosted by another pod in the namespace, or perhaps in some wild configuration in a container that's a sidecar of the helm provisioner (in which case `http://localhost:8080/my-chart.tgz` might be the URL)

internal/provisioner/helm/README.md
joelanford: I'd probably call them helm v3 bundles. Helm made some minor (maybe breaking, can't remember?) changes between v2 and v3.

api/v1alpha1/bundle_types.go
akihikokuroda: Yes, the scheme is determined by the URL and it can accept http URLs.

internal/provisioner/helm/README.md
akihikokuroda: "v3" makes sense.  I'll put "+v3".

internal/provisioner/helm/controllers/bundle_controller.go
```diff
@@ -0,0 +1,184 @@
...
+		return ctrl.Result{}, err
+	}
+
+	unpackResult, err := r.Unpacker.Unpack(ctx, bundle)
+	if err != nil {
+		return ctrl.Result{}, updateStatusUnpackFailing(&u, fmt.Errorf("source bundle content: %v", err))
+	}
+	if unpackResult.State == source.StateUnpacked {
+		if err := r.Storage.Store(ctx, bundle, unpackResult.Bundle); err != nil {
+			return ctrl.Result{}, updateStatusUnpackFailing(&u, fmt.Errorf("persist bundle objects: %v", err))
```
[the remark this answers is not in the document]
akihikokuroda: OK.  I'll work on it.

api/v1alpha1/bundle_types.go
```diff
@@ -93,6 +96,13 @@ type LocalSource struct {
...
+	// URL is where the bundle contents is.
+	URL string `json:"url"`
+	// Auth configures the authorization method if necessary.
+	Auth Authorization `json:"auth,omitempty"`
```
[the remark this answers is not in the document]
akihikokuroda: OK.  I'll update.

internal/source/http.go
```diff
@@ -0,0 +1,87 @@
...
+
+	url := bundle.Spec.Source.HTTP.URL
+	action := fmt.Sprintf("%s %s", http.MethodGet, url)
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
+	if err != nil {
+		return nil, fmt.Errorf("create http request %q for bundle content: %v", action, err)
+	}
+	var userName, password string
+	if bundle.Spec.Source.HTTP.Auth.Secret.Name != "" {
```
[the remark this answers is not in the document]
akihikokuroda: I'll add the support.

internal/provisioner/helm/README.md
```diff
@@ -0,0 +1,304 @@
+# Plain Provisioner
```
joelanford, line 1:
```suggestion
# Helm Provisioner
```

internal/provisioner/helm/README.md
```diff
@@ -0,0 +1,304 @@
...
+```yaml
+apiVersion: core.rukpak.io/v1alpha1
+kind: BundleDeployment
+metadata:
+  name: my-ahoy
+spec:
+  provisionerClassName: core.rukpak.io/helm
+  config:
+    valuesFrom:
+      configMapName: my-ahoy-values
```
joelanford, line 31: WDYT about (for now at least) embedding the values directly in the bundle deployment either as a `map[string]interface{}`, or as a `string`, rather than referencing a configmap? That would decrease the complexity (no need to handle all the possibilities around a configmap being present/not present/showing up/disappearing/etc). 

E.g.

```yaml
spec:
  config:
    values: # map[string]interface{}
      replicaCount: 1
      
      image:
        repository: nginx
        pullPolicy: IfNotPresent
        # Overrides the image tag whose default is the chart appVersion.
        tag: ""
    # or
    values: |  # string
      replicaCount: 1
      
      image:
        repository: nginx
        pullPolicy: IfNotPresent
        # Overrides the image tag whose default is the chart appVersion.
        tag: ""
```

internal/provisioner/helm/controllers/bundledeployment_controller.go
```diff
@@ -0,0 +1,480 @@
...
+	}
+
+	return ctrl.Result{}, nil
+}
+
+// reconcileDesiredBundle is responsible for checking whether the desired
+// Bundle resource that's specified in the BundleDeployment parameter's
+// spec.Template configuration is present on cluster, and if not, creates
+// a new Bundle resource matching that desired specification.
+func reconcileDesiredBundle(ctx context.Context, c client.Client, bd *rukpakv1alpha1.BundleDeployment) (*rukpakv1alpha1.Bundle, *rukpakv1alpha1.BundleList, error) {
```
joelanford, line 281: This function seems ripe for moving to a package that can be shared by plain and helm provisioners.

internal/source/https.go
```diff
@@ -0,0 +1,95 @@
...
+		return nil, fmt.Errorf("creating filesystem: %s", err)
+	}
+	_, err = io.Copy(file, resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("writing chart file: %s", err)
+	}
+	err = file.Close()
+	if err != nil {
+		return nil, fmt.Errorf("closing chart file: %s", err)
+	}
```
joelanford, line 77: The https source shouldn't be helm-specific. My thought would be to assume that all bundles downloaded via the https source are `.tgz` files, and we should return the ungzip'ed untar'd representation, like this: https://github.com/operator-framework/rukpak/blob/3b832db489499105211cda18f2217aad89e2bb8f/internal/storage/http.go#L84-L88

That way, if someone stored a tgz plain or registry bundle on an HTTP server somewhere, this http source could be used just like the other sources (all of which actually return an already unpacked view of the bundle)

internal/provisioner/helm/README.md
akihikokuroda: I see.  It seems better.  I'll change to the `string` one.

timflannagan (commented):
I believe we're planning to consolidate the registry+v1, plain provisioner, and upload manager executables into a single executable, which should also help cut down on the number of required manifests.

Which leads me to the following questions:

- Does it make sense to package any helm provisioner outside of this shared executable?
- And if yes, should we investigate moving this implementation out-of-tree?

Thoughts?

internal/provisioner/helm/controllers/bundledeployment_controller.go
akihikokuroda: I'll move this to internal/util/util.go

internal/source/https.go
akihikokuroda: I see.  I'll change it.

internal/provisioner/helm/README.md
akihikokuroda: Thanks!

akihikokuroda:
@timflannagan I think we eventually have to have out-of-tree provisioner and the helm provisioner can be a good candidate of the example.

internal/source/http.go
```diff
@@ -0,0 +1,87 @@
...
+		userName, password, err = b.getCredentials(ctx, bundle)
+		if err != nil {
+			return nil, err
+		}
+		req.SetBasicAuth(userName, password)
+	}
+
+	httpClient := http.Client{Timeout: 10 * time.Second}
+	resp, err := httpClient.Do(req)
```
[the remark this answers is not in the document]
akihikokuroda: Yes, I does.  I can put `CheckRedirect` function if it should not.

test/e2e/helm_provisioner_test.go
akihikokuroda: I added some failure path tests.

internal/provisioner/helm/controllers/bundle_controller.go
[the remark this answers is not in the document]
akihikokuroda: I added `lint` before storing the chart into the storage.

internal/provisioner/helm/README.md
akihikokuroda: done.

internal/source/https.go
akihikokuroda: done.

akihikokuroda:
This is ready for another round of review.  Thanks!

internal/source/http.go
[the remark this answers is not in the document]
joelanford: No, I think following redirects is probably better.

cmd/helm/main.go
```diff
@@ -0,0 +1,226 @@
...
+		setupLog.Error(err, "unable to register finalizer", "finalizerKey", finalizer.DeleteCachedBundleKey)
+		os.Exit(1)
+	}
+
+	unpacker := source.NewUnpacker(map[rukpakv1alpha1.SourceType]source.Unpacker{
+		rukpakv1alpha1.SourceTypeHTTP: &source.HTTP{
+			Reader:          mgr.GetAPIReader(),
+			SecretNamespace: systemNamespace,
+		},
+	})
```
joelanford, line 185: Any reason we shouldn't support all of the source types in the helm provisioner?

internal/provisioner/helm/README.md
```diff
@@ -0,0 +1,289 @@
...
+        app: my-ahoy
+    spec:
+      provisionerClassName: core.rukpak.io/helm
+      source:
+        https:
+          url: https://github.com/helm/examples/releases/download/hello-world-0.1.0/hello-world-0.1.0.tgz  
+        type: https
+```
+
+For the helm chart, the values file embedded in the `confg` can be applied during the installtation of the chart.
```
joelanford, line 65:
```suggestion
For the helm chart, the values file embedded in the `config` can be applied during the installation of the chart.
```

internal/provisioner/helm/controllers/bundle_controller.go
```diff
@@ -0,0 +1,241 @@
...
+
+	unpackResult, err := r.Unpacker.Unpack(ctx, bundle)
+	if err != nil {
+		return ctrl.Result{}, updateStatusUnpackFailing(&u, fmt.Errorf("source bundle content: %v", err))
+	}
+	if unpackResult.State != source.StateUnpacked {
+		return ctrl.Result{}, updateStatusUnpackFailing(&u, fmt.Errorf("unknown unpack state %q: %v", unpackResult.State, err))
+	}
+
+	// Writing chart contents in temp directory for linting
```
joelanford, line 141: Ohhh. Sorry I didn't see this earlier, but how about we do something along these lines:
```go
pr, pw := io.Pipe()
eg, ctx := errgroup.WithContext(ctx)
eg.Go(func() error {
	return pw.CloseWithError(util.FSToTarGZ(pw, bundleFS))
})
eg.Go(func() error {
	chart, err := loader.LoadArchive(pr)
	if err != nil {
		return err
	}
	return chart.Validate()			
})
return eg.Wait()
```

joelanford: Do we also need to run the linter? If there are lint errors, should we reject that bundle?

internal/source/http.go
```diff
@@ -0,0 +1,86 @@
...
+			return nil, err
+		}
+		req.SetBasicAuth(userName, password)
+	}
+
+	httpClient := http.Client{Timeout: 10 * time.Second}
+	if bundle.Spec.Source.HTTP.Auth.InsecureSkipVerify {
+		tr := &http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		}
```
joelanford, line 50: I'm not sure how much it matters, but the `http` package has a `DefaultTranport` that sets a lot more fields on this struct. Because of that, I did this in `rukpakctl run`:
```go
transport := http.DefaultTransport.(*http.Transport).Clone()
```

internal/provisioner/helm/controllers/bundle_controller.go
akihikokuroda: It rejects the bundle if it fails lint now.  The validate function just validates the chart.yaml file.  Is it enough check before it stores the chart?

cmd/helm/main.go
akihikokuroda: Oh, I didn't think about it.  There is no reason that we cannot support the other sources.

internal/provisioner/helm/README.md
akihikokuroda: Thanks.

cmd/helm/main.go
akihikokuroda: Does the registry provisioner support the gitsource?   In theory, it can but have we tested?

internal/source/http.go
akihikokuroda: Thanks!  I'll do the same way.

internal/provisioner/helm/controllers/bundle_controller.go
akihikokuroda: I changed to just call `Validate`.

tylerslaton (approved):
This looks good to me. We should try to come to a decision on if we want this to be brought into the `core` process or not. I'll go ahead and make an issue for that if/when this merges as is - but it isn't blocking this PR.

Nice work!

akihikokuroda:
I made changes to make values changes in the BD applied to the installation.

exdx (approved):

timflannagan:
(For release notes)

PR #480, opened 2022-08-05
Consolidate provisioners and upload manager into a single process

joelanford:
Our current trend seems to be to add a new binary for each new provisioner or somewhat standalone process. This has resulted in lots of duplicate code (see #431), overly large Docker images, more complex and longer builds, duplicate object caches at runtime, and more maintenance of kustomize yaml manifests.

This PR consolidates the upload manager and registry and plain provisioners into the "core" binary and deployment.

The webhook binary (formerly named "core") has also been renamed to "webhooks" but is otherwise unchanged.


timflannagan (approved):
Looks reasonable to me. Only question I had was around the "webhooks" naming, and whether cmd/webooks implies that we'll have a single deployment for all core rukpak webhooks in the future, which may be problematic? But we should be fine for now given there's a single bundle webhook right now, and we can re-evaluate over time.

cmd/core/main.go
```diff
@@ -71,38 +104,138 @@ func main() {
...
+	if err = (&plaincontrollers.BundleDeploymentReconciler{
+		Client:             mgr.GetClient(),
+		Scheme:             mgr.GetScheme(),
+		BundleStorage:      bundleStorage,
+		ReleaseNamespace:   ns,
+		ActionClientGetter: helmclient.NewActionClientGetter(cfgGetter),
+	}).SetupWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create controller", "controller", rukpakv1alpha1.BundleDeploymentKind)
 		os.Exit(1)
 	}
```
tylerslaton, line 240: Perhaps we could do this in an iterative way so that as we add more provisioners this doesn't scale too hard? Could be a follow-up so it isn't blocking.

internal/uploadmgr/gc.go
```diff
@@ -0,0 +1,96 @@
...
+)
+
+type bundleGC struct {
+	storageDir          string
+	storageSyncInterval time.Duration
+	cache               cache.Cache
+	log                 logr.Logger
+}
+
+func NewBundleGC(cache cache.Cache, storageDir string, storageSyncInterval time.Duration) manager.Runnable {
```
tylerslaton, line 27: nit: Would be great to get a comment here describing what this is for and how to use it.

tylerslaton (approved):
This looks good to me. Quick question with the two comments I had - would we not want to also combine the webhooks into a single process as well? Or does that make things a bit messy. I could be convinced either way.

joelanford:
> would we not want to also combine the webhooks into a single process as well?

Are you thinking bundle webhook and crd validator? If so, I left CRD validator separate because it seems there's still some open discussion about the final landing place for that. 

Otherwise, (to answer @timflannagan's question) I _think_ it's okay to assume co-located webhooks in the future. At the very least, it seems like the next time we add a webhook the default thought process will be "let's add it to the webhooks binary" and not "let's make a completely separate binary".  At that point, if someone sees a problem with a single process running multiple webhooks we could split that out.

cmd/core/main.go
joelanford: Let's consider for a follow-up, if that's alright. Looping or extracting into a separate function seems like it could move toward the direction of the `NewUnpacker` signature, which we've already raised as a potential rough edge because of all the required parameters.

Issue #463, opened 2022-07-19
Add documentation for provisioner developers


exdx:
Add some documentation on how to write your own provisioner, to support alternate provisioner implementations. 

Part of #425

PR #482, opened 2022-08-08
provisioner class name validation

### LINKED ISSUE #479: Introduce stricter validation around ProvisionerClass name CRDs
RukPak expects to support multiple provisioners in the future. These provisioners will likely be defined through a CRD. We should update the bundle.Spec.ProvisionerClassName and bundleDeployments.Spec.ProvisionerClassName fields to include strict validation matching that of `metadata.Name` in the k8s api:

```
# example metadata.Name regex
$ k create configmap test./t
error: failed to create configmap: ConfigMap "test./t" is invalid: metadata.name: Invalid value: "test./t": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
 ```

> Note: Services have unique metadata.Name requirements, so the names of services cannot be based off a ProvisionClassName that passes this regex. 
 
As a user, I would like to:
- have the `bundle` and `bundleDeployment` openapiv3 schema reject CRs that include ProvisionerClassNames that cannot be applied to the cluster

### PR DESCRIPTION
**Pull request by akihikokuroda**
Add ProvisionerClassName validation and adjust some e2e tests

Closes #479 

### DIFF
--- a/api/v1alpha1/bundle_types.go
+++ b/api/v1alpha1/bundle_types.go
@@ -50,6 +50,7 @@ const (
 
 // BundleSpec defines the desired state of Bundle
 type BundleSpec struct {
+	//+kubebuilder:validation:Enum:={"core.rukpak.io/plain","core.rukpak.io/registry",}
 	// ProvisionerClassName sets the name of the provisioner that should reconcile this BundleDeployment.
 	ProvisionerClassName string `json:"provisionerClassName"`
 	// Source defines the configuration for the underlying Bundle content.
--- a/api/v1alpha1/bundledeployment_types.go
+++ b/api/v1alpha1/bundledeployment_types.go
@@ -45,6 +45,7 @@ const (
 
 // BundleDeploymentSpec defines the desired state of BundleDeployment
 type BundleDeploymentSpec struct {
+	//+kubebuilder:validation:Enum:={"core.rukpak.io/plain","core.rukpak.io/registry",}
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: api/v1alpha1/bundle_types.go [hunk 1]
```diff
@@ -50,6 +50,7 @@ const (
 
 // BundleSpec defines the desired state of Bundle
 type BundleSpec struct {
+	//+kubebuilder:validation:Enum:={"core.rukpak.io/plain","core.rukpak.io/registry",}
```
exdx, line 53: I think this makes sense, but it directly conflicts with rukpak being an open framework of APIs that users can plug into with their own provisioner. For example, it's at odds with #425 at a minimum, also #463. I'm not sure we want to be this opinionated regarding the rukpak APIs, at least in the upstream project.

akihikokuroda: I see.  We probably wait until we define the way to register provisioner before we implement this validation function.

joelanford:
I think the goal of #479 is to ensure that `bundle.spec.provisionerClassName` and `bundledeployment.spec.provisionerClassName` meet the spec of kubernetes's `metadata.name`.

The reason being that if (and more likely when) we introduce a ProvisionerClass API, those provisioner class name fields actually refer to the `metadata.name` or ProvisionerClass objects.

PR #483, opened 2022-08-08
Add ConfigMapSyncer controller and rukpak-ca configmap

### LINKED ISSUE #475: Rukpak clients should be able to get CA cert without getting access to CA key
We use cert-manager to generate a rukpak-wide CA, and then use that CA to sign all of the certs for rukpak services. Clients of rukpak then need to read that CA to be able to verify connections to those rukpak services.

Currently, the situation is that clients read the `rukpak-ca` secret's `data["ca.crt"]` value to get the CA cert. However, that secret also contains the CA's key, which clients could use to sign more keys that would be trusted by anyone using the CA.

We need to ensure that the object read by clients contains _only_ the CA cert. It seems like cert-manager does not support this out of the box (see https://github.com/cert-manager/cert-manager/issues/2722#issuecomment-703766328), so we may need to run our own controller that knows how to inject the CA into a separate configmap.

joelanford:
The ConfigMapSyncer syncs secret data to configmaps based on injection
annotations present in configmaps in watched namespaces.

We include a rukpak-ca configmap with these annotations present so
that cluster administrators can share rukpak-ca trust without
exposing the CA key that's present in the rukpak-ca secret.

Closes #475 

### DIFF
--- a/cmd/core/main.go
+++ b/cmd/core/main.go
@@ -36,11 +36,13 @@ import (
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/cluster"
 	crfinalizer "sigs.k8s.io/controller-runtime/pkg/finalizer"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 
 	rukpakv1alpha1 "github.com/operator-framework/rukpak/api/v1alpha1"
+	"github.com/operator-framework/rukpak/internal/configmapsyncer"
 	"github.com/operator-framework/rukpak/internal/finalizer"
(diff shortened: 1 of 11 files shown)


joelanford:
This PR is two steps forward, but one step back:
1. Two steps forward
    - we're using configmaps instead of secrets in `rukpakctl`, so cluster admins don't have to give secret read permission to rukpakctl users.
    - we're making the build of `rukpakctl` more flexible so that its defaults can be reconfigured if vendors deploy rukpak with a different configuration.
2. One step back:
    - The ConfigMapSyncer (and the e2e test that accompanies it) solves an upstream problem we have using cert-manager to manage rukpak secrets. But rukpak doesn't directly depend on cert-manager, it depends on having certs in the right places. Other deployments of rukpak might use different certificate managers that don't share assumptions made by the ConfigMapSyncer (e.g. the CA bundle might not be in a secret in the rukpak system namespace).

Therefore
/hold
The step back is giving me some pause, and I think we should discuss how to handle it. Two thoughts:
1. Put the ConfigMapSyncer in a separate controller that rukpak distros can simply decide _not_ to include.
2. Leave the ConfigMapSyncer in the core controller, but make it possible to disable it somehow (flag at runtime, build flag that straight up removes the code?)

PR #484, opened 2022-08-08
api validation tests

### LINKED ISSUE #435: e2e webhook bundle test is outdated and potentially brittle
The e2e webhook test `"the bundle name is too long"` is outdated:
* Initially, the bundle name length check was part of the core webhook, and it would return the a custom error indicating the name length is too long. 
* At some point, the name validation was moved directly to the CRD schema, and so the name length error returned changed. The webhook is no longer responsible for validating the length of the bundle name. 
* Testing the entire error string is brittle -- maybe only a subset of the string needs to be compared

A/C:
* Move this validation test out of the webhook_test file and into a more appropriate file
* Compare only a subset of the error message, instead of the entire message, to be less brittle

akihikokuroda:
Closes #435 

Create a new test file api_validation_test.go.  Added some tests and moved some tests from webhook_test.go

### DIFF
--- a/test/e2e/api_validation_test.go
+++ b/test/e2e/api_validation_test.go
@@ -1,0 +1,204 @@
+package e2e
+
+import (
+	"context"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


exdx (approved):
LGTM. I would say we could also have this basic validation implemented in unit tests using envtest, to help our e2e suite run faster. But having it as e2e is fine too.

tylerslaton (approved):
This looks good to me!

PR #491, opened 2022-08-15
fix: Use top-level rukpakctl binary for e2e tests

### LINKED ISSUE #489: rukpakctl e2es fail on Mac workstations
I ran the e2e's this morning and noticed the `rukpakctl` e2e tests fail on Mac workstations, I believe due to the fact that we explicitly use the [`bin/linux/rukpakctl`](https://github.com/operator-framework/rukpak/blob/75e5a0974fa7a6f4d2c0a75dfdca584becbce44b/test/e2e/rukpakctl_test.go#L20) path in the e2e code.

We should make sure that this e2e uses a `rukpakctl` binary that is appropriate for the local workstation.

exdx:
Closes #489

### DIFF
--- a/test/e2e/rukpakctl_test.go
+++ b/test/e2e/rukpakctl_test.go
@@ -17,7 +17,7 @@ import (
 )
 
 const (
-	rukpakctlcmd = "../../bin/linux/rukpakctl "
+	rukpakctlcmd = "../../bin/rukpakctl "
 	testbundles  = "../../testdata/bundles/"
 )
 


test/e2e/rukpakctl_test.go
```diff
@@ -17,7 +17,7 @@ import (
 )
 
 const (
-	rukpakctlcmd = "../../bin/linux/rukpakctl "
+	rukpakctlcmd = "../../bin/rukpakctl "
```
timflannagan, line 20: It would be nice to give the testing suite a base directory vs. constructing the relatively path in the future.

timflannagan (approved):

tylerslaton (approved):

test/e2e/rukpakctl_test.go
joelanford: Does our e2e make task always ensure that the `../../bin/rukpakctl` binary is re-built?

akihikokuroda: Is the rukpakctl built there, too?

joelanford: It depends on how the Makefile is invoked.
- With `make build-container`, we do `export GOOS=linux` because we know that containers will only be running on linux systems. We also build rukpakctl into the rukpak container so that folks can consume it that way if they want to. In this case, we put these binaries in `bin/linux`
- With `make build` we leave GOOS alone and we put binaries in `bin`

I think what we probably need to do in this case is figure out how to:
1. build rukpakctl for the local environment at `./bin/rukpakctl`
2. build the rukpak container (as usual)

IIRC, there was some makefile issue with setting GOOS=linux _just_ in the context of the `build-container` target such that other targets would _not_ inherit that setting.

Issue #167, opened 2022-03-24
Provide validation tooling for plain bundles


exdx:
The plain bundle format is relatively straightforward and documented in #149. Even though it's very intuitive, it's still useful to be able to provide users of the format with some sort of static validation tooling to ensure that their bundle is valid and can successfully be applied to the cluster. Based on the `plain+v0` bundle spec the tooling would do the following:
* Confirm the bundle source is reachable
* Confirm every file in the /manifests directory (image-based bundles) or user-provided directory (git-based bundles) is a valid YAML manifest that can be applied via a Kubernetes client successfully 
* Confirm no extra, non-manifest files are included
* Confirm at least one file exists in the directory provided

asmacdo:
> Confirm every file in the /manifests directory (image-based bundles) or user-provided directory (git-based bundles) is a valid YAML manifest that can be applied via a Kubernetes client successfully

We can easily confirm that each file is valid yaml, but confirming that the manifest can be applied to Kubernetes is a little more complex. I discussed this with folks and @kevinrizza pointed out that at validation time, some kubernetes manifests may NOT be able to be applied because the relevent CRDs are not yet present on cluster.

For initial implementation, I am going to check "valid yaml or json" but not check "applicable via a Kubernetes client".

If we need to check for valid k8s manifests, lets file a new issue.

PR #494, opened 2022-08-16
167 plain bundle validation

asmacdo:
Add plain bundle v0 struct, use initialization for validation
Closes https://github.com/operator-framework/rukpak/issues/167

### DIFF
--- a/internal/bundle/plain.go
+++ b/internal/bundle/plain.go
@@ -1,0 +1,72 @@
+package bundle
+
+import (
+	"errors"
+	"fmt"
+	"io/fs"
+	"path/filepath"
+
+	"k8s.io/cli-runtime/pkg/resource"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type PlainV0 struct {
+	Objects []client.Object
+}
+
+const plainV0ManifestsDir = "manifests"
+
+func LoadPlainV0(bundle fs.FS) (*PlainV0, error) {
+	var objects []client.Object
+	dirEntries, err := fs.ReadDir(bundle, plainV0ManifestsDir)
+	if err != nil {
+		return nil, err
+	}
+	if len(dirEntries) == 0 {
+		return nil, errors.New("invalid bundle: found zero files")
+	}
+	for _, manifest := range dirEntries {
+		if manifest.IsDir() {
+			return nil, fmt.Errorf("subdirectories are not allowed within the %q directory "+
+				"of the bundle image filesystem: found %q", plainV0ManifestsDir, filepath.Join(plainV0ManifestsDir, manifest.Name()))
+		}
+		theseObjects, err := getObjects(bundle, manifest)
+		if err != nil {
+			return nil, err
+		}
+		objects = append(objects, theseObjects...)
+	}
+	if len(objects) == 0 {
+		return nil, errors.New("invalid bundle: found zero objects: " +
+			"plain+v0 bundles are required to contain at least one object")
+	}
+	return &PlainV0{Objects: objects}, nil
+}
+
+func getObjects(bundle fs.FS, manifest fs.DirEntry) ([]client.Object, error) {
+	manifestPath := filepath.Join(plainV0ManifestsDir, manifest.Name())
+	manifestReader, err := bundle.Open(manifestPath)
+	if err != nil {
+		return nil, err
+	}
+	defer manifestReader.Close()
+	// Stream().Do() visits and serializes each object
+	result := resource.NewLocalBuilder().Flatten().Unstructured().Stream(manifestReader, manifestPath).Do()
+	if err := result.Err(); err != nil {
+		return nil, err
+	}
+	infos, err := result.Infos()
+	if err != nil {
+		return nil, err
+	}
+	return infosToObjects(infos), nil
+}
+
+func infosToObjects(infos []*resource.Info) []client.Object {
+	objects := make([]client.Object, 0, len(infos))
+	for _, info := range infos {
+		clientObject := info.Object.(client.Object)
+		objects = append(objects, clientObject)
+	}
+	return objects
+}
--- a/internal/bundle/plain_test.go
+++ b/internal/bundle/plain_test.go
@@ -1,0 +1,121 @@
+package bundle
+
+import (
+	"io/fs"
+	"testing/fstest"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("LoadPlainV0", func() {
+	It("should fail if there is no manifests directory", func() {
+		mockfs := fstest.MapFS{}
+		_, err := LoadPlainV0(mockfs)
+		Expect(err).To(MatchError(ContainSubstring("file does not exist")))
+	})
+	It("should fail if there is an empty manifests directory", func() {
+		mockfs := fstest.MapFS{
+			"manifests": {
+				Mode: fs.ModeDir,
+			},
+		}
+		_, err := LoadPlainV0(mockfs)
+		Expect(err).To(MatchError(ContainSubstring("found zero files")))
+	})
+	It("should fail if there is a file, but no yaml objects", func() {
+		mockfs := fstest.MapFS{
+			"manifests/invalid.yaml": {
+				Data: []byte(""),
+			},
+		}
+		_, err := LoadPlainV0(mockfs)
+		Expect(err).To(MatchError(ContainSubstring("required to contain at least one object")))
+	})
+	It("should fail if any of the files are invalid json or yaml", func() {
+		mockfs := fstest.MapFS{
+			"manifests/valid.yaml": {
+				Data: []byte("kind: ofnecessary"),
+			},
+			"manifests/invalid.yaml": {
+				Data: []byte("not;valid"),
+			},
+		}
+		_, err := LoadPlainV0(mockfs)
+		Expect(err).To(MatchError(ContainSubstring("cannot unmarshal string into Go value")))
+	})
+	It("should fail if the YAML or JSON objects do not specify `kind`", func() {
+		mockfs := fstest.MapFS{
+			"manifests/invalid.yaml": {
+				Data: []byte("almost: valid"),
+			},
+		}
+		_, err := LoadPlainV0(mockfs)
+		Expect(err).To(MatchError(ContainSubstring("Object 'Kind' is missing")))
+	})
+	It("should pass if the YAML is valid and contains a `kind` element", func() {
+		mockfs := fstest.MapFS{
+			"manifests/valid.yaml": {
+				Data: []byte("kind: ofnecessary"),
+			},
+		}
+		_, err := LoadPlainV0(mockfs)
+		Expect(err).To(BeNil())
+	})
+	It("should pass if the JSON is valid and contains a `kind` element", func() {
+		mockfs := fstest.MapFS{
+			"manifests/valid-json.yaml": {
+				Data: []byte(`{"kind": "ofnecessary"}`),
+			},
+		}
+		plainv0, err := LoadPlainV0(mockfs)
+		Expect(len(plainv0.Objects)).To(Equal(1))
+		Expect(err).To(BeNil())
+	})
+	It("should pass if some manifests are JSON and some are YAML", func() {
+		mockfs := fstest.MapFS{
+			"manifests/valid-json.yaml": {
+				Data: []byte(`{"kind": "ofnecessary"}`),
+			},
+			"manifests/valid-yaml.yaml": {
+				Data: []byte(`kind: ofnecessary`),
+			},
+		}
+		plainv0, err := LoadPlainV0(mockfs)
+		Expect(len(plainv0.Objects)).To(Equal(2))
+		Expect(err).To(BeNil())
+	})
+
+	It("should fail if there are nested subdirectories", func() {
+		mockfs := fstest.MapFS{
+			"manifests/valid.yaml": {
+				Data: []byte("kind: ofnecessary"),
+			},
+			"manifests/invalid/valid.yaml": {
+				Data: []byte("kind: illegalsubdir"),
+			},
+		}
+		_, err := LoadPlainV0(mockfs)
+		Expect(err).To(MatchError(ContainSubstring("subdirectories are not allowed")))
+	})
+	It("should pass with multi-object yaml files", func() {
+		mockfs := fstest.MapFS{
+			"manifests/valid.yaml": {
+				Data: []byte("kind: ofnecessary\n---\nkind: me"),
+			},
+		}
+		plainv0, err := LoadPlainV0(mockfs)
+		Expect(len(plainv0.Objects)).To(Equal(2))
+		Expect(err).To(BeNil())
+	})
+	It("should flatten and pass with metav1.List yaml files", func() {
+		mockfs := fstest.MapFS{
+			"manifests/valid.yaml": {
+				Data: []byte("kind: List\napiVersion: v1\nitems:\n  - kind: me\n    apiVersion: v1\n  - kind: you\n    apiVersion: v1"),
+			},
+		}
+		plainv0, err := LoadPlainV0(mockfs)
+		Expect(len(plainv0.Objects)).To(Equal(2))
+		Expect(err).To(BeNil())
+	})
+})
--- a/internal/provisioner/plain/controllers/bundle_controller.go
+++ b/internal/provisioner/plain/controllers/bundle_controller.go
@@ -17,22 +17,15 @@ limitations under the License.
 package controllers
 
 import (
-	"bytes"
 	"context"
-	"errors"
 	"fmt"
-	"io"
-	"io/fs"
-	"path/filepath"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	apimacherrors "k8s.io/apimachinery/pkg/util/errors"
-	apimachyaml "k8s.io/apimachinery/pkg/util/yaml"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -41,6 +34,7 @@ import (
 	crsource "sigs.k8s.io/controller-runtime/pkg/source"
 
 	rukpakv1alpha1 "github.com/operator-framework/rukpak/api/v1alpha1"
+	bundlelib "github.com/operator-framework/rukpak/internal/bundle"
 	plain "github.com/operator-framework/rukpak/internal/provisioner/plain/types"
 	"github.com/operator-framework/rukpak/internal/source"
 	"github.com/operator-framework/rukpak/internal/storage"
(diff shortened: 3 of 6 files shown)


cmd/core/main.go
```diff
@@ -205,12 +206,19 @@ func main() {
 		os.Exit(1)
 	}
 
+	plainValidator, err := validators.NewPlainValidator()
+	if err != nil {
+		setupLog.Error(err, "unable to setup bundle validator")
+		os.Exit(1)
+	}
```
joelanford, line 213: Do we plan to use a different validator with the plain provisioner? If not, I'd suggest hiding this as an implementation detail of the plain bundle reconciler rather than making it a configurable thing on the struct.

cmd/core/main.go
```diff
@@ -205,12 +206,19 @@ func main() {
 		os.Exit(1)
 	}
 
+	plainValidator, err := validators.NewPlainValidator()
```
joelanford, line 209: IMO, anything specific to the plain provisioner should go in the `internal/proivisioner/plain` tree somewhere.

internal/validators/base.go
```diff
@@ -0,0 +1,53 @@
...
+	apimachyaml "k8s.io/apimachinery/pkg/util/yaml"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type Validator interface {
+	Validate(fs.FS) error
+}
+
+// TODO(asmacdo) I moved this *and* made it public, is this a problem?
+func GetObjects(bundleFS fs.FS) ([]client.Object, error) {
```
joelanford, line 21: Do we use this anywhere outside the plain provisioner (I think not since it has knowledge and makes assumptions about the FS that's passed in)?

If it is specific to the plain provisioner, I think it should be internal and within the plain package tree.

internal/validators/base.go
```diff
@@ -0,0 +1,53 @@
...
+	"path/filepath"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	apimachyaml "k8s.io/apimachinery/pkg/util/yaml"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type Validator interface {
+	Validate(fs.FS) error
+}
```
joelanford, line 18: This might be a premature interface definition. I'd expect to see this crop up only when there's a use case around passing different validator implementations around to the same provisioner.

internal/validators/plain.go
```diff
@@ -0,0 +1,41 @@
...
+	validationErrors = append(validationErrors, errors.New("hello"))
+	validationErrors = append(validationErrors, errors.New("world"))
+	validationErrors = append(validationErrors, errors.New("annnndd"))
+	validationErrors = append(validationErrors, errors.New("four Errors!"))
+	validationErrors = append(validationErrors, errors.New("moved this shit around!"))
+
+	err = errors.New("")
+	for _, each := range validationErrors {
+		err = fmt.Errorf("%v\n%v", err, each)
+	}
```
joelanford, line 39: This seems like it will always result in a non-nil error, which seems undesirable. Perhaps consider using https://pkg.go.dev/k8s.io/apimachinery/pkg/util/errors#Aggregate

asmacdo: ah thanks, I was looking for errors.Wrap and didn't see it

cmd/core/main.go
asmacdo: I wasn't sure, but I imagined but I figured the registry provisioner (or any other "conversion" provisioners) would use this as well.

I may have over engineered, but given the recent history of operator-lib validations I am imagining that the reconcilers may need to use different versions of the same validators, or that reconcilers might share validators.

timflannagan (commented):
Just did a quick run through of these changes, and I mainly had the same feedback as Joe. I think my preference would be to avoid making the first pass implementation here generalistic, and instead just focusing on the plain provisioner use case without exposing too much of this through configuration.

In terms of exposing a knob on the plain provisioner for configuring validators: would we need a way for admins to opt-out to this validation? Outside of insufficient testing for our validation logic, is there any way that an admin would want/need to be able to opt-out of these validations in the future?

internal/validators/base.go
asmacdo: bundle deployment uses it as well https://github.com/operator-framework/rukpak/blob/0f45b79342c53aab4e1ee4dee1958f7c95e3e422/internal/provisioner/plain/controllers/bundledeployment_controller.go#L436

tylerslaton:
@asmacdo Another thing, could we include a `Closes #167` in the description so that automation closes the issue out when this merges?

asmacdo:
@tylerslaton it's in the commit message, does "Fixed" work also?

tylerslaton:
> @tylerslaton it's in the commit message, does "Fixes" work also?

Ah, I see what you're saying. I think that the [automation](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword#linking-a-pull-request-to-an-issue-using-a-keyword) should detect the linking if you remove the `operator-framework` prefix or change it to be `operator-framework/rukpak#167`. You'll know if it works when the issue is linked in the `Development` section of the right hand bar.

cmd/core/main.go
joelanford: > I imagined the registry provisioner (or any other "conversion" provisioners) would use this as well.

I could see that as a sort of post-conversion step perhaps, but if our registry+v1 to plain+v0 conversion resulted in a failure in the plain bundle validation, that'd be a bug on our side, not really something that would be useful/actionable for a user.

Perhaps it's worth structuring it such that the registry+v1 to plain+v0 conversion tests can call it as a sanity check?

joelanford:
> would we need a way for admins to opt-out to this validation? 

I'd imagine that the level of validation we should do here is "will the bundle deployment provisioner fail to parse this bundle? If yes, validation error. If no, no error". In that case, there's no reason to give cluster admins a knob for it.

If we ever get into more linting kinds of validations, we would want cluster admin knobs. But IMO linting shouldn't be done at runtime.

internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -229,6 +223,32 @@ func updateStatusUnpackFailing(u *updater.Updater, err error) error {
 	return err
 }
 
+func (r *BundleReconciler) validate(bundle fs.FS) error {
+	var validationErrors []error
+	objects, err := getObjects(bundle)
+	if err != nil {
+		validationErrors = append(validationErrors, fmt.Errorf("get objects from bundle manifests: %v", err))
+	}
+	if len(objects) == 0 {
```
joelanford, line 232: `else if` here?

I assume that if we encounter an error getting the objects, the length of `objects` will always be 0.

internal/provisioner/plain/validator/validator.go
```diff
@@ -0,0 +1,60 @@
...
+	if err != nil {
+		validationErrors = append(validationErrors, fmt.Errorf("get objects from bundle manifests: %v", err))
+	} else if len(objects) == 0 {
+		validationErrors = append(validationErrors, errors.New("invalid bundle: found zero objects: "+
+			"plain+v0 bundles are required to contain at least one object"))
+	}
+	return apimacherrors.NewAggregate(validationErrors)
+}
+
+func GetObjects(bundleFS fs.FS) ([]pkgclient.Object, error) {
```
joelanford, line 29: Seems a bit odd to me that `GetObjects` is in the `validate` package. Perhaps we should have a `internal/bundle/plain/` package that contains a `LoadBundleV0(fs.FS) (*BundleV0, error)` function.

And then all of the logic for enforcing the plain+v0 bundle spec would be contained there (and could still be abstracted out into a separate set of validations functions)?

asmacdo:
@joelanford I attempted to use the cli-runtime library to serialize and validate the json/yaml. I was able to get this to work for validation purposes.

However, this turned out to be a dead end (AFAICT) because the [`Info.Objects`](https://github.com/kubernetes/cli-runtime/blob/v0.25.0/pkg/resource/visitor.go#L82) are `k8s.io/apimachinery/pkg/runtime.Objects` rather than `sigs.k8s.io/controller-runtime/pkg/client.Objects`. The bundle deployment controller uses `client.Object.GetLabels()` which is not implemented in the `runtime.Object`

Therefore, I've opted to keep the `getObjects` code (renamed and moved to `LoadV0`) rather than refactor it to use the cli-runtime library.

joelanford:
@asmacdo, we can type assert those `runtime.Object`s to `client.Object`s.

internal/bundle/plain.go
```diff
@@ -0,0 +1,57 @@
...
+			return PlainV0{}, fmt.Errorf("subdirectories are not allowed within the %q directory "+
+				"of the bundle image filesystem: found %q", manifestsDir, filepath.Join(manifestsDir, manifest.Name()))
+		}
+		manifestReader, err := bundle.Open(filepath.Join("manifests", manifest.Name()))
+		if err != nil {
+			return PlainV0{}, err
+		}
+		defer manifestReader.Close()
+		// Stream().Do() visits and serializes each object
+		result := resource.NewLocalBuilder().Flatten().Unstructured().Stream(manifestReader, "Load PlainV0 Bundle").Do()
```
joelanford, line 38: Rather than using `"Load PlainV0 Bundle"`, I think the intention of that string is to be something descriptive about the `manifestReader`. I'd suggest extracting `filepath.Join("manifests", manifest.Name())` as a variable and using that when we `Open` and here.

internal/bundle/plain.go
```diff
@@ -0,0 +1,57 @@
...
+	for _, manifest := range dirEntries {
+		if manifest.IsDir() {
+			return PlainV0{}, fmt.Errorf("subdirectories are not allowed within the %q directory "+
+				"of the bundle image filesystem: found %q", manifestsDir, filepath.Join(manifestsDir, manifest.Name()))
+		}
+		manifestReader, err := bundle.Open(filepath.Join("manifests", manifest.Name()))
+		if err != nil {
+			return PlainV0{}, err
+		}
+		defer manifestReader.Close()
```
joelanford, line 36: Nit: `defer`-ing in a loop results in all of the deferred functions running at the end of the function rather than immediately after we've consumed that specific file. Can we extract this loop code into a separate function, basically:
```go
for _, manifest := range dirEntries {
	manifestObjects, err := getObjects(bundle, manifest)
	if err != nil {
		return nil, err
	}
	objects = append(objects, manifestObjects...)
}
```

That way, when you `defer` in `getObjects`, you end up actually closing the file before moving on to the next file.

joelanford: We could also potentially do 
```go
fs.WalkDir(bundle, manifestsDir, func(path string, d fs.DirEntry, err error) error {
	// handle individual dir entry in this closure.
})
```

internal/bundle/plain_test.go
```diff
@@ -0,0 +1,96 @@
...
+				Data: []byte("kind: ofnecessary"),
+			},
+			"manifests/invalid/valid.yaml": {
+				Data: []byte("kind: illegalsubdir"),
+			},
+		}
+		_, err := LoadPlainV0(mockfs)
+		Expect(err.Error()).To(ContainSubstring("subdirectories are not allowed"))
+	})
+	It("should pass with multi-object yaml files", func() {
```
joelanford, line 87: Can you also add tests for:
1. All manifest files contain JSON
2. Manifest files are a mix of JSON and YAML (but each file is one or the other)
3. The ability to parse and flatten `v1.List` objects that contain a list of `items` that are objects.

internal/provisioner/plain/controllers/bundledeployment_controller.go
```diff
@@ -370,14 +371,13 @@ func (r *BundleDeploymentReconciler) loadBundle(ctx context.Context, bundle *ruk
...
+	plainV0, err := bundleLib.LoadPlainV0(bundleFS)
 	if err != nil {
 		return nil, fmt.Errorf("read bundle objects from bundle: %v", err)
 	}
 
-	objs := make([]client.Object, 0, len(objects))
-	for _, obj := range objects {
-		obj := obj
+	objs := make([]client.Object, 0, len(plainV0.Objects))
+	for _, obj := range plainV0.Objects {
```
joelanford, line 380: Any reason we can't/shouldn't modify the `plainV0.Objects` slice directly? It would use less memory if we modified the objects in place.

joelanford (commented):
Looks good overall. Just left a handful of suggestions for minor improvements.

internal/bundle/plain_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+import (
+	"io/fs"
+	"testing/fstest"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Plain bundle v0 loader", func() {
```
tylerslaton, line 11: @joelanford @timflannagan Do we have a preference for using Ginkgo vs Go's standard testing framework when unit testing? We use Go's standard framework for some instance and Ginkgo for others .

joelanford: I personally prefer ginkgo, but I realize its a contentious debate. I'm more interested in consistency and ease of adding new test cases to existing test suites than I am about any particular framework.

internal/provisioner/plain/controllers/bundledeployment_controller.go
asmacdo: at some point i was doing the type assert here, this was just an artifact of that. good catch!

internal/bundle/plain.go
joelanford: Looks like this one got lost in the latest push?

internal/bundle/plain_test.go
```diff
@@ -0,0 +1,124 @@
...
+		Expect(err).To(BeNil())
+	})
+	It("should pass if the JSON is valid and contains a `kind` element", func() {
+		mockfs := fstest.MapFS{
+			"manifests/valid-json.yaml": {
+				Data: []byte(`{"kind": "ofnecessary"}`),
+			},
+			"manifests/valid.yaml": {
+				Data: []byte("kind: ofnecessary"),
+			},
```
joelanford, line 72: Seems like this test is identical to the one below. Looks like this is supposed to test a JSON file by itself.
```suggestion
			"manifests/valid.json": {
				Data: []byte(`{"kind": "ofnecessary"}`),
			},
```

internal/bundle/plain_test.go
```diff
@@ -0,0 +1,124 @@
...
+	It("should flatten and pass with metav1.List yaml files", func() {
+		mockfs := fstest.MapFS{
+			"manifests/valid.yaml": {
+				Data: []byte("kind: List\napiVersion: v1\nitems:\n  - kind: me\n    apiVersion: v1\n  - kind: you\n    apiVersion: v1"),
+			},
+		}
+		plainv0, err := LoadPlainV0(mockfs)
+		Expect(len(plainv0.Objects)).To(Equal(2))
+		Expect(err).To(BeNil())
+	})
```
joelanford, line 123: Another potential test: 2 files, one is valid YAML, the other is empty.

From the code that exists right now, that looks valid. But I wonder if we should call that invalid (i.e. every file in the manifests dir MUST have at least one object)

internal/provisioner/plain/controllers/bundledeployment_controller.go
```diff
@@ -46,6 +46,7 @@ import (
 	"sigs.k8s.io/yaml"
 
 	rukpakv1alpha1 "github.com/operator-framework/rukpak/api/v1alpha1"
+	bundleLib "github.com/operator-framework/rukpak/internal/bundle"
```
joelanford, line 49: nit: drop the import alias and just use the `bundle` package name in the code.

if i'm missing another import already called `bundle` that requires this alias, i'd suggest `bundlelib` all lowercase. `bundleLib` makes mind mind drift to "that's a variable" rather than "that's a package".

internal/bundle/plain.go
```diff
@@ -0,0 +1,66 @@
...
+		objects = append(objects, *theseObjects...)
+	}
+	if len(objects) == 0 {
+		return nil, errors.New("invalid bundle: found zero objects: " +
+			"plain+v0 bundles are required to contain at least one object")
+	}
+	return &PlainV0{Objects: objects}, nil
+}
+
+func getObjects(bundle fs.FS, manifest fs.DirEntry) (*[]client.Object, error) {
```
joelanford, line 45: No need to return a pointer to a slice. A slice is already a pointer (so you can continue returning `nil, err`).
```suggestion
func getObjects(bundle fs.FS, manifest fs.DirEntry) ([]client.Object, error) {
```

internal/bundle/plain.go
```diff
@@ -0,0 +1,66 @@
...
+	manifestReader, err := bundle.Open(filepath.Join("manifests", manifest.Name()))
```
timflannagan, line 46: nit: avoid hardcoding the "manifests" directory and instead use a constant?

internal/bundle/plain.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+	infos, err := result.Infos()
+	if err != nil {
+		return nil, err
+	}
+	objects := make([]client.Object, 0, len(infos))
+	for _, info := range infos {
+		clientObject := info.Object.(client.Object)
+		objects = append(objects, clientObject)
+	}
```
timflannagan, line 64: You might be able to modularize this better, and move this logic into it's own function that returns an array of client.Object(s).

internal/bundle/plain_test.go
```diff
@@ -0,0 +1,124 @@
...
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Plain bundle v0 loader", func() {
+	It("should fail if there is no manifests directory", func() {
+		mockfs := fstest.MapFS{}
+		_, err := LoadPlainV0(mockfs)
+		Expect(err.Error()).To(ContainSubstring("file does not exist"))
```
timflannagan, line 15: It looks like we're fairly consistent with using `Expect(err).To(MatchError(ContainSubstring(...))` throughout ginkgo-related tests that perform this kind of testing check.

internal/bundle/plain.go
```diff
@@ -0,0 +1,66 @@
+package bundle
```
timflannagan, line 1: This might be a good candidate for it's own PR given we're moving around the getObjects function here.

timflannagan (commented):
Quick review. Looks like we're heading in the right direction.

internal/bundle/plain.go
asmacdo: nice catch thanks

internal/provisioner/plain/controllers/bundledeployment_controller.go
asmacdo: actually BundleReploymentReconciler.loadBundle has `bundle` as an arg. IMO that param name makes sense, so I'm keeping `bundleLib`

internal/bundle/plain.go
```diff
@@ -14,8 +14,9 @@ type PlainV0 struct {
 	Objects []client.Object
 }
 
+const manifestsDir = "manifests"
```
joelanford, line 17: nit: this is a variable that applies to the whole bundle package, so if its going to be a global constant, i'd name it something like `plainv0ManifestsDir`

internal/bundle/plain_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+import (
+	"io/fs"
+	"testing/fstest"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Plain bundle v0 loader", func() {
```
joelanford, line 11: Nit: pretty subjective, but when a Describe is clearly scoped to a particular function or type, I typically use the function name as the describe string.
```go
var _ = Describe("LoadPlainV0", func() { ... })
```

tylerslaton:
@asmacdo Mind rebasing this? I think this is likely ready to go in unless @timflannagan has any outstanding comments.

asmacdo:
Joes refactor seems to have done this generally, so no need to do it for plain bundles like this.

PR #497, opened 2022-08-17
unpacker export

### LINKED ISSUE #478: refactor NewDefaultUnpacker due to growing parameter list
https://github.com/operator-framework/rukpak/blob/3b832db489499105211cda18f2217aad89e2bb8f/internal/source/unpacker.go#L103

```golang

// NewDefaultUnpacker returns a new composite Source that unpacks bundles using
// a default source mapping with built-in implementations of all of the supported
// source types.
//
// TODO: refactor NewDefaultUnpacker due to growing parameter list
func NewDefaultUnpacker(mgr ctrl.Manager, namespace, provisionerName, unpackImage string, baseUploadManagerURL string, rootCAs *x509.CertPool) (Unpacker, error) {
	cfg := mgr.GetConfig()
	kubeClient, err := kubernetes.NewForConfig(cfg)
	if err != nil {
		return nil, err
	}
	httpTransport := http.DefaultTransport.(*http.Transport).Clone()
	if httpTransport.TLSClientConfig == nil {
		httpTransport.TLSClientConfig = &tls.Config{}
	}
	httpTransport.TLSClientConfig.RootCAs = rootCAs
	return NewUnpacker(map[rukpakv1alpha1.SourceType]Unpacker{
		rukpakv1alpha1.SourceTypeImage: &Image{
			Client:          mgr.GetClient(),

```

akihikokuroda:
This is part of #425
Closes #478 

This PR move some part (unpack.go) of source package under pkg.  The unpack.go is split into pkg/source/ and pkg/util to avoid circular references.


pkg/util/unpacker.go
```diff
@@ -0,0 +1,101 @@
...
+	Manager ctrl.Manager
+	// ConfigNamespace is key of the config for namespace string
+	Namespace string
+	// ConfigUnpackImage is key of the config for unpack image name string
+	UnpackImage string
+	// ConfigBaseUploadManagerURL is key of the config for base upload manager URL string
+	BaseUploadManagerURL string
+	// ConfigRootCAs is key of the config for *x509.CertPool value for Root CAs
+	RootCAs *x509.CertPool
+}
```
joelanford, line 53: This struct-based approach was discussed, as was the functional options pattern in #478.

The reason we use individual function parameters now is that every field is required. And the reason we're concerned with that function signature is because that list of parameters is growing. We've been adding more required fields.

With the current approach on `main`, we break the API by adding a new parameter, thus causing the compiler to complain at compile time.

With this approach, we'd add a new field to the struct, and existing code would still compile even if it was using the new struct definition. And a provisioner developer wouldn't find out until runtime that something went wrong. Maybe not the end of the world, but less good than what we have now.

FWIW, this Go language proposal for required struct fields resulted in the recommendation of "just use a constructor", which is what we're doing now.

So perhaps our existing `NewDefaultUnpacker` function is actually fine?

joelanford:
I also think we should resolve https://github.com/operator-framework/rukpak/issues/54 before we merge a change that brings in an exported API.

pkg/util/unpacker.go
akihikokuroda: OK.  I'll put the `NewDefaultUnpacker` back.  So the provisioners must be updated and rebuilt when the new source with additional configuration added.

internal/source/upload.go
```diff
@@ -9,37 +9,38 @@ import (
...
 )
 
 // Upload is a bundle source that sources bundles from the rukpak upload service.
 type Upload struct {
-	baseDownloadURL string
-	bearerToken     string
-	client          http.Client
+	BaseDownloadURL string
+	BearerToken     string
+	Client          http.Client
```
timflannagan, line 19: I had a similar comment in another one of the exporting an internal library PRs: does it make sense to create a constructor function here for clients that need access to the upload struct? It would allow us to keep these fields unexported and the fields would be immutable after assignment.

timflannagan (commented):
This also looks pretty close to the finish line.

> I also think we should resolve https://github.com/operator-framework/rukpak/issues/54 before we merge a change that brings in an exported API.

This also seems like a good idea.

internal/source/upload.go
akihikokuroda: I believe that the `upload.go` is under the `internal` directory so they are not exported to the other repository.  It's a good idea to make it  immutable.   I'll add the constructor.

timflannagan:
Can we rebase this now that go-apidiff has been added?

akihikokuroda:
Rebase done.

pkg/source/util/util.go
```diff
@@ -0,0 +1,81 @@
+package util
```
tylerslaton, line 1: Can we move this to `pkg/source` instead? I don't believe that we need a separate util package when the code aligns pretty closely with `pkg/source/unpacker.go`

tylerslaton (approved):
This looks good to me. I won't block it but I did have a suggestion.

pkg/source/util/util.go
akihikokuroda: @tylerslaton   Thanks for review. The util/util.go imports internal/source and go files in internal/source import pkg/source so the util/util.go cannot be in the pkg/source.  It causes circular dependencies. So it is put in a separate package.

awgreene:
@timflannagan @joelanford is this something we still want to pursue doing?

PR #500, opened 2022-08-18
export storage.go

akihikokuroda:
This is part of https://github.com/operator-framework/rukpak/issues/425

This PR export the storage API and the constructor.


pkg/util/storage.go
```diff
@@ -0,0 +1,56 @@
...
+	ctrl "sigs.k8s.io/controller-runtime"
+
+	"github.com/operator-framework/rukpak/internal/storage"
+	"github.com/operator-framework/rukpak/internal/util"
+	pkgstorage "github.com/operator-framework/rukpak/pkg/storage"
+)
+
+const DefaultBundleCacheDir = "/var/cache/bundles"
+
+func NewBundleStorage(mgr ctrl.Manager, provisionerStorageDirectory, httpExternalAddr, bundleCAFile, bearerToken string) (pkgstorage.Storage, error) {
```
joelanford, line 33: IMO, this should go in the `pkg/storage` package.

pkg/util/storage.go
```diff
@@ -0,0 +1,56 @@
...
+	storageURL, err := url.Parse(fmt.Sprintf("%s/bundles/", httpExternalAddr))
```
joelanford, line 34:
```suggestion
	storageURL, err := url.Parse(fmt.Sprintf("%s/%s/", httpExternalAddr, bundlesPath))
```

We should probably make this configurable because the caller has to know this path when it call the AddExtraHandler function.

joelanford: Oh and we should probably use https://pkg.go.dev/net/url#JoinPath instead of fmt.Sprintf here.

joelanford: Looking at the `url.JoinPath` code it looks like we'd be better suited to do this:
```go
storageURL, err := url.Parse(httpExternalAddr)
...
storageURL.JoinPath(bundlePath)
```

internal/storage/storage_test.go
```diff
@@ -15,6 +15,7 @@ import (
 
 	rukpakv1alpha1 "github.com/operator-framework/rukpak/api/v1alpha1"
 	"github.com/operator-framework/rukpak/internal/util"
+	pkgstorage "github.com/operator-framework/rukpak/pkg/storage"
```
tylerslaton, line 18: Should we move this test into the public storage package?

pkg/util/storage.go
joelanford, line 33: A few more improvements I'd suggest:
- Use the functional options pattern for bundleCAFile and bearerToken
- For bundleCAFile, accept an `io.Reader` instead of a filename. That will make it easier to adapt this to other use cases (e.g. unit testing)

akihikokuroda: I'm trying to avoid the `import cycle` like this.  
```
package github.com/operator-framework/rukpak/cmd/core
	imports github.com/operator-framework/rukpak/internal/finalizer
	imports github.com/operator-framework/rukpak/pkg/storage
	imports github.com/operator-framework/rukpak/internal/storage
	imports github.com/operator-framework/rukpak/pkg/storage: import cycle not allowed
```

`NewBundleStorage` refers to both internal/storage and pkg/storage.  It causes the import cycle.  What is the good way to avoid it?  @joelanford

akihikokuroda: oh, I see.  I'll move `pkg/util/storage.go` to `pkg/storage/util/util.go`

pkg/util/storage.go
akihikokuroda: I'll make changes.  Thanks!

internal/storage/storage_test.go
akihikokuroda: I tried but it causes import cycle.  I'll leave it here.
```
# github.com/operator-framework/rukpak/pkg/storage
package github.com/operator-framework/rukpak/pkg/storage
        imports github.com/operator-framework/rukpak/internal/storage: import cycle not allowed in test
FAIL    github.com/operator-framework/rukpak/pkg/storage [setup failed]
FAIL

```

pkg/util/storage.go
timflannagan: Heavy +1 on the io.Reader suggestion here.

pkg/util/storage.go
akihikokuroda: `url.JoinPath` is supported go1.19 or later.  I changed not to use it for now.

pkg/util/storage.go
akihikokuroda: I updated the bundleCAFile to `io.Reader" and to use the functional options pattern.

tylerslaton:
@akihikokuroda Could we get a rebase on this so that we can review it again to get it through?

cmd/core/main.go
```diff
@@ -135,40 +134,28 @@ func main() {
...
-
-	localStorage := &storage.LocalDirectory{
-		RootDirectory: provisionerStorageDirectory,
-		URL:           *storageURL,
-	}
-
+	var bundleCA io.Reader
 	var rootCAs *x509.CertPool
 	if bundleCAFile != "" {
+		bundleCA, err = os.Open(bundleCAFile)
```
timflannagan, line 141: Do we need a defer statement here to ensure we're closing the file descriptor?

cmd/core/main.go
```diff
@@ -135,40 +134,28 @@ func main() {
...
-	httpLoader := storage.NewHTTP(
-		storage.WithRootCAs(rootCAs),
-		storage.WithBearerToken(cfg.BearerToken),
-	)
-	bundleStorage := storage.WithFallbackLoader(localStorage, httpLoader)
-
-	// NOTE: AddMetricsExtraHandler isn't actually metrics-specific. We can run
-	// whatever handlers we want on the existing webserver that
-	// controller-runtime runs when MetricsBindAddress is configured on the
-	// manager.
```
timflannagan, line 169: It looks like this comment was cut during these changes. Do we feel like this comment doesn't provide any value anymore, or isn't still relevant?

cmd/helm/main.go
```diff
@@ -127,31 +130,25 @@ func main() {
...
+		if err != nil {
+			setupLog.Error(err, "unable to create storage")
+			os.Exit(1)
+		}
 		var err error
-		if rootCAs, err = util.LoadCertPool(bundleCAFile); err != nil {
+		if rootCAs, err = util.LoadCertPool(bundleCA); err != nil {
 			setupLog.Error(err, "unable to parse bundle certificate authority file")
 			os.Exit(1)
 		}
```
timflannagan, line 145: This logic might be a potential candidate for a dedicated function if we're using it throughout multiple provisioner implementations.

internal/storage/http.go
```diff
@@ -17,51 +14,8 @@ import (
...
-	s := &HTTP{client: http.Client{
-		Timeout:   time.Minute,
-		Transport: http.DefaultTransport.(*http.Transport).Clone(),
-	}}
-	for _, f := range opts {
-		f(s)
-	}
-	return s
+	Client      http.Client
+	RequestOpts []func(*http.Request)
```
timflannagan, line 18: I'm not opposed to these changes but curious whether we could've kept these unexported if we had a constructor function that was responsible for setting them. You'd still be able to use the HTTP struct through the constructor, but the fields would be immutable after creating that resource.

internal/util/util.go
```diff
@@ -392,13 +393,14 @@ func MergeMaps(maps ...map[string]string) map[string]string {
 	return out
 }
 
-func LoadCertPool(certFile string) (*x509.CertPool, error) {
-	rootCAPEM, err := os.ReadFile(certFile)
+func LoadCertPool(rootCAPEM io.Reader) (*x509.CertPool, error) {
```
timflannagan, line 396: Adding a godoc here would be a good follow-up imo.

pkg/storage/util/util.go
```diff
@@ -0,0 +1,111 @@
...
+	"time"
+
+	ctrl "sigs.k8s.io/controller-runtime"
+
+	"github.com/operator-framework/rukpak/internal/storage"
+	"github.com/operator-framework/rukpak/internal/util"
+	pkgstorage "github.com/operator-framework/rukpak/pkg/storage"
+)
+
+const DefaultBundleCacheDir = "/var/cache/bundles"
```
timflannagan, line 35: Adding a godoc here would be a good follow-up imo.

internal/storage/localdir.go
```diff
@@ -17,9 +17,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/operator-framework/rukpak/internal/util"
+	"github.com/operator-framework/rukpak/pkg/storage"
```
timflannagan, line 20: We may want/need to be consistent with the package import naming here.

pkg/storage/util/util.go
```diff
@@ -0,0 +1,111 @@
...
+	storageURL, err := url.Parse(fmt.Sprintf("%s/%s/", httpExternalAddr, bundlesPath))
+	if err != nil {
+		return nil, err
+	}
+	// URL.JoinPath needs go1.19 or later
+	// storageURL, err := url.Parse(httpExternalAddr)
+	// if err != nil {
+	// 	return nil, err
+	// }
+	// storageURL = storageURL.JoinPath("/", bundlesPath, "/")
```
timflannagan, line 47: Was this a problem in the previous implementation too?

pkg/storage/util/util_test.go
```diff
@@ -176,3 +179,59 @@ func newTLSServer(localStore *LocalDirectory, bearerToken string) *httptest.Serv
...
+		}
+	}
+	if err := fs.WalkDir(a, ".", walkFunc(a, aMap)); err != nil {
+		return false, err
+	}
+	if err := fs.WalkDir(b, ".", walkFunc(b, bMap)); err != nil {
+		return false, err
+	}
+	return reflect.DeepEqual(aMap, bMap), nil
+}
```
timflannagan, line 237: Are these net new changes?

timflannagan (commented):
It seems like we're on the right track here. The diff is fairly large so it was a bit difficult to parse the changeset here so I had a couple of questions on that front.

cmd/core/main.go
akihikokuroda: Yes, I'll add it.

cmd/core/main.go
akihikokuroda: Oh, I restore it.  Thanks!

internal/storage/http.go
akihikokuroda: I tried but I couldn't create the constructor without causing the cycle dependencies.

internal/util/util.go
akihikokuroda: I added a short comment.

pkg/storage/util/util.go
akihikokuroda: I added a short comment.

internal/storage/localdir.go
akihikokuroda: OK.  I put the same name as the others. `pkgstorage`

pkg/storage/util/util_test.go
akihikokuroda: This files is just moved.

pkg/storage/util/util.go
akihikokuroda: No, it wasn't.  This is the suggested implementation in this PR but it can not be implemented unless we update the go to go1.19.

cmd/helm/main.go
akihikokuroda: I created a function in `internal/util/util.go` for this logic.

timflannagan:
Can we rebase this now that go-apidiff has been added?

akihikokuroda:
Rebase done!

PR #517, opened 2022-08-22
add items to dynamic watch in helm

### LINKED ISSUE #514: helm provisioner doesn't dynamically reconcile underlying bundle contents
When deploying rukpak using the main branch tags, it looks like the helm provisioner doesn't have the same dynamic watch capabilities like the plain provisioner does.

Create the following BD resource:

```yaml
kind: BundleDeployment
metadata:
  name: ahoy-tmp
spec:
  provisionerClassName: core-rukpak-io-helm
  template:
    metadata:
      labels:
        app.kubernetes.io/name: ahoy
    spec:
      provisionerClassName: core-rukpak-io-helm
      source:
        http:
          url: https://github.com/helm/examples/releases/download/hello-world-0.1.0/hello-world-0.1.0.tgz
        type: http
```

And then query for the underlying deployment to verify the helm chart installation succeeded:

```bash
$ k get deployments -n rukpak-system ahoy-tmp-hello-world 
NAME                   READY   UP-TO-DATE   AVAILABLE   AGE
ahoy-tmp-hello-world   1/1     1            1           8s
```

And then delete that resource and verify it doesn't get recreated:

```bash
$ k delete deployments -n rukpak-system ahoy-tmp-hello-world 
deployment.apps "ahoy-tmp-hello-world" deleted
$ k get deployments -w -n rukpak-system
NAME               READY   UP-TO-DATE   AVAILABLE   AGE
core               1/1     1            1           34m
helm-provisioner   1/1     1            1           34m
rukpak-webhooks    1/1     1            1           34m
```

akihikokuroda:
Closes #514

### DIFF
--- a/internal/provisioner/bundledeployment/bundledeployment.go
+++ b/internal/provisioner/bundledeployment/bundledeployment.go
@@ -1,9 +1,11 @@
 package bundledeployment
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"strings"
 	"sync"
 
@@ -11,6 +13,7 @@ import (
 	"helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/chart"
 	"helm.sh/helm/v3/pkg/chartutil"
+	"helm.sh/helm/v3/pkg/postrender"
 	"helm.sh/helm/v3/pkg/release"
 	"helm.sh/helm/v3/pkg/storage/driver"
 	"k8s.io/apimachinery/pkg/api/equality"
@@ -21,6 +24,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	apimachyaml "k8s.io/apimachinery/pkg/util/yaml"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -264,7 +268,14 @@ func (p *bundledeploymentProvisioner) reconcile(ctx context.Context, bd *rukpakv
 		return ctrl.Result{}, err
 	}
 
-	rel, state, err := p.getReleaseState(cl, bd, chrt, values)
+	post := &postrenderer{
+		labels: map[string]string{
+			util.CoreOwnerKindKey: rukpakv1alpha1.BundleDeploymentKind,
+			util.CoreOwnerNameKey: bd.GetName(),
+		},
+	}
+
+	rel, state, err := p.getReleaseState(cl, bd, chrt, values, post)
 	if err != nil {
 		meta.SetStatusCondition(&bd.Status.Conditions, metav1.Condition{
 			Type:    rukpakv1alpha1.TypeInstalled,
@@ -280,7 +291,13 @@ func (p *bundledeploymentProvisioner) reconcile(ctx context.Context, bd *rukpakv
 		rel, err = cl.Install(bd.Name, p.releaseNamespace, chrt, values, func(install *action.Install) error {
 			install.CreateNamespace = false
 			return nil
-		})
+		},
+			// To be refactored issue https://github.com/operator-framework/rukpak/issues/534
+			func(install *action.Install) error {
+				post.cascade = install.PostRenderer
+				install.PostRenderer = post
+				return nil
+			})
 		if err != nil {
 			if isResourceNotFoundErr(err) {
 				err = errRequiredResourceNotFound{err}
@@ -294,7 +311,13 @@ func (p *bundledeploymentProvisioner) reconcile(ctx context.Context, bd *rukpakv
 			return ctrl.Result{}, err
 		}
 	case stateNeedsUpgrade:
-		rel, err = cl.Upgrade(bd.Name, p.releaseNamespace, chrt, values)
+		rel, err = cl.Upgrade(bd.Name, p.releaseNamespace, chrt, values,
+			// To be refactored issue https://github.com/operator-framework/rukpak/issues/534
+			func(upgrade *action.Upgrade) error {
+				post.cascade = upgrade.PostRenderer
+				upgrade.PostRenderer = post
+				return nil
+			})
 		if err != nil {
 			if isResourceNotFoundErr(err) {
 				err = errRequiredResourceNotFound{err}
@@ -415,7 +438,7 @@ const (
 	stateError        releaseState = "Error"
 )
 
-func (p *bundledeploymentProvisioner) getReleaseState(cl helmclient.ActionInterface, obj metav1.Object, chrt *chart.Chart, values chartutil.Values) (*release.Release, releaseState, error) {
+func (p *bundledeploymentProvisioner) getReleaseState(cl helmclient.ActionInterface, obj metav1.Object, chrt *chart.Chart, values chartutil.Values, post *postrenderer) (*release.Release, releaseState, error) {
 	currentRelease, err := cl.Get(obj.GetName())
 	if err != nil && !errors.Is(err, driver.ErrReleaseNotFound) {
 		return nil, stateError, err
@@ -426,7 +449,13 @@ func (p *bundledeploymentProvisioner) getReleaseState(cl helmclient.ActionInterf
 	desiredRelease, err := cl.Upgrade(obj.GetName(), p.releaseNamespace, chrt, values, func(upgrade *action.Upgrade) error {
 		upgrade.DryRun = true
 		return nil
-	})
+	},
+		// To be refactored issue https://github.com/operator-framework/rukpak/issues/534
+		func(upgrade *action.Upgrade) error {
+			post.cascade = upgrade.PostRenderer
+			upgrade.PostRenderer = post
+			return nil
+		})
 	if err != nil {
 		return currentRelease, stateError, err
 	}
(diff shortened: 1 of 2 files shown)


joelanford (commented):
This seems to take a different approach than the plain provisioner. Should this with how the plain provisioner works? That would hopefully make it easier to find re-use opportunities in the future, and it would ensure the UX is generally the same for both.

akihikokuroda:
@joelanford The UX should be the same.  It is doing the same thing as the plain provisioner but it can not do the same way.  It is adding to the watch using the release.manifest after the install because the yaml files in the charts can not be decoded without the rendering.  It is adding the labels using the postrenderer because it must be done against the rendered objects.

joelanford:
Oh right. In Plain, we know the full content of the "templates" ahead of time. In Helm, we have to run the chart through the helm libraries to know what the resulting manifest is.

One other thing that comes to mind. We use a post-renderer under the hood in the helm-operator-plugins repo to inject owner references. If we override the post-renderer to set these labels, we might be overriding the ownerref post renderer.

akihikokuroda:
> If we override the post-renderer to set these labels, we might be overriding the ownerref post renderer.

Yes, I found that.  I added the ownerref injection in the post-render function.

joelanford:
Not blocking for this PR, but I imagine that unknowingly setting a post-renderer like you did could be a footgun for others. I wonder if there's a way in helm-operator-plugins to support adding more caller-defined post-renderers to the chain.

akihikokuroda:
I made changes to call the preset PostRenderer after our PostRenderer if it is set.

joelanford:
@akihikokuroda I've got a PR up in helm-operator-plugins that I think should get rid of the awkward hacking you had to do to use both the built-in and extra post-renderer.

internal/provisioner/helm/controllers/bundledeployment_controller.go
```diff
@@ -340,6 +383,54 @@ func (r *BundleDeploymentReconciler) loadChart(ctx context.Context, bundle *rukp
...
+	var objects []client.Object
+	dec := apimachyaml.NewYAMLOrJSONDecoder(bytes.NewReader([]byte(release.Manifest)), 1024)
+	for {
+		obj := unstructured.Unstructured{}
+		err := dec.Decode(&obj)
+		if errors.Is(err, io.EOF) {
+			break
+		}
+		if err != nil {
+			return fmt.Errorf("read %q: %v", obj, err)
```
timflannagan, line 400: Is printing the obj here potentially problem for logs? Or does this just print the metadata information we care about?

internal/provisioner/helm/controllers/bundledeployment_controller.go
```diff
@@ -286,7 +323,7 @@ const (
 	stateError        releaseState = "Error"
 )
 
-func (r *BundleDeploymentReconciler) getReleaseState(cl helmclient.ActionInterface, obj metav1.Object, chrt *chart.Chart, values chartutil.Values) (*release.Release, releaseState, error) {
+func (r *BundleDeploymentReconciler) getReleaseState(cl helmclient.ActionInterface, obj metav1.Object, chrt *chart.Chart, values chartutil.Values, post *postrenderer) (*release.Release, releaseState, error) {
```
timflannagan, line 326: We may want/need a tech debt ticket here given the number of parameters for this method.

internal/provisioner/helm/controllers/bundledeployment_controller.go
```diff
@@ -297,11 +334,17 @@ func (r *BundleDeploymentReconciler) getReleaseState(cl helmclient.ActionInterfa
...
+		func(upgrade *action.Upgrade) error {
+			post.cascade = upgrade.PostRenderer
+			upgrade.PostRenderer = post
+			return nil
+		})
 	if err != nil {
 		return currentRelease, stateError, err
 	}
 	if desiredRelease.Manifest != currentRelease.Manifest ||
+		!reflect.DeepEqual(desiredRelease.Chart.Metadata, currentRelease.Chart.Metadata) ||
```
timflannagan, line 347: Any idea on why this comparison is needed?

internal/provisioner/helm/controllers/bundledeployment_controller.go
```diff
@@ -251,6 +278,16 @@ func (r *BundleDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		}),
 		updater.EnsureInstalledName(bundle.GetName()),
 	)
+	if err := r.addDynamicWatch(cl, bd, bundle); err != nil {
```
timflannagan, line 281: This feels like something the plain provisioner could also hook into?

internal/provisioner/helm/controllers/bundledeployment_controller.go
```diff
@@ -178,7 +189,13 @@ func (r *BundleDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		return ctrl.Result{}, err
 	}
 
-	rel, state, err := r.getReleaseState(cl, bd, chrt, values)
+	post := &postrenderer{
+		labels: map[string]string{
+			util.CoreOwnerKindKey: "BundleDeployment",
```
timflannagan, line 194: nit: I think we have a const variable somewhere for the BD GVK.

timflannagan (commented):
These changes seem reasonable. If we're hacking around the helm-operator-plugins implementation here, does it make sense to create a ticket that tracks cleaning up the implementation once Joe's changes land and a new release is cut?

akihikokuroda:
@timflannagan I created the issue for refactoring. https://github.com/operator-framework/rukpak/issues/534

timflannagan:
Awesome. Follow-up question: do we think it would be valuable to also add a comment linking to that issue for the postrenderer hacks?

akihikokuroda:
That's a good idea.  I'll do it.

internal/provisioner/helm/controllers/bundledeployment_controller.go
akihikokuroda: It doesn't seem defined.

internal/provisioner/helm/controllers/bundledeployment_controller.go
akihikokuroda: This is checking for the changes in Chart.yaml file.  The plain provisioner is generating the contents of the Chart.yam file on the fly so it's always same.

internal/provisioner/helm/controllers/bundledeployment_controller.go
akihikokuroda: This is checking for the changes in Chart.yaml file.  It's contents are [here](https://pkg.go.dev/helm.sh/helm/v3@v3.9.4/pkg/chart#Metadata).

internal/provisioner/helm/controllers/bundledeployment_controller.go
akihikokuroda: OK.  I'll add a TODO comment.

internal/provisioner/helm/controllers/bundledeployment_controller.go
akihikokuroda: The `err` has enough information.  I take out the `obj`.

akihikokuroda:
Refactor has done.  Please review.

internal/provisioner/bundledeployment/bundledeployment.go
```diff
@@ -468,3 +498,51 @@ func isResourceNotFoundErr(err error) bool {
 	//   the use of string comparisons for now.
 	return strings.Contains(err.Error(), "no matches for kind")
 }
+
+// SetupWithManager sets up the controller with the Manager.
+func (p *bundledeploymentProvisioner) SetupWithManager(mgr ctrl.Manager) error {
+	controller, err := ctrl.NewControllerManagedBy(mgr).
+		For(&rukpakv1alpha1.BundleDeployment{}, builder.WithPredicates(
+			util.BundleDeploymentProvisionerFilter(plain.ProvisionerID)),
+		).
```
timflannagan, line 507: I haven't had a chance to look over the new directory structure since #447 landed: we're specifying the plain.ProvisionerID here, and filtering BundleDeployments that don't match that unique provisioner class ID from the cache. Is that the right read here? If so, I think we'll need to solve this in an agnostic way s.t. our core provisioners can leverage this functionality without inter-depending on the plain provisioner libraries.

internal/provisioner/bundledeployment/bundledeployment.go
```diff
@@ -426,7 +450,13 @@ func (p *bundledeploymentProvisioner) getReleaseState(cl helmclient.ActionInterf
 	desiredRelease, err := cl.Upgrade(obj.GetName(), p.releaseNamespace, chrt, values, func(upgrade *action.Upgrade) error {
 		upgrade.DryRun = true
 		return nil
-	})
+	},
+		// To be refactored issue 534
```
timflannagan, line 454: nit: can we copy/paste the 534 issue link so it's easier to track when reading through the codebase locally?

internal/provisioner/bundledeployment/bundledeployment.go
```diff
@@ -264,7 +269,14 @@ func (p *bundledeploymentProvisioner) reconcile(ctx context.Context, bd *rukpakv
 		return ctrl.Result{}, err
 	}
 
-	rel, state, err := p.getReleaseState(cl, bd, chrt, values)
+	post := &postrenderer{
+		labels: map[string]string{
+			util.CoreOwnerKindKey: "BundleDeployment",
```
timflannagan, line 274: nit: we have a constant variable for the BundleDeployment Kind: https://github.com/operator-framework/rukpak/blob/main/api/v1alpha1/bundledeployment_types.go#L26.

internal/provisioner/bundledeployment/bundledeployment.go
akihikokuroda: Oh, it's my mistake.  This function is not used any more since #447 changes.   I'll take this out.

internal/provisioner/bundledeployment/bundledeployment.go
akihikokuroda: Sure.  It must be better.

internal/provisioner/bundledeployment/bundledeployment.go
akihikokuroda: Thanks!  I'll update.

timflannagan:
Sorry for the slow turnaround on this. I've been pretty heads down with other projects/work recently. Can we also re-title this PR to fix the typo?

timflannagan:
Pulled this down and quickly testing it out.

timflannagan (approved):
I tested these changes out locally, and it looks like we're good to go:

<details>

```yaml
kind: BundleDeployment
metadata:
  name: hello-world-helm
  namespace: demo1
spec:
  provisionerClassName: core-rukpak-io-helm
  template:
    metadata:
      labels:
        app: hello-world-helm
    spec:
      provisionerClassName: core-rukpak-io-helm
      source:
        http:
          url: https://github.com/helm/examples/releases/download/hello-world-0.1.0/hello-world-0.1.0.tgz
        type: http
```

</details>

And then deleted the deployment that got created from that Bundle resource a couple of times, and waited for it to be quickly recreated.

I haven't had a chance to catch up the new repository structure, but this looks like a reasonable short term hack while we wait for a release to be cut.

tylerslaton (approved):

PR #521, opened 2022-08-23
add pod watch to helm provisioner bundle controller

### LINKED ISSUE #510: Consider adding a Pod watch for the helm chart bundle controller
The helm chart bundle controller should in theory be able to support git-based source types along with the http sources. The former unpacks git repository contents via an asynchronous Pod, so setting up a watch based on that unpack Pod could help insulate failure states better.

akihikokuroda:
Closes #510 

### DIFF
--- a/internal/provisioner/helm/controllers/bundle_controller.go
+++ b/internal/provisioner/helm/controllers/bundle_controller.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 
+	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


timflannagan (approved):
The only question I had was around the controller.bundle logger name given it conflicts with the plain provisioner's bundle controller name. But I think we're fine here given we're deploying the helm provisioner outside of the core provisioner Pod.

akihikokuroda:
@timflannagan Thanks for review.  
> But I think we're fine here given we're deploying the helm provisioner outside of the core provisioner Pod.

Yes, I thought about it, too.  It is fine at least for now.

PR #525, opened 2022-08-24
add status.observedGeneration in BD

### LINKED ISSUE #504: The helm chart provisioner doesn't trigger a new reconciliation after updating the spec.config field
After deploying rukpak using the main branch images, and creating a BundleDeployment that specifies the helm chart provisioner's unique ID with an empty spec.config configuration, the result is a successful installation of the helm example application. When updating that BundleDeployment resource on the cluster, and then adding the spec.config configuration, it appears that the helm chart provisioner is able to validate the contents of those fields, but I didn't see any obvious changes in the status that would express those changes were persisted in the cluster.

akihikokuroda:
Closes #504

### DIFF
--- a/api/v1alpha1/bundledeployment_types.go
+++ b/api/v1alpha1/bundledeployment_types.go
@@ -66,8 +66,9 @@ type BundleTemplate struct {
 
 // BundleDeploymentStatus defines the observed state of BundleDeployment
 type BundleDeploymentStatus struct {
-	Conditions   []metav1.Condition `json:"conditions,omitempty"`
-	ActiveBundle string             `json:"activeBundle,omitempty"`
+	Conditions         []metav1.Condition `json:"conditions,omitempty"`
+	ActiveBundle       string             `json:"activeBundle,omitempty"`
+	ObservedGeneration int64              `json:"observedGeneration,omitempty"`
 }
 
 //+kubebuilder:object:root=true
--- a/internal/provisioner/helm/controllers/bundledeployment_controller.go
+++ b/internal/provisioner/helm/controllers/bundledeployment_controller.go
@@ -96,6 +96,7 @@ func (r *BundleDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Req
 			l.Error(err, "failed to update status")
 		}
 	}()
+	u.UpdateStatus(updater.EnsureObservedGeneration(bd.Generation))
 
 	bundle, allBundles, err := util.ReconcileDesiredBundle(ctx, r.Client, bd)
 	if err != nil {
--- a/internal/updater/bundle-deployment/updater.go
+++ b/internal/updater/bundle-deployment/updater.go
@@ -69,3 +69,13 @@ func EnsureInstalledName(bundleName string) UpdateStatusFunc {
 		return true
 	}
 }
+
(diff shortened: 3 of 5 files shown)


internal/updater/bundle-deployment/updater.go
```diff
@@ -69,3 +69,10 @@ func EnsureInstalledName(bundleName string) UpdateStatusFunc {
 		return true
 	}
 }
+
+func EnsureIncreaseGeneration() UpdateStatusFunc {
+	return func(status *rukpakv1alpha1.BundleDeploymentStatus) bool {
+		status.ObservedGeneration++
+		return true
+	}
+}
```
joelanford, line 78: Let's make it look like this: https://github.com/operator-framework/rukpak/blob/aaba57539dc3cb75e7f26870ccd3af48ded66707/internal/updater/bundle/updater.go#L80-L88

There's not a guarantee that the controller sees every generation of the object. It's also possible that the same generation of an object will be reconciled multiple times. Therefore, a simple increment like this would result in an incorrect skew between generation and observedGeneration.

internal/provisioner/helm/controllers/bundledeployment_controller.go
```diff
@@ -223,6 +223,9 @@ func (r *BundleDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Req
 				}))
 			return ctrl.Result{}, err
 		}
+		u.UpdateStatus(
+			updater.EnsureIncreaseGeneration(),
+		)
```
joelanford, line 228: We need to _always_ ensure the observed generation matches the generation. We should call this basically immediately after getting the BD, passing in the BD's `metadata.generation`

api/v1alpha1/bundledeployment_types.go
```diff
@@ -66,8 +66,9 @@ type BundleTemplate struct {
 
 // BundleDeploymentStatus defines the observed state of BundleDeployment
 type BundleDeploymentStatus struct {
-	Conditions   []metav1.Condition `json:"conditions,omitempty"`
-	ActiveBundle string             `json:"activeBundle,omitempty"`
+	Conditions         []metav1.Condition `json:"conditions,omitempty"`
+	ActiveBundle       string             `json:"activeBundle,omitempty"`
+	ObservedGeneration int                `json:"observedGeneration,omitempty"`
```
joelanford, line 71:
```suggestion
	ObservedGeneration int64                `json:"observedGeneration,omitempty"`
```

internal/updater/bundle-deployment/updater.go
akihikokuroda: I update.  Thanks!

joelanford (approved):

internal/provisioner/helm/controllers/bundledeployment_controller.go
```diff
@@ -96,6 +96,7 @@ func (r *BundleDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Req
 			l.Error(err, "failed to update status")
 		}
 	}()
+	u.UpdateStatus(updater.EnsureObservedGeneration(bd.Generation))
```
joelanford, line 99: Can you also make this change in the plain provisioner?

Issue #522, opened 2022-08-23
Separate source docs from bundle format docs


joelanford:
Today in the [plain bundle spec](https://github.com/operator-framework/rukpak/blob/main/docs/plain-bundle-spec.md), we talk about sourcing details unrelated to the actual spec of the plain bundle format. For example, the "Technical Details" section talks about images vs git repos.

I think we've intentionally defined a separation between sourcing a bundle and parsing a bundle that we should reflect in our documentation.

IMO, we should re-organize our docs such that we have something along the lines of:

```console
docs
├── bundles
│   ├── helm.md
│   ├── plain.md
│   └── registry.md
├── provisioners
│   ├── helm.md
│   ├── plain.md
│   └── registry.md
└── sources
    ├── git.md
    ├── http.md
    ├── image.md
    ├── local.md
    └── upload.md
```

Where:
- Bundle specs do not mention source implementations
- Source docs do not mention bundle specs
- Provisioner docs link to sources/bundle docs as necessary to explain what they support.

akihikokuroda:
I wonder if it's better to have a provisioners.md file in the docs/provisioners and point to the README.mds in each internal/provisioner/xxxxxx.

PR #516, opened 2022-08-22
Support upload source in helm provisioner

joelanford:
This PR adds support in the helm provisioner for uploaded charts. It also uses a helm library to load the chart, which ensures that the helm provisioner stays up to date with Helm's chart parsing logic.

Closes #513 
Closes #505

Signed-off-by: Joe Lanford <joe.lanford@gmail.com>

PR #535, opened 2022-08-27
add docs/bundles/helm.md

akihikokuroda:
Part of #522 

### DIFF
--- a/docs/bundles/helm.md
+++ b/docs/bundles/helm.md
@@ -1,0 +1,50 @@
+# Helm Bundle Spec
+
+## Overview
+
+This document is meant to define the helm bundle format as a reference for those publishing helm bundles for use with
+RukPak. A bundle is a collection of Kubernetes resources that are packaged together for the purposes of installing onto
+a Kubernetes cluster. A bundle can be unpacked onto a running cluster, where controllers can then create the underlying
+content embedded in the bundle. The bundle can be used as the underlying `spec.source` for
+a [Bundle](https://github.com/operator-framework/rukpak#bundle) resource.
+
+A helm bundle is a [helm chart](https://helm.sh/docs/chart_template_guide/getting_started/#charts).
+The helm chart can be created by [helm create](https://helm.sh/docs/helm/helm_create/).
+
+The current helm bundle format name is the `helm+v3` that
+combines the type of bundle (helm) with the current helm format version (v3).
+
+## Example
+
+For example, below is some examples of a file tree in a "helm+v3" bundle.
+
+```tree
+<bundleRoot>
+└── hello-world
+    ├── .helmignore
+    ├── Chart.yaml
+    ├── README.md
+    ├── templates
+    │   ├── NOTES.txt
+    │   ├── _helpers.tpl
+    │   ├── deployment.yaml
+    │   ├── service.yaml
+    │   └── serviceaccount.yaml
+    └── values.yaml
+```
+
+and
+
+```tree
+<bundleRoot>
+├── .helmignore
+├── Chart.yaml
+├── README.md
+├── templates
+│   ├── NOTES.txt
+│   ├── _helpers.tpl
+│   ├── deployment.yaml
+│   ├── service.yaml
+│   └── serviceaccount.yaml
+└── values.yaml
+```


docs/bundles/helm.md
```diff
@@ -0,0 +1,52 @@
...
+
+Supported source types for a helm bundle currently include the following:
+
+* A container image
+* A directory in a git repository
+* A [http](../sources/http.md)
+* An [upload](../uploading-bundles.md)
+
+Additional source types, such as a local volume are on the roadmap. These source types
+all present the same content, a directory containing a helm chart, in a different ways.
```
joelanford, line 27: I think this is more appropriate for the helm provisioner docs.

The helm+v3 format itself doesn't require a specific set of "transport" formats. If someone else wants to implement a different helm provisioner with a different set of source types, that has no impact on the helm bundle format.

docs/bundles/helm.md
```diff
@@ -0,0 +1,52 @@
...
+content embedded in the bundle. The bundle can be used as the underlying `spec.source` for
+a [Bundle](https://github.com/operator-framework/rukpak#bundle) resource.
+
+A helm bundle is a [helm chart](https://helm.sh/docs/chart_template_guide/getting_started/#charts).
+The helm chart can be created by [helm create](https://helm.sh/docs/helm/helm_create/).
+
+The currently implemented helm bundle format is the `helm+v3` format. The name of the bundle format, `helm+v3`
+combines the type of bundle (helm) with the current helm format version (v3).
+The [helm bundle provisioner](https://github.com/operator-framework/rukpak/blob/main/internal/provisioner/helm/README.md)
+is able to source `helm+v3` bundles and install them onto a Kubernetes cluster.
```
joelanford, line 17: I think conceptually, we shouldn't be talking about what's implemented or not in this doc. In my mind, this doc is simply "here the spec of a helm+v3 bundle; if you're a helm provisioner and you accept a helm+v3 bundle, you must implement unpacking/validation logic according to this spec."

docs/bundles/helm.md
```diff
@@ -0,0 +1,52 @@
...
+│   ├── _helpers.tpl
+│   ├── deployment.yaml
+│   ├── service.yaml
+│   └── serviceaccount.yaml
+└── values.yaml
+```
+
+## Technical Details
+
+* Only one chart should be in the bundle.
```
joelanford, line 49: IMO, this is a slightly strange statement to make and could be confusing (e.g. "can my helm+v3 bundle specify dependencies?). I think we should define the helm+v3 bundle as a single helm chart. So by definition, a helm+v3 bundle _is_ a helm chart.

docs/bundles/helm.md
akihikokuroda: OK.  I'll move to the helm provisioner doc.

docs/bundles/helm.md
akihikokuroda: OK.  I take this out.

docs/bundles/helm.md
```diff
@@ -0,0 +1,36 @@
...
+
+The current helm bundle format name is the `helm+v3` that
+combines the type of bundle (helm) with the current helm format version (v3).
+
+## Example
+
+For example, below is an example of a file tree in a "helm+v3" bundle.
+
+```tree
+hello-world
```
joelanford, line 22: As of https://github.com/operator-framework/rukpak/pull/516, we support helm chart bundles with or without a root chart directory. So at least the helm provisioner supports both of these:
```
<bundleRoot>
└── nginx
    ├── Chart.yaml
    ├── charts
    ├── templates
    │   ├── NOTES.txt
    │   ├── _helpers.tpl
    │   ├── deployment.yaml
    │   ├── hpa.yaml
    │   ├── ingress.yaml
    │   ├── service.yaml
    │   ├── serviceaccount.yaml
    │   └── tests
    │       └── test-connection.yaml
    └── values.yaml
```

and

```
<bundleRoot>
├── Chart.yaml
├── charts
├── templates
│   ├── NOTES.txt
│   ├── _helpers.tpl
│   ├── deployment.yaml
│   ├── hpa.yaml
│   ├── ingress.yaml
│   ├── service.yaml
│   ├── serviceaccount.yaml
│   └── tests
│       └── test-connection.yaml
└── values.yaml
```

IMO, we should specify in the spec that implementations MUST support both of these.

akihikokuroda: The contents of the tgz file generated by the `helm package` command has the first format.  So the current implementation of the helm provisioner is only supporting the first format.  We can create a new issue to support the second tree and update the docs/bundle/helm.md.  Is it OK?

joelanford: I added support for the second in that PR I linked so that `rukpakctl run my-bd ./<chartDir>` would work.

akihikokuroda: I see.  I'll add the second tree.

akihikokuroda:
Put into #552

PR #539, opened 2022-08-30
feat: add PSA compliance

tylerslaton:
## Summary
This PR brings Pod Security Admission (PSA) standards to RukPak in the form of updating the Namespace manifest (in accordance with the [Kubernetes documentation](https://kubernetes.io/docs/tasks/configure-pod-container/enforce-standards-namespace-labels/)) and Deployments. 

> **Note**: The namespace is running in a `baseline` model; However, we eventually want this to be restricted. For the time being I've implemented everything in such a way that `restricted` would be satisfied if we switch over *except* that the unpacker pod is running as root currently. This is because we cannot guarantee, at least at the moment, that the container image that we are going to unpack does not run as root. We'll need to think up a solution for this either here or in the future.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM gcr.io/distroless/static:debug
+FROM gcr.io/distroless/static:nonroot
 WORKDIR /
 
 COPY helm helm
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,5 @@
-FROM gcr.io/distroless/static:nonroot
+FROM gcr.io/distroless/static:debug-nonroot
+
 WORKDIR /
 
 COPY helm helm
--- a/internal/source/image.go
+++ b/internal/source/image.go
@@ -14,6 +14,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
@@ -94,6 +95,8 @@ func (i *Image) ensureUnpackPod(ctx context.Context, bundle *rukpakv1alpha1.Bund
 		pod.Spec.Containers[0].Command = []string{"/bin/unpack", "--bundle-dir", "/"}
 		pod.Spec.Containers[0].VolumeMounts = []corev1.VolumeMount{{Name: "util", MountPath: "/bin"}}
 
+		addSecurityContext(pod)
+
 		if bundle.Spec.Source.Image.ImagePullSecretName != "" {
 			pod.Spec.ImagePullSecrets = []corev1.LocalObjectReference{{Name: bundle.Spec.Source.Image.ImagePullSecretName}}
 		}
@@ -102,18 +105,6 @@ func (i *Image) ensureUnpackPod(ctx context.Context, bundle *rukpakv1alpha1.Bund
 		}
 		return nil
 	})
-}
-
-func pendingImagePodResult(pod *corev1.Pod) *Result {
-	var messages []string
-	for _, cStatus := range append(pod.Status.InitContainerStatuses, pod.Status.ContainerStatuses...) {
-		if waiting := cStatus.State.Waiting; waiting != nil {
-			if waiting.Reason == "ErrImagePull" || waiting.Reason == "ImagePullBackOff" {
-				messages = append(messages, waiting.Message)
-			}
-		}
-	}
-	return &Result{State: StatePending, Message: strings.Join(messages, "; ")}
 }
 
 func (i *Image) failedPodResult(ctx context.Context, pod *corev1.Pod) error {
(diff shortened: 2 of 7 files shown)


joelanford:
We should also document somewhere that this change will require bundle image contents to be readable by the unpack pod, which may have an unpredictable process UID.

In general, we can probably say bundle files and directories should have GID=0 and they should have permissions 0750 and 0640 at a minimum.

tylerslaton:
> We should also document somewhere that this change will require bundle image contents to be readable by the unpack pod, which may have an unpredictable process UID.

@joelanford Is this a potential non-starter for things like the `registry` provisioner which is leveraging image content from pretty establish ecosystems? If I'm understanding correctly, that may require `registry+v1` format Bundles to convert over.

joelanford:
@tylerslaton, maybe that'll dictate that the registry provisioner runs in a separate namespace with less restrictive PSA (e.g. baseline)?

Dockerfile
```diff
@@ -1,4 +1,6 @@
-FROM gcr.io/distroless/static:debug
+FROM gcr.io/distroless/static:debug as builder
+FROM gcr.io/distroless/static:nonroot
```
joelanford, line 2: Perhaps we use `debug-nonroot` tag here?

tylerslaton: Today I learned this was an image, thanks @joelanford  😄

joelanford: Yeah those distroless images are super difficult to discover. It's a lot of trial and error for me.

joelanford:
As a follow-up, can we:
1. Add a knob to the image unpacker that lets a provisioner opt-in to the more restrictive security context stuff
2. Update the plain and helm provisioners to use that knob
3. Update the tests for plain and helm images to pass with the restrictive security contexts.

internal/source/image.go
```diff
@@ -190,3 +181,48 @@ func (i *Image) handleUnexpectedPod(ctx context.Context, pod *corev1.Pod) error
...
+	return &Result{State: StatePending, Message: strings.Join(messages, "; ")}
+}
+
+// addSecurityContext is responsible for taking a container and defining the
+// relevant security context values. By having a function do this, we can keep
+// that configuration easily consistent and maintainable.
+func addSecurityContext(pod *corev1.Pod) {
+	// Add security context for overall pod
+	pod.Spec.SecurityContext = &corev1.PodSecurityContext{
+		RunAsNonRoot: pointer.Bool(false),
```
joelanford, line 203: Add a TODO comment here to remind us to think about how to make this work under restricted?

joelanford (approved):
Just a minor nit about adding a TODO comment. Otherwise LGTM.

internal/source/image.go
tylerslaton: Good call out, updated with a detail description of what we should be thinking about here.

internal/source/image.go
```diff
@@ -190,3 +181,60 @@ func (i *Image) handleUnexpectedPod(ctx context.Context, pod *corev1.Pod) error
...
+			}
+		}
+	}
+	return &Result{State: StatePending, Message: strings.Join(messages, "; ")}
+}
+
+// addSecurityContext is responsible for taking a container and defining the
+// relevant security context values. By having a function do this, we can keep
+// that configuration easily consistent and maintainable.
+func addSecurityContext(pod *corev1.Pod) {
```
timflannagan, line 200: Do we need to nil check pod or pod.Spec here?

timflannagan (approved):

internal/source/image.go
tylerslaton: I don't think it would hurt, there could be a situation where this function gets broken out into another package and is passed a nil pod. Added in the latest commit.

internal/source/image.go
```diff
@@ -190,3 +181,65 @@ func (i *Image) handleUnexpectedPod(ctx context.Context, pod *corev1.Pod) error
...
+		// TODO (tyslaton): Address unpacker pod allowing root users for image sources
+		//
+		// In our current implementation, we are creating a pod that uses the image
+		// provided by an image source. This pod is not always guaranteed to run as a
+		// non-root user and thus will fail to initialize if running as root in a PSA
+		// restricted namespace due to violations. As it currently stands, our compliance
+		// with PSA is baseline which allows for pods to run as root users. However,
+		// all RukPak processes and resources, except this unpacker pod for image sources,
+		// are runnable in a PSA restricted environment. We should consider ways to make
+		// this PSA definition either configurable or workable in a restricted namespace.
```
anik120, line 217: "As it currently stands, our compliance with PSA is baseline which allows for pods to run as root users" 

So I'm a little confused about this part.

"we are creating a pod that uses the image provided by an image source": what's an example of this kind of an image? 

My confusing is surrounding the fact that we're labeling the namespace as `enforce:baseline`, and setting `runAsNonRoot: false` (which means we'll allow containers to be run as root), but then setting `allowPrivilegeEscallation:false`. Did we use an example image that fits the criteria to test out this configuration works out for those images? 

Also, since we're deploying rukpak downstream in the platform-operator namespace, I'm guessing this means we want that namespace to be labeled as `baseline` instead of restricted [here](https://github.com/openshift/platform-operators/pull/14)?

anik120 (commented):
@tylerslaton I have a couple of clarifying questions otherwise looks good for the most part 🚀

internal/source/image.go
anik120: > As a result, I turned that feature off but left the other configuration you defined in the downstream PR outside of that. Do you have a recommendation on what we'd change here?

My guess is that `allowPrivilegeEscalation` has to be set to `true` here, and I was hoping there's an image that needs to be run as root that we can use to test out if `allowPrivilegeEscalation:false` is enough for those images.

tylerslaton: So this image [here](https://github.com/operator-framework/rukpak/tree/main/internal/provisioner/plain#:~:text=Create%20the%20combo%20BundleDeployment%20referencing%20the%20desired%20combo%20Bundle%20configuration%3A) runs as root, was failing with `runAsNonRoot` set to true but succeeded when I turned that to false and kept `allowPrivilegeEscalation` at false.

anik120: Okay that answers my question then 

/lgtm 🚀

joelanford: Docs about `allowPrivilegeEscalation` are here: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/

> allowPrivilegeEscalation: Controls whether a process can gain more privileges than its parent process. This bool directly controls whether the [no_new_privs](https://www.kernel.org/doc/Documentation/prctl/no_new_privs.txt) flag gets set on the container process. allowPrivilegeEscalation is always true when the container:
>  -  is run as privileged, or
>  -  has CAP_SYS_ADMIN

PR #541, opened 2022-08-30
fix second helm bundle format handle error

### LINKED ISSUE #540: Helm bundle fails without chart directory with git source
Creating the helm bundle without the chart directory fails with the following error 
```
1.6618967089468236e+09	ERROR	Reconciler error	{"controller": "bundle", "controllerGroup": "core.rukpak.io", "controllerKind": "Bundle", "bundle": {"name":"my-ahoy-dbf4zx"}, "namespace": "", "name": "my-ahoy-dbf4zx", "reconcileID": "2546d990-9bde-4139-b3d8-e9b540fcb10e", "error": "generate tar.gz from FS: cannot open directory: /charts/hello-world"}
```

Here is the BD for this.
```
kind: BundleDeployment
metadata:
  name: my-ahoy
spec:
  provisionerClassName: core-rukpak-io-helm
  template:
    metadata:
      labels:
        app: my-ahoy
    spec:
      provisionerClassName: core-rukpak-io-helm
      source:
        git:
          ref:
            branch: main
          repository: https://github.com/akihikokuroda/examples
          directory: ./charts/hello-world
        type: git
```

akihikokuroda:
Closes #540 

There are some change and addition to the custom FSes.

Adding DirRead to BaseDirFS
Add Close for billyFS
Change not to fail Open for a  directory for billyFS

### DIFF
--- a/internal/source/git.go
+++ b/internal/source/git.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"syscall"
 
 	"github.com/go-git/go-billy/v5"
 	"github.com/go-git/go-billy/v5/memfs"
@@ -209,12 +210,18 @@ func (f *billyFS) ReadFile(name string) ([]byte, error) {
 }
 
 func (f *billyFS) Open(path string) (fs.File, error) {
+	fi, err := f.Filesystem.Stat(path)
+	if err != nil {
+		return nil, err
+	}
+	if fi.IsDir() {
+		return &billyDirFile{billyFile{nil, fi}, f, path}, nil
+	}
 	file, err := f.Filesystem.Open(path)
 	if err != nil {
 		return nil, err
 	}
-	fi, err := f.Filesystem.Stat(path)
-	return &billyFile{file, fi, err}, nil
+	return &billyFile{file, fi}, nil
 }
 
 func (f *billyFS) ReadDir(name string) ([]fs.DirEntry, error) {
@@ -231,10 +238,34 @@ func (f *billyFS) ReadDir(name string) ([]fs.DirEntry, error) {
 
 type billyFile struct {
 	billy.File
-	fi    os.FileInfo
-	fiErr error
+	fi os.FileInfo
 }
 
 func (b billyFile) Stat() (fs.FileInfo, error) {
-	return b.fi, b.fiErr
+	return b.fi, nil
+}
+
+func (b billyFile) Close() error {
+	if b.File == nil {
+		return nil
+	}
+	return b.File.Close()
+}
+
+type billyDirFile struct {
+	billyFile
+	fs   *billyFS
+	path string
+}
+
+func (d *billyDirFile) ReadDir(n int) ([]fs.DirEntry, error) {
+	entries, err := d.fs.ReadDir(d.path)
+	if n <= 0 || n > len(entries) {
+		n = len(entries)
+	}
+	return entries[:n], err
+}
+
+func (d billyDirFile) Read(data []byte) (int, error) {
+	return 0, &fs.PathError{Op: "read", Path: d.path, Err: syscall.EISDIR}
 }
--- a/test/e2e/helm_provisioner_test.go
+++ b/test/e2e/helm_provisioner_test.go
@@ -432,6 +432,69 @@ var _ = Describe("helm provisioner bundledeployment", func() {
 			})
 		})
 	})
+	When("a BundleDeployment targets a valid Bundle with no chart directory in Github", func() {
+		var (
+			bd  *rukpakv1alpha1.BundleDeployment
+			ctx context.Context
+		)
+		BeforeEach(func() {
+			ctx = context.Background()
+
+			bd = &rukpakv1alpha1.BundleDeployment{
+				ObjectMeta: metav1.ObjectMeta{
+					GenerateName: "ahoy-",
+				},
+				Spec: rukpakv1alpha1.BundleDeploymentSpec{
+					ProvisionerClassName: helm.ProvisionerID,
+					Template: &rukpakv1alpha1.BundleTemplate{
+						ObjectMeta: metav1.ObjectMeta{
+							Labels: map[string]string{
+								"app.kubernetes.io/name": "ahoy",
+							},
+						},
+						Spec: rukpakv1alpha1.BundleSpec{
+							ProvisionerClassName: helm.ProvisionerID,
+							Source: rukpakv1alpha1.BundleSource{
+								Type: rukpakv1alpha1.SourceTypeGit,
+								Git: &rukpakv1alpha1.GitSource{
+									Repository: "https://github.com/helm/examples",
+									Directory:  "./charts/hello-world",
+									Ref: rukpakv1alpha1.GitRef{
+										Branch: "main",
+									},
+								},
+							},
+						},
+					},
+				},
+			}
+			err := c.Create(ctx, bd)
+			Expect(err).To(BeNil())
+		})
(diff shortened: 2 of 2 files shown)


internal/util/fs.go
```diff
@@ -88,3 +88,21 @@ func (f baseDirFS) Open(name string) (fs.File, error) {
 	}
 	return nil, fs.ErrNotExist
 }
+
+func (f *baseDirFS) ReadDir(name string) ([]fs.DirEntry, error) {
```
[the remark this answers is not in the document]
akihikokuroda: I see.  The error of `Open` is from [here](https://github.com/go-git/go-billy/blob/7ab80d7c013de28ffbb1ca64b9bbf8dd1cbd81c5/memfs/memory.go#L64). I don't see an issue opened related to this there.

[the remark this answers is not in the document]
joelanford: BillyFS was implemented prior to `fs.FS` existing. I think the ability to `Open(dirPath)` is something `fs.FS` expects to work, but not necessarily other fs-style implementations that pre-date `fs.FS`.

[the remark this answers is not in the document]
akihikokuroda: I see.

internal/source/git.go
```diff
@@ -238,3 +238,7 @@ type billyFile struct {
 func (b billyFile) Stat() (fs.FileInfo, error) {
 	return b.fi, b.fiErr
 }
+
+func (b billyFile) Close() error {
+	return nil
+}
```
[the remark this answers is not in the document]
akihikokuroda: I'll take it.  Thanks!

internal/source/git.go
```diff
@@ -211,7 +211,7 @@ func (f *billyFS) ReadFile(name string) ([]byte, error) {
 func (f *billyFS) Open(path string) (fs.File, error) {
 	file, err := f.Filesystem.Open(path)
 	if err != nil {
-		return nil, err
+		file = nil
 	}
 	fi, err := f.Filesystem.Stat(path)
```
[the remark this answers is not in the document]
akihikokuroda: That's better.  I'll make changes.

internal/source/git.go
```diff
@@ -209,11 +209,17 @@ func (f *billyFS) ReadFile(name string) ([]byte, error) {
 }
 
 func (f *billyFS) Open(path string) (fs.File, error) {
+	fi, err := f.Filesystem.Stat(path)
+	if err != nil {
+		return nil, err
+	}
+	if fi.IsDir() {
+		return &billyFile{nil, fi, err}, nil
+	}
```
joelanford, line 218: Not sure if you're still working, but I think we should:
- drop the `fiErr` field from the `billyFile` struct. It's unnecessary now because you're `err` ends up always being `nil` when a `billyFile` is returned.
- do something like `return &billyDir{billyFile{nil, fi}}` such that we return a `fs.File` from `Open` that also implements `fs.FileReadDir`, which seems to be the expectation of `fs.Open`.

internal/util/fs.go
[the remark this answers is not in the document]
joelanford: Yeah, so I'd suggest going ahead and removing this to make sure the fixes in billyFS are sufficient.

internal/source/git.go
akihikokuroda: OK.  I'll try.

akihikokuroda: I couldn't make it work without the `ReadDir`.   I might not do correctly.  Can I make a follow up issue and try later?

akihikokuroda: I' doing one more try.

akihikokuroda: It seems working now.  Thanks!

internal/source/git.go
```diff
@@ -209,12 +209,18 @@ func (f *billyFS) ReadFile(name string) ([]byte, error) {
 }
 
 func (f *billyFS) Open(path string) (fs.File, error) {
+	fi, err := f.Filesystem.Stat(path)
+	if err != nil {
+		return nil, err
+	}
+	if fi.IsDir() {
+		return &billyDirFile{billyFile{nil, nil}, f, path}, nil
```
joelanford, line 217:
```suggestion
		return &billyDirFile{billyFile{nil, fi}, f, path}, nil
```

internal/source/git.go
```diff
@@ -231,10 +237,26 @@ func (f *billyFS) ReadDir(name string) ([]fs.DirEntry, error) {
...
+}
+
+type billyDirFile struct {
+	billyFile
+	fs   *billyFS
+	path string
+}
+
+func (d *billyDirFile) ReadDir(n int) ([]fs.DirEntry, error) {
+	return d.fs.ReadDir(d.path)
```
joelanford, line 261: The [docs](https://pkg.go.dev/io/fs#ReadDirFile) for implementing this function are pretty confusing to me, specifically:
> 	// In [the case of n > 0], if ReadDir returns an empty slice, it will return
	// a non-nil error explaining why.
	// At the end of a directory, the error is io.EOF.
	// (ReadDir must return io.EOF itself, not an error wrapping io.EOF.)

I assume we can rely on `b.fs.ReadDir` to handle those specifics, but I'd suggest something like this, so we at least honor the `n` parameter and follow the other piece of the ReadDirFile docs that talk about returning a partial list in the event of an error.
```go
	entries, err := b.fs.ReadDir(b.path)
	if n <= 0 {
		return entries, err
	}
	if n > len(entries) {
		n = len(entries)
	}
	return entries[:n], err
```

internal/source/git.go
```diff
@@ -231,10 +237,26 @@ func (f *billyFS) ReadDir(name string) ([]fs.DirEntry, error) {
...
-	return b.fi, b.fiErr
+	return b.fi, nil
+}
+
+func (b billyFile) Close() error {
+	if b.File != nil {
+		return b.File.Close()
+	}
+	return nil
+}
```
joelanford, line 252: Should we also implement a similar custom `Read` function on `billyFile` to avoid a panic when `b.File` is `nil` (in the case of a billyDirFile)?

akihikokuroda:
I addressed all comments.

internal/source/git.go
```diff
@@ -231,10 +237,37 @@ func (f *billyFS) ReadDir(name string) ([]fs.DirEntry, error) {
...
+		return nil
+	}
+	return b.File.Close()
+}
+
+func (b billyFile) Read(data []byte) (int, error) {
+	if b.File != nil {
+		return b.File.Read(data)
+	}
+	return 0, fs.ErrNotExist
```
joelanford, line 258: Sorry for all the back and forth on this. I'm learning this at the same time you are...

`fs.ErrNotExist` would be returned by `Stat()`. So by the time we have a `billyFile` we know _something_ exists at the opened path.

[Seems like](https://go.dev/play/p/ycgOLd_n7dQ) reading an OS directory results in:
```go
return 0, &PathError{Op:"read", Path: path, Err: syscall.EISDIR}
```

And this makes me think that we need to check `billyFile.fi.IsDir()` during `Read`, and if we're doing _that_, then maybe what we want is just a custom `Read()` on `billyDirFile`:

```go
func (b billyDirFile) Read(data []byte) (int, error) {
	return 0, &PathError{Op:"read", Path: path, Err: syscall.EISDIR}
}
```

and then get rid of the `billyFile` `Read()` altogether (so that we just use the underlying `billy.File.Read()` as we were before)

akihikokuroda: OK.  Updated.

joelanford (approved):
Awesome! Thanks for bearing with me. Looks great!

PR #544, opened 2022-08-31
add docs/sources/image.md

akihikokuroda:
Part of #522 

### DIFF
--- a/docs/sources/image.md
+++ b/docs/sources/image.md
@@ -1,0 +1,104 @@
+# Image source
+
+## Summary
+
+The image source provides the contents in a container image as the source of the bundle.  The `source.type` for the image source is `image`.
+When creating a image source, a reference to an URL of the image is specified. It is expected that a proper format of bundle contents is present
+in the image.
+
+## Example
+
+For example, below is a minimal example of a Dockerfile that builds a bundle for image source from a directory
+containing bundle contents.
+
+```dockerfile
+FROM scratch
+COPY "Bundle contents directory" "File location for the bundle format" 
+```
+
+Build the image using a container tool like docker or podman. Use an image tag that references a repository that you
+have push access to. For example,
+
+```bash
+docker build -f Dockerfile.example -t quay.io/operator-framework/rukpak:example .
+```
+
+Push the image to the remote registry
+
+```bash
+docker push quay.io/operator-framework/rukpak:example
+```
+
+The bundle source is referred in the following Bundle Deployment example.  
+```yaml
+apiVersion: core.rukpak.io/v1alpha1
+kind: BundleDeployment
+metadata:
+  name: my-bundle
+spec:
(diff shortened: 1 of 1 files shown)


docs/sources/image.md
```diff
@@ -0,0 +1,97 @@
...
+```
+
+#### Method 2: Add the secret to the `imagePullSecrets` in the `default` service account in the provisioner deployed namespace
+
+```bash
+kubectl patch serviceaccount default -p '{"imagePullSecrets": [{"name": "mysecret"}]}' -n rukpak-system
+```
+* This command replaces the secrets already in the `imagePullSecrets`.  To add the secret to the existing secrets, add the secret in the imagePullSecrets array of the existing secrets like `imagePullSecrets": [{"name": "mysecret"}, {"name": "existing_secret1"}, {"name": "existing_secret2"}]`
+
+## Technical Details
```
joelanford, line 95: It seems appropriate in this section to talk about the mechanism by which the image source actually unpacks bundle content.

In particular:
- It launches a pod to ensure that the bundle image is pulled using the CRI, which is beneficial because it means bundle images are pulled with the exact same set of credentials, proxies, and other configurations that would be accessible to kubelet for any other workload images.
- The pod must be schedulable in the namespace in which the provisioner using the image source is running. There are implications with [PSA](https://kubernetes.io/docs/concepts/security/pod-security-admission/) which can cause bundle images to fail to unpack. To avoid unpack failures and ensure widest compatibility with various provisioners, bundle image authors should ensure that bundle images can be scheduled in a namespace with the `restricted` mode enforced. Bundle directory hierarchies in images should be traversable/readable by arbitrary users.

docs/sources/image.md
```diff
@@ -0,0 +1,97 @@
+# Image source
+
+## Summary
+
+The image source provides the contents in a container image as the source of the bundle.  The `source.type` for the image source is `image`.
+When creating a image source, a reference to an URL of the image is specified. It is expected that a proper format of bundle contents is present
+in the image.
```
joelanford, line 7: Are these docs meant to be descriptions of our implementations of the various source types, or are they meant to be specs for _how_ source implementations must work? It's probably important to distinguish in the docs so that readers can understand.

Generally, I'd expect most provisioners to vendor and use the source implementations from this repo, but technically a provisioner has some freedom in its implementation of the source API (barring any conformance tests we might come up with).

docs/sources/image.md
akihikokuroda: I see.  I'll put this.  Thanks!

docs/sources/image.md
akihikokuroda: I see.  That's a good point.  I was focusing on the former.  Do we create some docs for provisioner providers?

akihikokuroda:
Put into #552

PR #555, opened 2022-09-06
image source: use job for bundle unpacking

joelanford:
It turns out that directly managing pods for image unpacking is more
susceptible to runtime mutation issues (since pods are more likely to
have mutation admission webhooks associated with them than jobs).
Mutating webhooks are problematic for our controller because we need to
reliably detect when a diff between the desired pod and the existing
on-cluster pod is a result of a change triggered via our controller vs.
a change due to some unrelated mutating webhook.

The problem still exists with a Job if the cluster has a mutating
webhook for jobs, but we'll assume for now that the problem is much less
pronounced with jobs than with pods.

NOTE: the pod specs in both jobs and pods are immutable. Therefore, our
controller cannot simply use server-side apply to patch objects back to
desired state. If we could, this problem could be avoided entirely
because we'd never need to detect diffs (we'd just send our SSA patch to
the apiserver and have it do the work).

### DIFF
--- a/internal/provisioner/plain/controllers/bundle_controller.go
+++ b/internal/provisioner/plain/controllers/bundle_controller.go
@@ -268,9 +268,10 @@ func (r *BundleReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		For(&rukpakv1alpha1.Bundle{}, builder.WithPredicates(
 			util.BundleProvisionerFilter(plain.ProvisionerID),
 		)).
-		// The default unpacker creates Pod's ownerref'd to its bundle, so
-		// we need to watch pods to ensure we reconcile events coming from these
-		// pods.
-		Watches(&crsource.Kind{Type: &corev1.Pod{}}, util.MapOwneeToOwnerProvisionerHandler(context.Background(), mgr.GetClient(), l, plain.ProvisionerID, &rukpakv1alpha1.Bundle{})).
-		Complete(r)
+		// The default image source unpacker creates a Job, which creates a pod,
+		// resulting in an ownerRef chain from Bundle to Pod. We'll watch pods
+		// directly because we care about events at the granularity of the pod,
+		// and we'll map those pod events back up the ownerRef chain to the root
+		// bundle.
+		Watches(&crsource.Kind{Type: &corev1.Pod{}}, util.MapUnpackPodToBundleHandler(context.Background(), mgr.GetClient(), l, plain.ProvisionerID)).Complete(r)
 }
--- a/internal/source/image.go
+++ b/internal/source/image.go
@@ -5,12 +5,14 @@ import (
 	"compress/gzip"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"io/fs"
 	"strings"
 
 	"github.com/nlepage/go-tarfs"
+	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
@@ -25,7 +27,7 @@ import (
 type Image struct {
 	Client       client.Client
 	KubeClient   kubernetes.Interface
-	PodNamespace string
+	JobNamespace string
 	UnpackImage  string
 }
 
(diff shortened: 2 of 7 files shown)


akihikokuroda:
Is there any way to create the image bundle without creating the pod with the image?

exdx:
> Is there any way to create the image bundle without creating the pod with the image?

The current behavior is largely a hold-over from the way OLM does bundle unpacking. Essentially, we just need a way to copy the contents of the container filesystem (really just the directories we care about) to disk. Instead of creating the pod and running the unpacker, we could use a tool like dcp (or similar process) to just copy what we need out.

joelanford:
The reason OLM (and rukpak) use a pod is because it results in kubelet and the CRI being the actual components that pull the image. This is important because gives users a consistent experience in terms of "what and how are images pulled by the cluster?" It ensures that workload images and bundle images are pulled using the exact same mechanisms.

We _could_ potentially introduce a fully in-process image puller, but I don't think we could use that as a full substitute for the pod/job-based bundle image unpacker.

internal/source/image.go
```diff
@@ -197,49 +254,57 @@ func pendingImagePodResult(pod *corev1.Pod) *Result {
 // addSecurityContext is responsible for taking a container and defining the
 // relevant security context values. By having a function do this, we can keep
 // that configuration easily consistent and maintainable.
-func addSecurityContext(pod *corev1.Pod) {
+func addSecurityContext(podSpec *corev1.PodSpec) {
 	// Check that pod is defined before proceeding
```
tylerslaton, line 258: nit:
```suggestion
	// Check that podSpec is defined before proceeding
```

internal/source/image.go
```diff
@@ -164,22 +190,53 @@ func (i *Image) getBundleImageDigest(pod *corev1.Pod) (string, error) {
...
-	defer logReader.Close()
-	buf := &bytes.Buffer{}
-	if _, err := io.Copy(buf, logReader); err != nil {
+	if err := i.Client.List(ctx, &podList, client.InNamespace(job.Namespace), client.MatchingLabelsSelector{Selector: ls}); err != nil {
 		return nil, err
 	}
-	return buf.Bytes(), nil
+	if len(podList.Items) == 0 {
+		return nil, nil
+	}
```
tylerslaton, line 204: Could you help me understand why no pods found for a specific job is not an error?

tylerslaton (commented):
These changes look right. I had a few small comments. Other than that, it looks like the E2E is failing because we are not updating the tests to no longer reference pods instead of jobs. 

For example:

https://github.com/operator-framework/rukpak/blob/e15bf28c047d3c0118d5da70cd6facc34c2ded99/test/e2e/plain_provisioner_test.go#L152-L186

internal/provisioner/plain/controllers/bundle_controller.go
```diff
@@ -268,9 +268,9 @@ func (r *BundleReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 			util.BundleProvisionerFilter(plain.ProvisionerID),
 		)).
-		// The default unpacker creates Pod's ownerref'd to its bundle, so
-		// we need to watch pods to ensure we reconcile events coming from these
-		// pods.
-		Watches(&crsource.Kind{Type: &corev1.Pod{}}, util.MapOwneeToOwnerProvisionerHandler(context.Background(), mgr.GetClient(), l, plain.ProvisionerID, &rukpakv1alpha1.Bundle{})).
+		// The default image source unpacker creates Jobs ownerref'd to its bundle, so
+		// we need to watch jobs to ensure we reconcile events coming from these
+		// jobs.
+		Watches(&crsource.Kind{Type: &batchv1.Job{}}, util.MapOwneeToOwnerProvisionerHandler(context.Background(), mgr.GetClient(), l, plain.ProvisionerID, &rukpakv1alpha1.Bundle{})).
```
akihikokuroda, line 274: Is it enough just to watch the Job?  Do we care about the Pod deleted manually?

joelanford: Yep, good catch. I had also found another corner case where watching just jobs is not enough granularity for bubbling up specific events from pods (e.g. the e2e is failing the image pull backoff test)

internal/source/image.go
joelanford: I need to double check if we need to also handle the case where the job completed, the pod completed, and then the pod was deleted without the job being deleted.

joelanford: And... we do. Good catch!

akihikokuroda (approved):

joelanford:
/hold

I'm also going to try out a hashing technique, where we add an annotation to the unpack pod that contains the desired pod hash, and then base our diff logic on a hash change rather than a DeepEqual comparison.

joelanford:
Closing in favor of #559

PR #556, opened 2022-09-06
ci: exempt issues placed in the icebox milestone

tylerslaton:
Adding a small change that makes it so issues in the `icebox` milestone no longer get updated by the `stale` action. The main thought here is that, since these issues are placed in the `icebox`, they shouldn't be life-cycled as normal.

Alternatively, we could close this issue and the `icebox` milestone and apply all current `icebox` issues with the `lifecycle/frozen` label; this would have the same effect as this PR.

### DIFF
--- a/.github/workflows/stale.yaml
+++ b/.github/workflows/stale.yaml
@@ -27,5 +27,6 @@ jobs:
           days-before-issue-close: -1
           stale-issue-label: lifecycle/stale
           exempt-issue-labels: lifecycle/frozen
+          exempt-issue-milestones: icebox
       - name: Print outputs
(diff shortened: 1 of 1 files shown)


joelanford:
> Alternatively, we could close this issue and the icebox milestone and apply all current icebox issues with the lifecycle/frozen label. This would have the same affect as this PR.

This would be my vote. I'd rather have exactly one mechanism to "freeze" issues.

timflannagan:
That icebox milestone was a short term hack. I'm +1 on removing it as well.

tylerslaton:
Nice, I'm in agreement with that. I'll close this PR, the icebox milestone, and start moving those issues over to `lifecycle/frozen`.

PR #559, opened 2022-09-07
image source: use applyconfiguration with server-side apply

### LINKED ISSUE #560: Unpacker pod is constantly being re-created
Our diff logic that decides whether or not to replace an unpack pod is not working correctly. When rukpak is deployed in a cluster with a pod mutation webhook that affects the unpack pod, we currently see those unrelated pod mutations as diffs and decide we need to re-create the pod, which results in those same mutations, which results in our re-spinning the pod... and the cycle continues forever.

joelanford:
This is another take on #555 that makes use of applyconfigurations and server-side apply to "apply or re-create" unpack pods.

This should solve problems that we're seeing with our existing diff comparisons where we do not handle mutations performed by unrelated webhooks.

This should be more resilient than a "hash the pod struct" approach, which is susceptible to hash changes when the Pod struct changes. This approach should theoretically not result in an unpack pod respin unless something in our apply configuration actually changed.

I tested this on a cluster that performs pod mutations and verified that Bundles are able to be successfully unpacked, and the controller does not continuously re-create the unpack pod.

Closes #560 

### DIFF
--- a/internal/source/image.go
+++ b/internal/source/image.go
@@ -12,9 +12,12 @@ import (
 
 	"github.com/nlepage/go-tarfs"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	applyconfigurationcorev1 "k8s.io/client-go/applyconfigurations/core/v1"
+	v1 "k8s.io/client-go/applyconfigurations/meta/v1"
 	"k8s.io/client-go/kubernetes"
-	"k8s.io/utils/pointer"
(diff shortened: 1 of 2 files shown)


internal/source/image.go
```diff
@@ -41,6 +44,7 @@ func (i *Image) Unpack(ctx context.Context, bundle *rukpakv1alpha1.Bundle) (*Res
 
 	pod := &corev1.Pod{}
 	op, err := i.ensureUnpackPod(ctx, bundle, pod)
+	fmt.Println(op, err)
```
timflannagan, line 47: nit: debug artifact?

timflannagan (approved):
It would be nice to have some sort of tests to verify we're not regressing but I tested these changes out on a cluster that mutates the Pod, which passes the e2e suite, and I didn't see any immediate issues.

Issue #546, opened 2022-09-01
Introduce a bot to automate mark issues/PRs


awgreene:
Created on behalf of @tylerslaton

It would be great to introduce a `needs-triage` label that could be used to drive conversation in the Issue Triage meeting.

It would be great to introduce a bot that could apply a `needs-triage` label to issues when:
- A new issue has been submitted
- A PR has not been updated for X days 
- An Issue has not been updated for Y days

I doubt that we need a custom solution, there are many bots that might be able to fulfill this need, such as the [k8s-triage-robot](https://github.com/kubernetes-sigs/kustomize/pull/4336#issuecomment-1175820894).

Notes:
- We must discuss the number of days until an Issue/PR becomes "stale"
- The bot should convey enough information to identify why the label was applied (new issue vs stale issue)

joelanford:
perhaps this is an option? https://github.com/marketplace/stale

tylerslaton:
A bot has been created and will triage our issues over time as stale. There is a hard limit on the Github API that it uses as it has a rate limit. Accordingly, it can only do some many issues at a time which feels a bit lacking as large parts of our issue backlog remains incorrectly not stale. This is only because the action is unable to reach them due to rate limiting. I'm going to take a look into how we can get around that issue through various ways but the immediate term may result in just lowering the amount of issues we have open at any given time.

Also, I don't think we need a bot to add labels for new issues. Instead, we can just query issues in Github that have no milestone for new issues:

https://github.com/issues?q=is%3Aopen+no%3Amilestone+repo%3Aoperator-framework%2Frukpak+is%3Aissue

tylerslaton:
Alright, with the most recent PR related to this issue #563, we should be good on this front. Issues are automatically marked as `lifecycle/stale` after 60 days. PR's are marked as `lifecycle/stale` after 30 and get closed after another 30 with the `lifecycle/rotten` label. Both issues and PR's can ignore life-cycling events by adding the `lifecycle/frozen` label.

For new issues, we can use the query I linked above.

PR #563, opened 2022-09-12
fix: update stale action to have more capacity to process items

tylerslaton:
## Summary
Right now the bot is having a hard time processing our backlog due to its size. By default, there are 30 operations allowed before it stops attempting to process things to avoid rate limiting. We can get around this by increasing the amount of operations allowed per run. This has more potential for rate limiting, but should be fine in off hours.

In addition to adding these changes, this commit sets the the action only run once per day and updates the messaging it leaves when marking issues/prs as stale.

## Notes
- Should we consider moving to another bot to perform this work? It seems there are some natural limitations around this.
- The more we triage our issues, the less operations that will need to be performed.

Relates to #546

### DIFF
--- a/.github/workflows/stale.yaml
+++ b/.github/workflows/stale.yaml
@@ -3,7 +3,7 @@ on:
   workflow_dispatch:
   schedule:
     # The start of everyday and at hour 12
-    - cron: '00 00,12 * * *'
+    - cron: '00 00 * * *'
 
 jobs:
   stale:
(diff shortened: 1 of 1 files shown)


joelanford:
I'm curious if we're catching up slowly or if the current configuration leaves us in a place where we'll be unlikely to have everything labelled correctly?

Can we manually trigger the action a few times a day to speed up the process of getting through the backlog such that this change is unnecessary? Related, do we plan to revert this PR once the bot has made its way through the backlog?

tylerslaton:
So out of the last 5 runs no issues we would expect to be marked as stale get marked as such. This is because the action fairly consistently batches the same issues together and runs them in order. When it detects an issue marked as stale it uses seems to use 2 operations to determine if an update occurred and if it can thus remove the stale label (this feels odd and I'm trying to figure out a way around it). As a result it uses up the operations quickly and it exits before it gets to check any new issues/pr for staleness. I've seen some use of this PR set [ignore-updates](https://github.com/actions/stale#ignore-updates) for larger communities but that feels somewhat bad for the community long term.

> Can we manually trigger the action a few times a day to speed up the process of getting through the backlog such that this change is unnecessary?

Based on my understanding, no. This is because it always processes the same batch with the same items in a fairly predictable way. If no updates happen to the backlog then it will produce the same output.

>  Related, do we plan to revert this PR once the bot has made its way through the backlog?

I would be fine with this, but we need to be pretty diligent in our backlog health. I think we can keep the operations at this level and if we continue to run into problems we can set the `ignore-updates` option.

joelanford (approved):
Yep, I ran a few runs manually and saw the same behavior. So I'm +1 on bumping our operation count up. Once the stale issues start to close, that should help keep things in check.

tylerslaton:
Nice, I'm going to go ahead and merge so we can see how much progress we make after tonight's run.

tylerslaton:
@joelanford Looks like we got through our whole backlog with 68 operations so we should be good as long as we start triaging our issues to avoid becoming stale moving forward.

PR #562, opened 2022-09-12
fix: avoid bundle update errors during deletion

joelanford:
Our finalizer code is not integrated with our updater code, which results in two multiple update calls being made during a single reconcile when a bundle is deleted. The first call removes the finalizer, the second attempts to update the status. The second call fails because the bundle has been deleted in etcd after the finalizer is removed with the first call.

This commit fully rips out the bundle updater and refactors things to ensure there’s never more than one status and bundle update per reconcile.

Closes #566

PR #564, opened 2022-09-12
fix: align bundle deployment update mechanism with bundles

joelanford:
This commit fully rips out the bundle deployment updater and refactors things to ensure there’s never more than one status and bundle deployment update per reconcile.

Related to #562


akihikokuroda (approved):
LGTM

timflannagan (approved):
LGTM once the sanity issues are fixed.

PR #569, opened 2022-09-16
fix: update docs directory to be more condensed and update information slightly

tylerslaton:
Updating the docs directory to be more condensed, remove old documents, fix various erroneous information, and renamed a few files to be clearer.


timflannagan (approved):
This is fine for now imo. The alternative is we have a top-level "concepts" folder that could house the bundle immutability and provisioner spec documentation.

tylerslaton:
> This is fine for now imo. The alternative is we have a top-level "concepts" folder that could house the bundle immutability and provisioner spec documentation.

Oh nice, I like that better. I'll go ahead and add that and then merge.

PR #570, opened 2022-09-16
define constants for common defaults

joelanford:
Also, start using build tags to make it easier to change defaults for downstream builds.

### DIFF
--- a/internal/util/defaults_upstream.go
+++ b/internal/util/defaults_upstream.go
@@ -1,0 +1,9 @@
+//go:build upstream
+
+package util
+
+const (
+	DefaultSystemNamespace   = "rukpak-system"
+	DefaultUnpackImage       = "quay.io/operator-framework/rukpak:latest"
+	DefaultUploadServiceName = "core"
+)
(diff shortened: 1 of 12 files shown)


internal/util/defaults_upstream.go
```diff
@@ -0,0 +1,9 @@
+//go:build upstream
+
+package util
+
+const (
```
tylerslaton, line 5: Is the idea that downstream would have its own defaults?

tylerslaton (approved):
Nice! This will help a lot for downstream readability.

internal/util/defaults_upstream.go
joelanford: Yep! downstream, we can add a `defaults_downstream.go` file (which should never conflict because it has `_downstream` in the name), and then override `GO_BUILD_TAGS` to `downstream` when we build.

joelanford: This build tag mechanism is also the way we can manage differences in code/functionality between upstream and downstream as well.

exdx (approved):

timflannagan (approved):

PR #576, opened 2022-09-22
Consolidate bundle and BD controller code

### LINKED ISSUE #575: DRY up copy/pasted provisioner reconciler code
We currently have a large amount of copy/pasted code in our various in-tree provisioners. We have three bundle provisioners: plain, registry, and helm. And we have two BD provisioners: plain and helm.

A very large majority of the code in each of these provisioners is identical, and the only differences are in:
- how the bundle content is validated and converted before storing
- how the bundle content is validated and rendered into a helm chart before being applied to the cluster.

We should look into introducing a few interfaces to capture these implementation-specific differences along with generic bundle and BD reconcilers that use these interfaces such that the copy/pasted code can be shared and DRY-ed up.

joelanford:
Closes #575


joelanford:
New libraries are:
- `internal/provisioner/bundle`
- `internal/provisioner/bundledeployment`

Interface implementations are still under `internal/provisioner/<format>`, but code in those directories is now limited to _just_ the interface implementations.

Main commands were updated to build the reconcilers with the new libraries and format-specific implementations.

cmd/core/main.go
```diff
@@ -205,37 +207,44 @@ func main() {
...
-		Scheme:     mgr.GetScheme(),
-		Storage:    bundleStorage,
-		Finalizers: bundleFinalizers,
-		Unpacker:   unpacker,
-	}).SetupWithManager(mgr); err != nil {
-		setupLog.Error(err, "unable to create controller", "controller", rukpakv1alpha1.BundleKind)
+	if err := bundle.SetupProvisioner(mgr, append(
+		commonBundleProvisionerOptions,
+		bundle.WithProvisionerID(registry.ProvisionerID),
+		bundle.WithHandler(bundle.HandlerFunc(registry.HandleBundle)),
```
akihikokuroda, line 236: It took some time for me to understand this line.  Is this a popular way to pass an interface as an argument?

internal/provisioner/bundledeployment/bundledeployment.go
```diff
@@ -409,19 +471,19 @@ func isResourceNotFoundErr(err error) bool {
 }
 
 // SetupWithManager sets up the controller with the Manager.
-func (r *BundleDeploymentReconciler) SetupWithManager(mgr ctrl.Manager) error {
+func (p *bundledeploymentProvisioner) SetupWithManager(mgr ctrl.Manager) error {
```
akihikokuroda, line 474: Is this still be used?

internal/provisioner/registry/types.go
```diff
@@ -0,0 +1,28 @@
+package registry
```
akihikokuroda, line 1: Should this file be renamed to retistry.go for consistency?

akihikokuroda:
This looks good.  I have some minor comments.

cmd/core/main.go
joelanford: This seems common with single-method interfaces. I think the most well-known one is `http.Handler`/`http.HandlerFunc`.

internal/provisioner/bundle/bundle.go
```diff
@@ -1,59 +1,125 @@
...
+
+func (p *bundleProvisioner) setDefaults() {
+	if p.handler == nil {
+		p.handler = HandlerFunc(func(_ context.Context, fsys fs.FS, _ *rukpakv1alpha1.Bundle) (fs.FS, error) { return fsys, nil })
+	}
+}
+
+func (p *bundleProvisioner) validateConfig() error {
+	errs := []error{}
+	if p.handler == nil {
```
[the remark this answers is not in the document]
joelanford: I was trying to separate concerns a little bit and also have defensive code in the validate function (i.e. don't assume that `setDefaults` exists or ensures handler is always set).

I can combine into a single function if there's consensus.

internal/provisioner/bundledeployment/bundledeployment.go
joelanford: Don't think so. Good catch.

akihikokuroda (approved):

tylerslaton:
The code here looks good and it is blocking some work that @akihikokuroda is doing. I'm going to go ahead and hit the merge button on @joelanford's behalf as he is on PTO for awhile.

PR #578, opened 2022-09-26
test(e2e): add ability to configure the image registry for test images

tylerslaton:
Adding some small changes to enable testing to be a bit more configurable. This will allow downstream consumers of Rukpak to specify what image registry that their test images are served from.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -95,8 +95,9 @@ test-unit: setup-envtest ## Run the unit tests
 	eval $$($(SETUP_ENVTEST) use -p env $(ENVTEST_VERSION)) && go test -tags $(GO_BUILD_TAGS) -count=1 -short $(UNIT_TEST_DIRS)
 
 FOCUS := $(if $(TEST),-v -focus "$(TEST)")
+E2E_FLAGS ?= ""
 test-e2e: ginkgo ## Run the e2e tests
-	$(GINKGO) --tags $(GO_BUILD_TAGS) -trace -progress $(FOCUS) test/e2e
+	$(GINKGO) --tags $(GO_BUILD_TAGS) $(E2E_FLAGS) -trace -progress $(FOCUS) test/e2e
 
 e2e: KIND_CLUSTER_NAME=rukpak-e2e
 e2e: rukpakctl run image-registry kind-load-bundles registry-load-bundles test-e2e kind-cluster-cleanup ## Run e2e tests against an ephemeral kind cluster
--- a/test/e2e/plain_provisioner_test.go
+++ b/test/e2e/plain_provisioner_test.go
@@ -67,7 +67,7 @@ var _ = Describe("plain provisioner bundle", func() {
 					Source: rukpakv1alpha1.BundleSource{
 						Type: rukpakv1alpha1.SourceTypeImage,
 						Image: &rukpakv1alpha1.ImageSource{
-							Ref: "testdata/bundles/plain-v0:valid",
+							Ref: fmt.Sprintf("%v/%v", ImageRepo, "plain-v0:valid"),
 						},
 					},
 				},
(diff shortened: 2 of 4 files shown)


timflannagan (commented):
Do we anticipate the need to configure the image repository and tag here?

Makefile
```diff
@@ -95,8 +95,9 @@ test-unit: setup-envtest ## Run the unit tests
 	eval $$($(SETUP_ENVTEST) use -p env $(ENVTEST_VERSION)) && go test -tags $(GO_BUILD_TAGS) -count=1 -short $(UNIT_TEST_DIRS)
 
 FOCUS := $(if $(TEST),-v -focus "$(TEST)")
+IMAGE_REGISTRY := "testdata"
```
exdx, line 98: super nit: should this be `?=` since we are setting the variable to `testdata` if no value is provided? 

https://www.gnu.org/software/make/manual/html_node/Setting.html

exdx (approved):

tylerslaton:
> Do we anticipate the need to configure the image repository and tag here?

I'm not anticipating needing to do tags. I mainly say this because we can set it up across environments to have the same tags but the repository would stay the same.

Makefile
tylerslaton: Ah, yep you're totally right. Today I learned!

tylerslaton:
> Do we anticipate the need to configure the image repository and tag here?

@timflannagan I'm not anticipating needing to do tags. I mainly say this because we can set it up across environments to have the same tags but the repository would stay the same.

Makefile
```diff
@@ -95,8 +95,9 @@ test-unit: setup-envtest ## Run the unit tests
 	eval $$($(SETUP_ENVTEST) use -p env $(ENVTEST_VERSION)) && go test -tags $(GO_BUILD_TAGS) -count=1 -short $(UNIT_TEST_DIRS)
 
 FOCUS := $(if $(TEST),-v -focus "$(TEST)")
+IMAGE_REGISTRY ?= "testdata"
```
joelanford, line 98: I expected to see some changes in `kind-load-bundles` to reflect the fact that the registry name is now coming from this variable.

Also, should we go ahead and make this `IMAGE_REPO ?= "testdata/bundles"`? It feels like the "bundles" repo name is not something we can always count on containing the images we want.

test/e2e/e2e_suite_test.go
```diff
@@ -22,9 +22,10 @@ import (
 )
 
 var (
-	cfg        *rest.Config
-	c          client.Client
-	kubeClient kubernetes.Interface
+	cfg           *rest.Config
+	c             client.Client
+	kubeClient    kubernetes.Interface
+	ImageRegistry string
```
joelanford, line 28: Can we put this in a separate file with a comment about the fact that it's expected to be populated via build flags? Should there be a default value (e.g. `testdata`)?

test/e2e/plain_provisioner_test.go
```diff
@@ -67,7 +67,7 @@ var _ = Describe("plain provisioner bundle", func() {
 					Source: rukpakv1alpha1.BundleSource{
 						Type: rukpakv1alpha1.SourceTypeImage,
 						Image: &rukpakv1alpha1.ImageSource{
-							Ref: "testdata/bundles/plain-v0:valid",
+							Ref: ImageRegistry + "/bundles/plain-v0:valid",
```
joelanford, line 70: Nit: use `fmt.Sprintf`?

test/e2e/e2e_suite_test.go
tylerslaton: Sure, that seems like a good way to clarify this.

tylerslaton: For the default value, what would you think about creating a default and then in the Makefile changing it to look like this?

```
FOCUS := $(if $(TEST),-v -focus "$(TEST)")
E2E_FLAGS ?= ""
test-e2e: ginkgo ## Run the e2e tests
	$(GINKGO) --tags $(GO_BUILD_TAGS) $(E2E_FLAGS) -trace -progress $(FOCUS) test/e2e
```

joelanford: That seems pretty good. Then downstream we can override E2E_FLAGS, and that's hopefully flexible enough for other cases that might pop up.

tylerslaton: Cool, those changes are up now. With that in do you see anything outstanding?

PR #581, opened 2022-09-29
add resources/requests to containers

### LINKED ISSUE #571: The kustomize manifests for the provisioner deployments should specify requests
The current core and webhook deployments don't specify a requests configuration which goes against kubernetes best practices.

akihikokuroda:

Closes #571

All the patches for the resources/requests are in the manifests/kustomize.yaml file.  The `vars` of the kustomize seems better way but it is [deprecated](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/vars/).

### DIFF
--- a/manifests/provisioners/helm/resources/deployment.yaml
+++ b/manifests/provisioners/helm/resources/deployment.yaml
@@ -44,6 +44,10 @@ spec:
           volumeMounts:
             - name: certs
               mountPath: /etc/pki/tls
+          resources:
+            requests:
+              cpu: 1m
+              memory: 15Mi
         - name: manager
           securityContext:
             allowPrivilegeEscalation: false
@@ -66,6 +70,10 @@ spec:
               mountPath: /var/cache/bundles
             - name: certs
               mountPath: /etc/pki/tls
+          resources:
+            requests:
+              cpu: 10m
+              memory: 160Mi
       volumes:
         - name: bundle-cache
           emptyDir: {}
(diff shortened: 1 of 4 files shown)


timflannagan (commented):
Do we have a need for using patches here vs. just updating the default manifests with the desired resources configuration? I think if we were managing multiple overlays, then having a patch makes sense.

akihikokuroda:
I put everything in the base kustomization.yaml to make it easier to update them.  Either way works.  I can rework if there is a better way for this time.

timflannagan:
I think my preference would be to either have dedicated patch files for injecting these resource configurations, or adding them explicitly to the inline YAML. And the rationale for this, is that it's a bit harder (at least imo) for me to find this configuration if it's housed in the kustomization.yaml file. Thoughts?

akihikokuroda:
OK.  I'll put them in the deployment YAML files.

manifests/provisioners/helm/resources/deployment.yaml
```diff
@@ -66,6 +70,10 @@ spec:
               mountPath: /var/cache/bundles
             - name: certs
               mountPath: /etc/pki/tls
+          resources:
+            requests:
+              cpu: 10m
+              memory: 160Mi
```
timflannagan, line 76: I think this is something we can tune over time. What I had landed on for a component that uses rukpak was [100Mi](https://github.com/openshift/platform-operators/blob/main/manifests/0000_50_cluster-platform-operator-manager_04-rukpak-core.deployment.yaml#L78) given that's what OLM is configured to use.

We do store an in-memory bundle cache, which may warrant some benchmarking to get a sense on the right request values here.

timflannagan (approved):

timflannagan:
I'm cool with merging this without a second approver if you want to hit the green button.

akihikokuroda:
@timflannagan I'll do that.  Thanks!

PR #586, opened 2022-10-06
local git repository for e2e test

### LINKED ISSUE #295: Use local sources for git-bundle based e2e tests
The current set of git bundle e2e tests reference a remote github repository which is configured as a plain bundle git repo. Because the tests need to reachout to github.com to clone the repository it is subject to rate-limiting and potential flakiness. It would be optimal to use "local" git remotes for our e2e testing purposes like we currently do with image-based Bundle source types (which get loaded onto the cluster after being build from the local filesystem). This would help alleviate a potential source of flakes and allow for a better testing experience.

akihikokuroda:

Closes #295 

Adding an in-cluster git repository for the e2e test.  It has a clone of **combo** repository in it now.  I added one test for using the repository.  We can change some more tests to use it if it's good to do so.


.gitmodules
```diff
@@ -0,0 +1 @@
+
```
tylerslaton, line 1: Do we need this file if its empty?

tylerslaton (approved):
This looks good to me. I'd advocate for moving `tools/git` into `test/testutil` instead though.

.gitmodules
akihikokuroda: Probably not.  I tried once to use the submodule so it was generated but I decided not use it.  I'll take it out.  Thanks!

akihikokuroda:
> I'd advocate for moving tools/git into test/testutil instead though.

The test/testutil has some go files used in the tests.  The tools has imageregistry now so I added git there.  I can move both imageregistry and git under test/tools.

tylerslaton:
> The test/testutil has some go files used in the tests. The tools has imageregistry now so I added git there. I can move both imageregistry and git under test/tools.

That sounds reasonable to me. We only use either of those utilities in a testing environment.

akihikokuroda:
@tylerslaton I relocated the tools/imageregistry and tools/git to test/tools/imageregistry and test/tools/git.

tylerslaton:
> @tylerslaton I relocated the tools/imageregistry and tools/git to test/tools/imageregistry and test/tools/git.

This looks good to merge to me then. @operator-framework/rukpak-maintainers any additional thoughts?

tylerslaton:
@akihikokuroda Anything blocking merging here? Looks like no one was outspoken about it not merging.

PR #588, opened 2022-10-10
Refactor source/git and add unit tests 

asmacdo:
The git.Unpack called into the git library too much to be reasonably unit testable. Some functions broken out to improve coverage.

To test the ssh cert usage, we needed a real private key, which I generated for this purpose. We need .gitleaks.toml to identify this private key as safe to upload.

Signed-off-by: Austin Macdonald <austin@redhat.com>


asmacdo:
/hold
The unit tests can pass, but this PR introduces a flake that I could use some help figuring out.

```
• [FAILED] [0.003 seconds]
git.createCloneOptions
/home/austin/devel/rukpak/internal/source/git_test.go:55
  [It] should set Reference name to Branch
  /home/austin/devel/rukpak/internal/source/git_test.go:94

  Expected
      <*fmt.wrapError | 0xc00069d2c0>: {
          msg: "error listening on :8080: listen tcp :8080: bind: address already in use",
          err: <*net.OpError | 0xc0007c6eb0>{
              Op: "listen",
              Net: "tcp",
              Source: nil,
              Addr: <*net.TCPAddr | 0xc0009953b0>{IP: nil, Port: 8080, Zone: ""},
              Err: <*os.SyscallError | 0xc00069d2a0>{
                  Syscall: "bind",
                  Err: <syscall.Errno>0x62,
              },
          },
      }
  to be nil
  In [It] at: /home/austin/devel/rukpak/internal/source/git_test.go:97
```

akihikokuroda:
Can you do this

```
		client := fake.NewClientBuilder().WithScheme(scheme).Build()
		git := Git{client, "fakenamespace"}
```
instead of
```
		cfg := ctrl.GetConfigOrDie()
		mgr, err := ctrl.NewManager(cfg, ctrl.Options{})
		Expect(err).To(BeNil())
		git := Git{mgr.GetClient(), "fakenamespace"}
```
so that you don't need to create the manager?

internal/source/git.go
```diff
@@ -126,8 +107,40 @@ func (r *Git) Unpack(ctx context.Context, bundle *rukpakv1alpha1.Bundle) (*Resul
 	return &Result{Bundle: bundleFS, ResolvedSource: resolvedSource, State: StateUnpacked}, nil
 }
 
+func (r *Git) createCloneOptions(bundle *rukpakv1alpha1.Bundle, progress bytes.Buffer) (*git.CloneOptions, error) {
```
timflannagan, line 110: Should we refactor this method to be a function since there's no internal Git structure state being utilized in this function?

internal/source/git_test.go
```diff
@@ -0,0 +1,329 @@
...
+		Expect(err).To(BeNil())
+		git := Git{mgr.GetClient(), "fakenamespace"}
+		bundle := &rukpakv1alpha1.Bundle{
+			Spec: rukpakv1alpha1.BundleSpec{
+				Source: rukpakv1alpha1.BundleSource{
+					Type: "faketype",
+				},
+			},
+		}
+		_, err = git.Unpack(context.TODO(), bundle)
```
timflannagan, line 34: nit: we should use context.Background for tests to remain consistent with the rest of the codebase.

internal/source/git.go
```diff
@@ -39,41 +39,21 @@ func (r *Git) Unpack(ctx context.Context, bundle *rukpakv1alpha1.Bundle) (*Resul
...
-	if bundle.Spec.Source.Git.Auth.Secret.Name != "" {
-		auth, err := r.configAuth(ctx, bundle)
-		if err != nil {
-			return nil, fmt.Errorf("configuring Auth error: %w", err)
-		}
-		cloneOpts.Auth = auth
+	// Set options for clone
+	cloneOpts, err := r.createCloneOptions(bundle, progress)
+	if err != nil {
+		return nil, fmt.Errorf("bundle clone options error: %v", err)
```
timflannagan, line 46: Is there any value in returning a wrapped error here? WYDT?

internal/source/git.go
```diff
@@ -126,8 +107,40 @@ func (r *Git) Unpack(ctx context.Context, bundle *rukpakv1alpha1.Bundle) (*Resul
...
+	}
+
+	if gitsource.Ref.Branch != "" && gitsource.Ref.Tag != "" {
+		return nil, errors.New("git branch and tag cannot be set simultaneously")
+	}
+	if gitsource.Ref.Branch != "" {
+		cloneOpts.ReferenceName = plumbing.ReferenceName(fmt.Sprintf("refs/heads/%s", gitsource.Ref.Branch))
+		cloneOpts.SingleBranch = true
+		cloneOpts.Depth = 1
+	} else if gitsource.Ref.Tag != "" {
```
timflannagan, line 131: nit: we can improve readability here if we remove the if/else conditional structure, and exit early in the if statement by returning the cloneOpts variable there.

timflannagan (commented):
I took a super quick look at a had a couple of nits/comments/etc. I'll try and take a better look later today or sometime this week.

internal/source/git.go
asmacdo: makes sense, but I did it this way so we could validate against setting Branch and Tag. If we return after Branch, then Branch+Tag simultaneously is legal, but Branch wins.

PR #591, opened 2022-10-24
.github/workflows: Get Go version from root go.mod file

timflannagan:
Use the actions/setup-go v3 major version, which has support for configuring Go based on the version specified in the root go.mod file. This reduces the amount of changes needed to bump Go versions over time.


timflannagan:
(recycling checks)

Issue #131, opened 2022-03-08
The plain-v0 provisioner cannot rollout Bundles containing CRDs and instances of those CRDs


timflannagan:
When deploying the plain-v0 provisioner using `make run KIND_CLUSTER_NAME=kind` and creating the following Bundle/BundleInstance resources:

```yaml
apiVersion: core.rukpak.io/v1alpha1
kind: Bundle
metadata:
  name: olm-v0.20.0
spec:
  image: quay.io/tflannag/olm-plain-bundle:olm-v0.20.0-combined
  provisionerClassName: core.rukpak.io/plain-v0

---

apiVersion: core.rukpak.io/v1alpha1
kind: BundleInstance
metadata:
  name: olm-v0.20.0
spec:
  bundleName: olm-v0.20.0
  provisionerClassName: core.rukpak.io/plain-v0
```

The provisioner fails to rollout that Bundle, with the BundleInstance reporting a failed state:

```bash
$ k get bundleinstances olm-v0.20.0 -o yaml
apiVersion: core.rukpak.io/v1alpha1
kind: BundleInstance
metadata:
  ...
spec:
  bundleName: olm-v0.20.0
  provisionerClassName: core.rukpak.io/plain-v0
status:
  conditions:
  - lastTransitionTime: "2022-03-08T17:42:25Z"
    message: 'error while running post render on files: [unable to recognize "": no
      matches for kind "CatalogSource" in version "operators.coreos.com/v1alpha1",
      unable to recognize "": no matches for kind "ClusterServiceVersion" in version
      "operators.coreos.com/v1alpha1", unable to recognize "": no matches for kind
      "OLMConfig" in version "operators.coreos.com/v1", unable to recognize "": no
      matches for kind "OperatorGroup" in version "operators.coreos.com/v1"]'
    reason: InstallFailed
    status: "False"
    type: Installed
```

timflannagan:
Poked around the helm chart library and I'm fairly sure I know how to address this, so I'm going to take a stab at it and assign myself.

joelanford:
If your idea is to separate the CRDs out into Helm's separate CRD handling -- think again.

I'm pretty sure Helm will ignore those CRDs during upgrades.  :)

timflannagan:
[...]

PR #603, opened 2022-11-29
WIP: Crd plus cr install

dtfranz:
PR to address issue #131 

As noted in the issue, helm has limitations around upgrading CRDs. This was initially circumvented in the plain provisioner by applying the CRDs as 'normal' files without using helm's built-in CRD management, but this creates an issue when installing a manifest which includes CRDs as well as instances of those CRDs. The proposed solution is to:
1. Place the CRDs into helm's managed `crds/` folder.
2. Set helm's `SkipCRDs` flag to true for both install and upgrade actions.
3. Create or Update the CRDs ourselves.


internal/provisioner/bundledeployment/bundledeployment.go
```diff
@@ -255,6 +258,33 @@ func (p *bundledeploymentProvisioner) reconcile(ctx context.Context, bd *rukpakv
...
+			if err != nil {
+				meta.SetStatusCondition(&bd.Status.Conditions, metav1.Condition{
+					Type:    rukpakv1alpha1.TypeInstalled,
+					Status:  metav1.ConditionFalse,
+					Reason:  rukpakv1alpha1.ReasonInstallFailed,
+					Message: err.Error(),
+				})
+				return ctrl.Result{}, err
+			}
+			err = crd.CreateOrUpdateCRD(ctx, p.cl, &c)
```
dtfranz, line 276: Seeing test failures presumably from failing to set the controller reference on the CRDs - will add.

internal/crd/crd.go
```diff
@@ -63,6 +64,38 @@ func Validate(ctx context.Context, cl client.Client, newCrd *apiextensionsv1.Cus
 	return nil
 }
 
+// CreateOrUpdateCRD performs all the necessary actions to either Create or Update a CRD,
+// including running validation when an Update is required.
+func CreateOrUpdateCRD(ctx context.Context, cl client.Client, newCrd *apiextensionsv1.CustomResourceDefinition) error {
+	// First try to create the CRD
```
joelanford, line 70: What's the code path for the case where the CRD already exists and is identical to `newCRD`? It seems like we try (and fail) to `Create`, and then we `Update`, with the end result being a no-op.

Do you think we can:
1. `Get`
2. If not exist, `Create`
3. If exist, compare
4. If diff, update.

That way we don't hit the apiserver at all when there's nothing to do.

internal/provisioner/bundledeployment/bundledeployment.go
```diff
@@ -291,12 +321,14 @@ func (p *bundledeploymentProvisioner) reconcile(ctx context.Context, bd *rukpakv
...
 		rel, err = cl.Install(bd.Name, p.releaseNamespace, chrt, values, func(install *action.Install) error {
 			install.CreateNamespace = false
+			install.SkipCRDs = true
 			return nil
 		},
 			// To be refactored issue https://github.com/operator-framework/rukpak/issues/534
 			func(install *action.Install) error {
 				post.cascade = install.PostRenderer
 				install.PostRenderer = post
+				install.SkipCRDs = true
```
joelanford, line 331: No need to add this twice. The setting in like 324 should suffice.

dtfranz: Great, will remove. Followup for that though - why the two separate install actions?

internal/provisioner/plain/plain.go
```diff
@@ -19,9 +19,9 @@ import (
 
 const (
 	// ProvisionerID is the unique plain provisioner ID
-	ProvisionerID = "core-rukpak-io-plain"
-
-	manifestsDir = "manifests"
+	ProvisionerID            = "core-rukpak-io-plain"
+	customResourceDefinition = "CustomResourceDefinition"
+	manifestsDir             = "manifests"
```
joelanford, line 24: _total_ nit: group plain provisioner specific constants separately from sort of general helper constants?
```suggestion
	ProvisionerID = "core-rukpak-io-plain"
	manifestsDir  = "manifests"
	
	customResourceDefinition = "CustomResourceDefinition"
```

Also, IMO since there's just one place we use this constant, maybe it's okay to just hardcode `"CustomResourceDefinition"` in the comparison and drop the constant?

dtfranz: Yup, makes sense to me. Will update. I was hoping to find an external const I could use in the apiextensionsv1 pkg but came up empty :(

internal/crd/crd.go
```diff
@@ -63,6 +64,38 @@ func Validate(ctx context.Context, cl client.Client, newCrd *apiextensionsv1.Cus
...
+	err := cl.Create(ctx, newCrd)
+	if err != nil && apierrors.IsAlreadyExists(err) {
+		// If it exists already check if it's safe to update
+		err = Validate(ctx, cl, newCrd)
+		if err != nil {
+			// If not, return reason why we cannot update safely
+			return err
+		}
```
joelanford, line 78: I think we should drop this `Validate` check. IIRC, we use this in the CRD validator webhook, and I'm pretty sure our plan is still to get rid of that, or at least turn it off.

joelanford:
Just a few comments, but I think the approach looks solid and sound.

Looks like some of the e2e tests need to be updated to account for the fact that rukpak can now support bundles that contain both CRDs and instances of those CRDs.

internal/provisioner/bundledeployment/bundledeployment.go
joelanford: The one with the PostRender stuff is going to be completely factored out when that helm-operator-plugins change is available (I think it's on `helm-operator-plugins` main branch now, so we might be able to go ahead and do that).

I think the thought was to keep it separate from the other install configurations to make the refactoring a clean removal of that second inline function.

But if we want to combine into one right now, that's also fine with me.

internal/crd/crd.go
joelanford: Yeah, that's a good question. I think we generally _always_ want to take _exactly_ what's in `new` and essentially erase stuff in `old` that isn't in `new`. 

But you're right that there are exceptions. Off the top of my head, I'd exclude:
- everything in `metadata` except name, namespace, labels, and annotations.
- all of `status`

Does `last-applied-configuration` show up in rukpak-managed CRDs right now?

internal/provisioner/bundledeployment/bundledeployment.go
joelanford: There's a controller-runtime utility that makes this easy. https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.13.1/pkg/controller/controllerutil#SetControllerReference

joelanford:
Oh one more thing. I think there's some code that adds dynamic watches to the controller based on the contents of the helm release manifest. Since the CRDs are no longer in the manifest, we probably need to update that code to make sure that we start watching CRDs the first time we reconcile a BD that contains a CRD.

joelanford:
Oh crap, one more thing. In the helm `Install` function, we handle install failure by uninstalling the failed release. And similarly in the `Upgrade` function, we handle upgrade failure by rolling back to the previous release just before the failed upgrade release.

In the current code, that uninstall/rollback includes deleting and patching back the CRDs because they are in the release manifest.  With this PR, it seems like we lose the CRD uninstall/rollback?

awgreene (changes requested):
Looks good, my comments are mostly nits, but I agree with the design changes that @joelanford proposed.

internal/crd/crd.go
```diff
@@ -63,6 +64,38 @@ func Validate(ctx context.Context, cl client.Client, newCrd *apiextensionsv1.Cus
 	return nil
 }
 
+// CreateOrUpdateCRD performs all the necessary actions to either Create or Update a CRD,
+// including running validation when an Update is required.
+func CreateOrUpdateCRD(ctx context.Context, cl client.Client, newCrd *apiextensionsv1.CustomResourceDefinition) error {
+	// First try to create the CRD
+	err := cl.Create(ctx, newCrd)
+	if err != nil && apierrors.IsAlreadyExists(err) {
```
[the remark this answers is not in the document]
dtfranz: This looks much better, thanks!

internal/provisioner/bundledeployment/bundledeployment.go
```diff
@@ -255,6 +258,33 @@ func (p *bundledeploymentProvisioner) reconcile(ctx context.Context, bd *rukpakv
 		})
 		return ctrl.Result{}, err
 	}
+	if len(chrt.CRDObjects()) > 0 {
+		// Install the CRDs ourselves before installing the rest of the helm chart
+		// This allows us to work around helm's limitations around updating CRDs
+		for _, crdFile := range chrt.CRDObjects() {
+			var c apiextensionsv1.CustomResourceDefinition
+			err := yaml.Unmarshal(crdFile.File.Data, &c)
+			if err != nil {
```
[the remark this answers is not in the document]
dtfranz: No particular reason - just a habit I picked up somewhere

dtfranz:
> Oh crap, one more thing. In the helm `Install` function, we handle install failure by uninstalling the failed release. And similarly in the `Upgrade` function, we handle upgrade failure by rolling back to the previous release just before the failed upgrade release.
> 
> In the current code, that uninstall/rollback includes deleting and patching back the CRDs because they are in the release manifest. With this PR, it seems like we lose the CRD uninstall/rollback?

This is really important behavior to keep. Though I wonder, since we leave the CRDs in helm and are only skipping them on install and upgrade, will helm still remove them when we uninstall/rollback? I'll check this before moving forward.

dtfranz:
> I'm not sure the changes related to uninstall/rollback handling are sufficient (pending your investigation of what helm might give us for free. I have my doubts that it ever changes existing CRDs though, and it might not even delete them during `helm uninstall`). We also need to handle the scenario where the CRD are correctly applied just fine, but the release install/upgrade fails.
> 
> In that case, we also want to uninstall/rollback the CRDs.

Right, of course. I'll see how I can get that in. Unfortunately I don't think we get anything from helm for free, since the rollback is performed against the release manifest returned from the install/update, and the CRDs are missing from that manifest when we ask helm to skip them.

PR #604, opened 2022-12-22
Remote debug make target

dtfranz:
Adds `debug` make target to rukpak. The `debug` target is essentially the same as the `run` target except that, once finished, it exposes a local port to allow for remote debugging in the rukpak `core` container. This is done with the following:
* kind config to establish nodePort mapping
* deployment patches which:
  * add an initContainer to install `dlv`
  * prepend the `core` container command and args to run with dlv
* a service to link the `dlv` listener to the exposed nodePort

README is also updated to document the `debug` target and provides example launch config for vscode debugging.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -109,7 +109,7 @@ e2e: KIND_CLUSTER_NAME=rukpak-e2e
 e2e: rukpakctl run image-registry local-git kind-load-bundles registry-load-bundles test-e2e kind-cluster-cleanup ## Run e2e tests against an ephemeral kind cluster
 
 kind-cluster: kind kind-cluster-cleanup ## Standup a kind cluster
-	$(KIND) create cluster --name ${KIND_CLUSTER_NAME}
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -124,14 +124,15 @@ local-git: ## Setup in-cluster git repository
...
+.PHONY: install install-manifests wait run debug debug-helper cert-mgr uninstall
 
 ##@ install/run:
 
 install: generate cert-mgr install-manifests wait ## Install rukpak
 
+MANIFESTS_DIR ?= manifests
 install-manifests:
-	$(KUBECTL) apply -k manifests
+	$(KUBECTL) apply -k $(MANIFESTS_DIR)
```
kevinrizza, line 135: doesn't this mean that even for normal (non debug) install that you're going to get the debug manifests?

dtfranz: By default the manifests dir is set to the old value, `manifests` on line 133, so unless the `debug` target is used nothing from the debug manifests should be included with the apply command.

perdasilva (approved):
Great job! Thank you ^^

PR #607, opened 2023-02-21
Fix e2e: Force localhost for some containers

tmshort:
Explicitly specify localhost for some containers, otherwise docker.io ends up being used as the default repository.


perdasilva (approved):
lgtm

PR #609, opened 2023-03-02
OCI Artifacts Source

anik120:
Ref Design doc: https://hackmd.io/nhAivsWVSJ6Jt0zniadDZA


anik120:
cc: @joelanford @bparees

docs/provisioners/oci_artifacts.md
```diff
@@ -0,0 +1,141 @@
+# Plain Provisioner
```
bparees, line 1: OCI Artifacts Provisioner?

that said, this seems like abstraction bleed-through.  I would have expected provisioners to care about the content, not the source of that content.  So i'm surprised to see this being defined as a provisioner type.

I'd expect to have different "bundle retrieval" plugins (http, git, OCI artifact, traditional image) which know how to extract the file content, and then different provisioner plugins (plain, registryv1, etc) that know how to deal with the different kinds of file content we can expect to see.

I realize there's some bleed through because the type of bundle (plain, registryv1) may influence what files need to be extracted, but i'm still leery of a world where any new bundle format being introduced means we need to update all the bundle retrieval plugins (vs just introducing one new provisioner), and i think we can build a system that doesn't require that.

docs/provisioners/oci_artifacts.md
```diff
@@ -0,0 +1,141 @@
...
+  name: my-bundle-deployment
+spec:
+  provisionerClassName: core-rukpak-io-plain
+  template:
+    metadata:
+      labels:
+        app: my-bundle
+    spec:
+      source:
+        type: image
```
bparees, line 37: shouldn't this be `oci_artifacts` per bundle_types.go above?

docs/provisioners/oci_artifacts.md
```diff
@@ -0,0 +1,141 @@
...
+        image:
+          ref: my-bundle@sha256:xyz123
+      provisionerClassName: core-rukpak-io-oci-artifacts
```
bparees, line 40: reiterating my comment above, i would not have expected a new provisioner type to be introduced for a new bundle source type.

docs/provisioners/oci_artifacts.md
```diff
@@ -0,0 +1,141 @@
...
+
+As the bundle content is unpacked onto the cluster via the defined storage mechanism, the bundle status will be updated to Unpacked, indicating that all its contents have been stored on-cluster.
+
+```console
+$ kubectl get bundle my-bundle
+NAME           TYPE    PHASE      AGE
+my-bundle      oci_artifacts   Unpacked   10s
+```
+
+Now that the bundle has been unpacked, the `plain` provisioner is able to create the resources in the bundle on the cluster.
```
bparees, line 54: so this implies it is the plain provisioner, so does that mean `core-rukpak-io-oci-artifacts` is then calling the plain provisioner?  I think that further argues for my claim that `core-rukpak-io-oci-artifacts` should not be considered a provisioner, it's a bundle source type/extractor.

api/v1alpha1/bundle_types.go
```diff
@@ -110,6 +113,10 @@ type HTTPSource struct {
 	Auth Authorization `json:"auth,omitempty"`
 }
 
+type OCIArtifactSource struct {
+	// Ref contains the reference to a OCI Artifact Image containing Bundle contents.
+	Ref string `json:"ref"`
```
bparees, line 118: Since we're pulling the content directly, I assume a fully fleshed out version of this api would have to add fields for things like:

1) pull secret (presumbly a ref to a secret in the same NS)

and OLM itself would need a configuration for (or again respect the cluster's configuration for):
1) allowed/denied registries
2) image signatures
3) additional trusted CAs
4) proxy

some of these are applicable to the http and git source types also.

docs/provisioners/oci_artifacts.md
joelanford: > I'd expect to have different "bundle retrieval" plugins (http, git, OCI artifact, traditional image) which know how to extract the file content, and then different provisioner plugins (plain, registryv1, etc) that know how to deal with the different kinds of file content we can expect to see.

Yes, Ben is correct. The general idea is that there are sources which convert `bundle.spec.source` to a Go `fs.FS`. Thus far all of the provisioners use identical source implementations for consistency. At that point the provisioners job is focused on the contents of that `fs.FS` (i.e. plain vs. registry+v1 vs. helm vs. kustomize, etc.)

joelanford (commented):
I'm trying to understand the motivation for defining a new bundle distribution mechanism via OCI artifacts vs standard scratch container images. Right now the way that rukpak defines a bundle is "anything that can be represented as a go `fs.FS`".

Rukpak already has the `image` source that just straight up pulls the entire filesystem out of an image.

Is the focus of this more about a different _way_ to pull the existing content out of existing bundle images? As in the provisioner itself reaches out directly to the image registry to pull the content rather than the existing complexity of creating a pod and delegating the image pull/extraction to the CRI and unpack code we've written.

If that's the idea, I'd propose modifying the `image` source specification, perhaps to allow specifying a `method` which could be either "CRI" or "direct". At the end of the day, its still the same image we're just pulling it differently?

Or is this saying "there's a different protocol and/or format that necessitates a completely new source definition"?

internal/source/artifact_image.go
```diff
@@ -0,0 +1,49 @@
...
+		return nil, err
+	}
+	defer tmpfs.Close()
+	refComponents := strings.Split(bundle.Spec.Source.Image.Ref, ":")
+	repo, err := remote.NewRepository(refComponents[0])
+	if err != nil {
+		return nil, err
+	}
+	// 2. Copy from the remote repository to the file store
+	manifestDescriptor, err := oras.Copy(ctx, repo, refComponents[1], tmpfs, refComponents[1], oras.DefaultCopyOptions)
```
everettraven, line 39: I'm curious if there is a chance that this call blocks the reconciliation for an extended period of time. I know that the OCI artifacts are meant to be on the smaller size for pulling anyways, but I can imagine a scenario where someone's OCI artifact bundle is fairly large and pulling it takes significantly longer than usual or expected.

IMO we should probably take this into consideration and prevent this from happening. Off the top of my head, one way to do this would be to use unpack `Job`s similar to how the current `image` source unpacks bundles. The only difference here is that in the `Pod` created by the `Job` we could use something like the `oras` cli to copy the bundle information to a local directory and then run the same `/bin/unpack` command the `image` source does. While this is a significantly more complex implementation than you have here, I think the biggest benefit is that we eliminate a case where reconciliation gets blocked and holds up the reconciliation process for other `BundleDeployment`s.

One thing that I think could be useful to prevent a large amount of code duplication is to update the `image` source to maybe instead be an `oci` source that can work with both images *and* OCI artifacts and make adjustments accordingly.

internal/source/artifact_image.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+func (i *ArtifactImage) Unpack(ctx context.Context, bundle *rukpakv1alpha1.Bundle) (*Result, error) {
+	if bundle.Spec.Source.Type != rukpakv1alpha1.SourceTypeOCIArtifacts {
+		return nil, fmt.Errorf("bundle source type %q not supported", bundle.Spec.Source.Type)
+	}
+	if bundle.Spec.Source.Image == nil {
+		return nil, fmt.Errorf("bundle source image configuration is unset")
+	}
+	tmpfs, err := file.New(unpackLocation)
```
everettraven, line 28: Do these temp files ever get cleaned up after a certain amount of time? As far as I understand it, Kubernetes treats the `/tmp` directory as an in memory file store on the pod and contributes to it's overall memory consumption. I couldn't find any information on whether or not kube will automatically clean these up after a certain amount of time so I am a bit concerned that we will just be consistently increasing the memory consumption of the pod with this route.

If this is a valid concern, it might be nice to consider using an in memory file store like the [`afero.MemMapFs`](https://pkg.go.dev/github.com/spf13/afero#MemMapFs) which I *think* still implements the Go `fs.FS` interface. I *think* using an implementation like this would allow for the Go garbage collection to automatically free up the memory when it is no longer needed.

everettraven:
> I'm trying to understand the motivation for defining a new bundle distribution mechanism via OCI artifacts vs standard scratch container images. Right now the way that rukpak defines a bundle is "anything that can be represented as a go fs.FS".

@joelanford IIUC, the motivation for defining a new bundle distribution mechanism via OCI artifacts vs images is that using OCI artifacts will make the overall process of building and pushing bundles to an OCI registry easier for operator authors. I don't think that this really impacts rukpak's definition of a bundle as we should still be able to relatively easily convert the bundle OCI artifacts to that of a go fs.FS

> Rukpak already has the image source that just straight up pulls the entire filesystem out of an image.
>
> Is the focus of this more about a different way to pull the existing content out of existing bundle images? As in the provisioner itself reaches out directly to the image registry to pull the content rather than the existing complexity of creating a pod and delegating the image pull/extraction to the CRI and unpack code we've written.

I don't think that is the focus of this, but rather just adding a source implementation that simply knows how to take an OCI artifact and convert it to the necessary go fs.FS similar to how the existing image source works with images.

Just as an aside, if we did want to consider a direct pull rather than offloading to a pod I have some concerns around this causing long running process that block the reconciliation process. IMO we should attempt to avoid situations like that as much as possible. 


> If that's the idea, I'd propose modifying the image source specification, perhaps to allow specifying a method which could be either "CRI" or "direct". At the end of the day, its still the same image we're just pulling it differently?
>
> Or is this saying "there's a different protocol and/or format that necessitates a completely new source definition"?

I definitely agree that this *probably* doesn't warrant a completely new source definition but rather appropriate modifications to the image source specification to make it handle both images and oci artifacts.

joelanford:
@everettraven @anik120 I guess my question is: what is the material difference between using a standard scratch container image to store bundle manifests vs an OCI artifact? And why would an operator author use one over the other?

It's totally possible I'm missing something fundamental, but I'm not seeing what it buys anyone to tar up files in the same (or at least substantially similar) way and then ship that to an image registry using a different schema.

everettraven:
> I guess my question is: what is the material difference between using a standard scratch container image to store bundle manifests vs an OCI artifact? And why would an operator author use one over the other?
>
> It's totally possible I'm missing something fundamental, but I'm not seeing what it buys anyone to tar up files in the same (or at least substantially similar) way and then ship that to an image registry using a different schema.

I think this is a great question. IIUC using OCI artifacts offers a simpler workflow of actually pushing, pulling, and unpacking bundles. Based on my understanding this is essentially what the two workflows look like:
- Image based bundle:
    - Releasing/Pushing a bundle:
        - Create/Maintain a Dockerfile for your bundle that looks something like:
        ```dockerfile
        FROM scratch
        COPY bundle/ bundle/
        ```
        - Create your bundle (`make bundle` or something equivalent)
        - Build your bundle image
        - Push your bundle image
    - Pulling/Consuming a bundle (i.e installing the operator on cluster)
        - Pull the bundle image
        - Unpack the bundle image (depending on how well someone knows image manipulation this isn't necessarily a trivial task)
        - Apply the bundle contents
- OCI artifact based bundle:
    - Releasing/Pushing a bundle:
        - Create your bundle (`make bundle` or something equivalent)
        - Push your bundle OCI artifact
    - Pulling/Consuming a bundle:
        - Pull the bundle artifact
        - Apply the bundle contents

While this may only slightly simplify the workflow, I can imagine a scenario where an operator author would prefer this workflow because there is slightly less overhead and they could tell users to simply install an operator onto a cluster using the raw manifests by running something like:
```
oras pull <bundle>:<version> | kubectl apply -f <bundle-dir>/
```
I'm sure there are other scenarios where OCI artifacts can be used but this is just the one I could think of at the moment. 

While these are pretty similar, I think adding the OCI artifact support to rukpak just makes it more approachable for operator authors who do/want to use OCI artifacts to package their bundles.

>To me, catalog metadata is the place OCI artifacts make the most sense. Rukpak essentially is dealing with the leaf nodes (i.e. the bundle manifests) of the catalogs. And by the time we're dealing with just bundle manifests, it seems like a standard scratch image is more approachable.
>
>There is an argument I can come up with for using a custom schema though: we can have a versioned bundle image spec that the image registry itself knows about. But it seems like our existing image source could have a supported set of bundle image schemas and automatically handle new bundle image schemas. Therefore no need for a new source or new provisioner, I think.

I don't know much about OCI artifacts, but I agree with this for the most part. One question this does raise for me is: While a standard scratch image may be more approachable for rukpak, is it necessarily more approachable for operator authors? 

IMO we should ensure that our tooling will reasonably support the various approaches that operator authors may take to packaging their bundles. That being said, OCI artifacts may be gaining some traction but I am not aware of any operators that currently package their manifests as OCI artifacts as opposed to images/github release artifacts (although I do wonder how much the image delivery format is specifically because of OLM's influence in this area). Due to this, I would *personally* be okay with postponing any bundle as OCI artifact support until there is significant demand for it.

joelanford:
> IIUC using OCI artifacts offers a simpler workflow of actually pushing, pulling, and unpacking bundles.

I think simplicity of push/pull is orthogonal to the schema decision. With a custom schema, we'd need to build and support a custom registry client (likely in `opm`) that could push/pull OCI artifact bundles. It may be that `opm push-bundle <bundleDir>` is easier than `docker build` + `docker push`, and we might want to do that to hide the details from operator authors, but we could build `opm push-bundle` and have it build/push a scratch container under the hood, too.

So to me, the advantage of OCI artifacts is that we can define a completely custom schema and organization of content addressable "stuff" in an image registry. If our current use case is to stuff a bunch of files into a directory structure, I'm not convinced a custom OCI schema is buying us anything.

> OCI artifacts may be gaining some traction but I am not aware of any operators that currently package their manifests as OCI artifacts

OCI artifacts is a spec that (if supported by an image registry) enable service providers to build extensions into the storage mechanisms of the registry. We would be the ones building a specific implementation of that extension mechanism. For example, Helm has defined an OCI artifact custom schema that enables distribution oh helm charts via image registries. We could do a similar thing for OLM bundles, but what I'm poking at is: "why that over a scratch image"? 

I'm not say we _shouldn't_ define our own schema. I'm wanting to get into the nitty gritty of what the advantages are and decide if now is the right time to take that on.

docs/provisioners/oci_artifacts.md
anik120: @joelanford I did however have to do [this](https://github.com/operator-framework/rukpak/pull/609/files#diff-1231be96e91d457fa4b313a8f9421885b600f9be6cb15c7ec4aa978f0f2d6b27R61-R63) though, to allow a `metadata.yaml` file to be included among the artifacts, which I think brings back the discussion we were having [here](https://hackmd.io/nhAivsWVSJ6Jt0zniadDZA#Operator-Bundles-for-OLM-V1) (ref: comment on the words `The only additional, optional file is the olm.bundle.metadata.yaml file`, looks like I can't link specific comments from hackMD 😮‍💨 )

joelanford:
@anik120 The observation that bundles and catalogs should have similar treatment is astute. Though there is one material difference.

I've had conversations with some other OF maintainers over that past couple of days and one thing that kept coming up was: why should catalog data _have_ to come from an image? In rukpak, bundle data can come from an image, git, an http server, a configmap, or a direct upload. 

Perhaps we should design the OLMv1 CatalogSource almost identically to the Rukpak bundle. The bundle API has two key elements:
1. The source of the bundle. How do I pull down the bundle's filesystem? (protocol)
2. The provisioner. How do I interpret that bundle's filesystem? (format)

If we had a similar design for catalogs it would mean:
1. We could support FBC coming from images, http servers, git repos, configmaps, directly uploaded, or any other protocol. We could extend the set of supported sources later.
2. We could simultaneously support multiple catalog formats. Today it's FBC, tomorrow perhaps its something entirely different.

> So this all begs the fundamental question, what's different when it comes to catalog metadata. After all, if we're talking about fbc (or even a sqlite db for that matter), it's a single file(yaml/json/.db) inside a directory structure instead of multiple different files (deployment.yaml/rbac.yaml/sa.yaml etc).

Here's the material difference between bundles and catalogs: bundles are inseparable atomic units, catalogs might be a collection of individually useful units (multiple packages, each of which are valid catalogs on their own).

So the reason I'm saying that a custom OCI artifact type might be more appropriate for catalogs is because of the way updates would work. 

Consider a napkin math scenario where a catalog is 10GB and has 1000 packages with 100 bundles each.

If a single package updates a single bundle or adds a single channel edge, that FBC image layer is now different, so now I'm uploading the entire 10GB again, despite making a maybe 20 byte change. Not only that -- all clients also have to download the full 10GB.

If we sliced the catalog format up into a few separate OCI archive types for form a hierarchy (e.g. catalog, package, channel, bundle), a bundle or channel change is now:
1. Upload the new bundle or channel (up to ~1MB)
2. Update the OCI package metadata to tell it the new reference to the updated bundle or channel (maybe 10s of KBs)
3. Update the OCI catalog metadata to tell it the new reference to the updated package. (maybe 10s of KBs)

So comparing the two:
1. Naive -- stuff the entire catalog into a single image layer -- 10GB upload and 10GB download * number of clients
2. OCI artifacts -- hierarchical layers -- 1MB upload and 1MB download * number of clients.

anik120:
>If we had a similar design for catalogs it would mean:
>We could support FBC coming from images, http servers, git repos, configmaps, directly uploaded, or any other protocol. We could extend the set of supported sources later.
We could simultaneously support multiple catalog formats. Today it's FBC, tomorrow perhaps it's something entirely different.

That actually makes a lot of sense to me. 

> Consider a napkin math scenario

And so does this. 

(it's late on a Friday evening in a week that's been long already, so I'm keeping my response terse 😅 , but there's a lot of action items that's probably coming out of this for the v1 CatalogSource work. I'll come back to this conversation later once those action items are laid out more concretely)

joelanford:
Closing since this served its purpose for demos and there are no plans to merge this.

PR #612, opened 2023-03-29
Update CONTRIBUTING.md

awgreene:


ncdc:
@awgreene still want to proceed with this?

PR #614, opened 2023-04-19
Rukpak Actions Updates

dtfranz:
Updates our github actions to bring the in-line with operator-controller and silence deprecation warnings.

Also fixes the `go-apidiff` action.

### DIFF
--- a/.github/workflows/sanity.yaml
+++ b/.github/workflows/sanity.yaml
@@ -11,8 +11,8 @@ jobs:
   verify:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-go@v3
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v4
         with:
           go-version-file: "go.mod"
       - name: Run verification checks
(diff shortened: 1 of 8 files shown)


dtfranz:
/hold 
until actions health is verified

dtfranz:
/unhold
everything looks to be running well :)

joelanford (approved):

.github/workflows/sanity.yaml
```diff
@@ -21,29 +21,19 @@ jobs:
...
       - name: Checkout repository
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
-      - uses: actions/setup-go@v3
+      - uses: actions/setup-go@v4
         with:
           go-version-file: "go.mod"
 
-      - name: Cache build and module paths
```
timflannagan, line 30: @dtfranz @joelanford Do either of you guys remember why we're removing the cache action here? I was under the impression that the first-party golangci-lint GHA was doing some form of caching under-the-hood that I had to add when removing that action in https://github.com/operator-framework/rukpak/commit/9a4968ea8a35c68a19062d60790c6848e9662491.

The context here is I was looking into doing something similar for my companies' repository, and wanted to see how we solved it back in the day.

dtfranz: The v4 `setup-go` action should be caching the module and build paths automatically, so the additional caching steps aren't necessary anymore. That being said though I do still see some downloading steps in [here](https://github.com/operator-framework/rukpak/actions/runs/5414554767/jobs/9841781770) for instance so we might need to add it back...

timflannagan: Thanks for the quick reply - I just wanted to double check I wasn't missing anything obvious, or any context for why it would no longer be needed 👍.

PR #615, opened 2023-04-20
version: use go1.18+ embedded build info

joelanford:
@everettraven Gave me some inspiration in operator-framework/catalogd#46 to simplify rukpak's version handling/printing.

### DIFF
--- a/cmd/rukpakctl/cmd/version.go
+++ b/cmd/rukpakctl/cmd/version.go
@@ -1,6 +1,19 @@
 /*
-Copyright © 2022 NAME HERE <EMAIL ADDRESS>
+Copyright 2023 The Operator Framework Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
 */
+
 package cmd
 
 import (
--- a/cmd/rukpakctl/cmd/version.go
+++ b/cmd/rukpakctl/cmd/version.go
@@ -1,0 +1,24 @@
+/*
+Copyright © 2022 NAME HERE <EMAIL ADDRESS>
+*/
+package cmd
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	"github.com/operator-framework/rukpak/internal/version"
+)
+
+// newVersionCmd creates the version command
+func newVersionCmd() *cobra.Command {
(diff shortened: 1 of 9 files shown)


internal/version/version.go
```diff
@@ -1,8 +1,41 @@
...
+			commitDate = setting.Value
+		case "vcs.modified":
+			switch setting.Value {
+			case "false":
+				repoState = "clean"
+			case "true":
+				repoState = "dirty"
+			default:
+				repoState = "unknown"
+			}
```
everettraven, line 38: super nit: is the switch here necessary? The go docs for the `vcs.modified` key mention it will be true or false - I'm interpreting that as the values will only ever be "true" or "false". Something a bit more simple:
```suggestion
if setting.Value != "true" {
    repoState = "clean"
} else {
    repoState = "dirty"
}
```

If we want to still use the switch, I would vote for just leaving the default case empty since the default value is already `"unknown"`.

That being said this is all just a nit about me not liking the poor readability of deep nesting and shouldn't block the PR from merging :smile:

internal/version/version.go
```diff
@@ -1,8 +1,41 @@
...
-var GitCommit string
+var (
+	gitCommit  string = "unknown"
+	commitDate string = "unknown"
+	repoState  string = "unknown"
+)
 
 func String() string {
-	return GitCommit
+	return fmt.Sprintf("revision: %q, date: %q, state: %q", gitCommit, commitDate, repoState)
```
everettraven, line 16: Nit: Would it make sense to have a Go struct represent the version info and just print the object as a string? Not sure how that string would look off the top of my head and can always be done in a follow up if we think it would look nicer.

cmd/rukpakctl/cmd/version.go
```diff
@@ -0,0 +1,24 @@
+/*
+Copyright © 2022 NAME HERE <EMAIL ADDRESS>
```
everettraven, line 2: Is there a copyright header that rukpak uses? This seems like an improper copyright header and the year is wrong

everettraven (approved):
Just some nits but I don't think they should block the PR

internal/version/version.go
joelanford: Right now, we aren't exporting anything other than the string to print. Not sure how important that is, but given the only current use is to print a string, I figured I'd preserve that. If we need a struct and the ability to get the specific fields, we could reconsider when that need arises?

cmd/rukpakctl/cmd/version.go
joelanford: Yeah, this is a copy/paste. I'll fix this one. Probably need to do a scrub for the others.

internal/version/version.go
joelanford: I guess we exported `GitCommit`, but we didn't actually use it anywhere. Either way, adding the struct seemed to require more boilerplate and lines of code, so I kinda think what's here now is fine.

internal/version/version.go
everettraven: Looks nicer for sure :)

PR #625, opened 2023-05-23
🌱 remove the upstream build tag

ncdc:
Remove the upstream build tag, as these are just defaults, and it's better to let downstream consumers override the defaults through configuration instead of compilation.


joelanford (approved):
> override the defaults through configuration instead of compilation.

In fact, I think that's what's happening downstream for rukpak anyway.

PR #626, opened 2023-05-25
[fix] Fix tag of default unpack image

varshaprasad96:
The default tag for unpack image is currently "latest". Based on the recently built images in quay:
https://quay.io/repository/operator-framework/rukpak?tab=tags it looks like the "latest" tag does not exist. Instead "main" is used. This PR fixes the same.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@
 ORG := github.com/operator-framework
 PKG := $(ORG)/rukpak
 export IMAGE_REPO ?= quay.io/operator-framework/rukpak
-export IMAGE_TAG ?= latest
+export IMAGE_TAG ?= devel
 export GO_BUILD_TAGS ?= ''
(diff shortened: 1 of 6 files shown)


varshaprasad96:
cc: @joelanford

joelanford:
I think the idea we're actually going for is "refer to a local-only development tag in all the places" because that's actually what we build and use almost exclusively when interacting with rukpak via this source code repository. We never actually push it this local image anywhere, so the intent is to avoid using a name that we push (or that sounds like something that would be pushed).

So I'd suggest:
1. Changing `latest` to `devel`
2. Except in the default unpack image, where I think it's okay to leave `main`.

For 2, I'm thinking of doing a follow-up to further change that to look for and use a git tag, and if not set, then fall back to `main`. IIRC, we do something similar in `operator-sdk` when we scaffold Ansible and Helm project Dockerfiles.

PR #629, opened 2023-06-07
Switch tools to bingo

dtfranz:
Removes the hack/tools folder in favor of using bingo.


.github/workflows/sanity.yaml
```diff
@@ -39,6 +39,7 @@ jobs:
         uses: github/super-linter/slim@v4
         env:
           VALIDATE_ALL_CODEBASE: true
+          FILTER_REGEX_EXCLUDE: .*\.bingo/.*
```
dtfranz, line 42: The generated bingo readme sets off alarms from the validator :(

ncdc (approved):

joelanford (approved):

PR #630, opened 2023-06-12
(refactor): externalize and improve the `source` package

### LINKED ISSUE #627: externalize the source/unpacker implementation
The logic under `internal/source` is something that would be of use to [operator-framework/catalogd](https://github.com/operator-framework/catalogd). Currently catalogd more or less copies the implementation from rukpak. It would be much nicer if this logic was externally available and catalogd could import and reuse this logic.

everettraven:
## Description
- Moves `internal/source` --> `pkg/source` so that the rukpak unpacking logic can be imported and used in other projects
- Updates the implementations in the `source` package to be more library user friendly

## Motivation
- resolves #627 
- Externalizing the package allows https://github.com/operator-framework/catalogd to reuse the rukpak unpacking logic rather than re-implementing it on its own.


everettraven:
I'm a bit confused by this lint error: https://github.com/operator-framework/rukpak/actions/runs/5259375486/jobs/9504861526?pr=630#step:4:32

I've run `goimports` and there were no changes to anything to push so I'm not sure how to fix this. Unfortunately I can't run the linter locally as every time I run `make lint` it ends up crashing my terminal so :shrug:

everettraven:
A good follow-up to this PR would be to add tests for the individual unpacking implementations. I left that out for this PR as that felt out of scope for externalizing this logic.

everettraven:
After working on https://github.com/operator-framework/catalogd/pull/94 there are some more changes I want to make to this. Holding and turning back to WIP

/hold

pkg/source/types.go
```diff
@@ -0,0 +1,117 @@
...
+
+const (
+	SourceTypeImage      SourceType = "image"
+	SourceTypeGit        SourceType = "git"
+	SourceTypeConfigMaps SourceType = "configMaps"
+	SourceTypeUpload     SourceType = "upload"
+	SourceTypeHTTP       SourceType = "http"
+)
+
+// TODO: Can we make source a bit more generic and composable/extensible?
```
everettraven, line 18: I was thinking something like:
```go
type Source struct {
    Type SourceType `json:"type"`
    Config json.RawMessage `json:"config"`
}
```
This would allow for different configurations to be parsed and validated by the individual unpacker types similar to how this was done for the [opm composite template in their respective builders](https://github.com/operator-framework/operator-registry/blob/6c602841934d6e154e38c0574cc140471dc063e6/alpha/template/composite/builder.go#L53-L78)

However, this ends up being a runtime validation rather than one that could occur on the CRD, *unless* we created a configurable webhook implementation that other projects could reuse for validating CRs with this source type before admitting it.

I'm wondering if the extensibility/configurability feature is valuable enough to warrant requiring users of the `source` package to implement an admission webhook if they want validation before their CR is created.

If we do decide it is worth it, I think this should be done in a follow up as there would likely be changes across the repo to support this and I feel like this PR is already relatively large

pkg/source/unpacker.go
```diff
@@ -0,0 +1,85 @@
...
+	// and processed.
+	StateUnpacking State = "Unpacking"
+
+	// StateUnpacked conveys that the source has been successfully unpacked.
+	StateUnpacked State = "Unpacked"
+)
+
+type unpacker struct {
+	sources map[SourceType]Unpacker
+}
```
joelanford, line 71: WDYT about changing this to:
```go
type unpacker struct {
	image *Image
	git *Git
	http *HTTP
	upload *Upload
	configMaps *ConfigMaps
}
```

And then implement a `Builder` pattern for the constructor, such that a caller would do:
```go
u, err := source.UnpackerBuilder()
	.WithImage(...)
	.WithGit(...)
	.WithHTTP(...)
	.WithUpload(...)
	.WithConfigMaps(...)
	.Build()
```

And `Build()` could check that all of the underlying source implementations were non-nil (or return an error). I think this would make the unpacker setup a bit more discoverable, and it would help guarantee that all users of the composite unpacker had implementations for all of the known source types.

pkg/source/unpacker.go
```diff
@@ -0,0 +1,85 @@
...
+// can be reconciled as progress updates are available.
+//
+// For asynchronous Sources, multiple calls to Unpack should be made until the
+// returned result includes state StateUnpacked.
+//
+// NOTE: A source is meant to be agnostic to specific source formats and
+// specifications. A source should treat a source root directory as an opaque
+// file tree and delegate source format concerns to source parsers.
+type Unpacker interface {
+	Unpack(context.Context, *Source, client.Object) (*Result, error)
```
joelanford, line 26: +1 moving this from `Bundle` to `client.Object`, but it'd be even better if this interface didn't depend on having a kubernetes object. It seems like the primary reason for this needing to exist is to make sure the image unpacker is able to add an owner reference to the pod that it creates.  But it looks like the other source type implementations either don't use it, or use it for what amounts to a unique identifier.

I wonder if there's something we could do to remove this from the interface while also still giving the image unpacker enough information to create a valid owner reference?

pkg/source/unpacker.go
everettraven: I'm not opposed to this, but I have a couple questions:
- How would someone add a custom source to the unpacker?
- How could someone elect to not implement one of the known source types? (this one in particular is important for catalogd - in [my PoC](https://github.com/operator-framework/catalogd/pull/94) we don't currently use the `Upload` source since it depends on having the rukpak upload service running on cluster. IMO this unpacker should easily extensible rather than locked into using specific source types)

pkg/source/unpacker.go
everettraven: Maybe the solution is adding an "optional" field to the `Source` type that is of `client.Object` (or maybe even better `metav1.Object`?)? I say "optional" because some sourcing implementations may require it while others may not.

pkg/source/unpacker.go
joelanford: Re: custom type: if/when we have that use case, we could extend the struct and builder to accept arbitrary type names and implementations.

Re: require implementations of all types. My thought here was mainly to make sure that when we add a new core type in the rukpak bundle api, all rukpak provisioners that bump to the new version of rukpak that adds the type are given a signal to pick it up. On main right now, we get a compiler error when we change the signature of NewDefaultUnpacker to accommodate a new type, which is really nice (other than the unruly growth of the signature parameters).

There are other ways to do this though. For example, we could build a conformance test that includes expectations that for each supported type. Its just that the feedback loop on that kind of thing is so much longer.

joelanford: Re: upload source, maybe we should position the upload service alongside the upload source implementation to make the upload source portable?

everettraven: >Re: custom type: if/when we have that use case, we could extend the struct and builder to accept arbitrary type names and implementations.

Fair enough. In my head I had it that there was already a use case for it, but now that you pointed it out I can't think of one...

> Re: require implementations of all types. My thought here was mainly to make sure that when we add a new core type in the rukpak bundle api, all rukpak provisioners that bump to the new version of rukpak that adds the type are given a signal to pick it up. On main right now, we get a compiler error when we change the signature of NewDefaultUnpacker to accommodate a new type, which is really nice (other than the unruly growth of the signature parameters).

This is definitely a nice to have. I'll have to circle back around to this and think about how we can do this. I'm still hesitant to require that all usages implement _every_ source type, but maybe we can instead of erroring on `Build` we return an error when that source type is attempted to be used that states something like `ABC doesn't support the XYZ source`? 

I think in this scenario you would still get the ideal scenario of requiring to add an instantiation when creating an `Unpacker` but you could explicitly set it to `nil` if you didn't want to support said source type.

> Re: upload source, maybe we should position the upload service alongside the upload source implementation to make the upload source portable?

I think that is reasonable, but I still think there are use cases where someone doesn't want to implement all the source types due to potential overhead. Probably a conversation to be had for catalogd if the additional service overhead is worth the value provided to users. On top of that, how does a user consume it? Would `rukpakctl` work for uploading something other than a bundle or would catalogd have to implement its own uploader CLI?

PR #632, opened 2023-06-13
Updates Super-Linter (it changed org)

m1kola:
* Updates to use the correct github-action reference
* Updates to use the latest version


m1kola:
Replacing it with another linter action which is much faster and provides better output (adds annotatiosn into PRs). See #633

PR #634, opened 2023-06-14
Bingo kustomize

dtfranz:
Fixes a few deprecation warnings and adds kustomize to bingo. Can't yet find a way to make string replacements in the urls though :(


.bingo/Variables.mk
```diff
@@ -53,6 +53,12 @@ $(KIND): $(BINGO_DIR)/kind.mod
 	@echo "(re)installing $(GOBIN)/kind-v0.17.0"
 	@cd $(BINGO_DIR) && GOWORK=off $(GO) build -mod=mod -modfile=kind.mod -o=$(GOBIN)/kind-v0.17.0 "sigs.k8s.io/kind"
 
+KUSTOMIZE := $(GOBIN)/kustomize-v5.0.1
```
m1kola, line 56: What version of kubectl do we use? This version of kustomize is included in 1.26

grokspawn: I don't know if we need to standardize versions, but [operator-controller](https://github.com/openshift/operator-framework-operator-controller/blob/main/.bingo/kustomize.mod) uses v4.5.7 and [catalogd](https://github.com/openshift/operator-framework-catalogd/blob/main/.bingo/kustomize.mod) uses v5.0.1

everettraven (commented):
Changes look fine to me, but my kustomize-fu isn't the strongest and I don't have as much background on the manifests for rukpak so I'll leave approval up to someone with more historical knowledge

joelanford:
@dtfranz, is there a diff in the resulting `kustomize build *` outputs after this change?

joelanford (approved):

PR #635, opened 2023-06-14
Updates golangci-lint

m1kola:
Starting version v1.51.0 `golangci-lint` comes with a new `ginkgolinter` linter which checks for common ginkgo issues like wrong error assertions (`Expect(err).To(BeNil())` instead of `Expect(err).NotTo(HaveOccurred())`), etc.

This PR does not add `ginkgolinter`: it will be a followup PR because it will further increase the diff. 

In this PR:
* golangci-lint gets updated from v1.49.0 to v1.53.2
* `depguard` linter removed. We do not have configuration for it and new version of `golangci-lint` comes with the configuration where only Go standard library is allowed.
* Changes necessary to satisfy new versions of linters (`revive` and `unparam` were complaining)


internal/helm-operator-plugins/predicate/depedent.go
```diff
@@ -61,18 +61,18 @@ func DependentPredicateFuncs() crtpredicate.Funcs {
 		// necessary. Ignore updates that only change the status and
 		// resourceVersion.
 		UpdateFunc: func(e event.UpdateEvent) bool {
-			old := e.ObjectOld.(*unstructured.Unstructured).DeepCopy()
-			new := e.ObjectNew.(*unstructured.Unstructured).DeepCopy()
```
m1kola, line 65: New was shadowing built-in `new`. Renamed `old` into `oldObj` for consistency.

internal/provisioner/helm/helm.go
```diff
@@ -21,7 +21,7 @@ const (
 	ProvisionerID = "core-rukpak-io-helm"
 )
 
-func HandleBundle(ctx context.Context, fsys fs.FS, bundle *rukpakv1alpha1.Bundle) (fs.FS, error) {
+func HandleBundle(_ context.Context, fsys fs.FS, _ *rukpakv1alpha1.Bundle) (fs.FS, error) {
```
m1kola, line 24: Here and in other places with `_` args - we don't use args in functions, but we still need args to statisfy interfaces.

internal/controllers/bundle/bundle.go
```diff
@@ -159,19 +159,19 @@ func (c *controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
 		if updateErr := c.cl.Update(ctx, reconciledBundle); updateErr != nil {
-			return res, apimacherrors.NewAggregate([]error{reconcileErr, updateErr})
+			return ctrl.Result{}, apimacherrors.NewAggregate([]error{reconcileErr, updateErr})
 		}
 	}
-	return res, reconcileErr
+	return ctrl.Result{}, reconcileErr
 }
 
-func (c *controller) reconcile(ctx context.Context, bundle *rukpakv1alpha1.Bundle) (ctrl.Result, error) {
```
m1kola, line 174: Here and in `internal/controllers/bundledeployment/bundledeployment.go` we always return an empty value in `ctrl.Result{}`. It makes sense in `Reconcile` to satisfy the interface, but not in this unexported func.

m1kola (commented):
Added comments explaining why linters were complaining.

internal/controllers/bundle/bundle.go
joelanford: I'm not sure I agree with this change. If we ever _do_ want to return something other than `ctrl.Result{}`, it'll likely be coming from `reconcile()`, not `Reconcile()`.

I'd be slightly worried that our future selves would be less inclined to change the return values back to `(ctrl.Result, error)` and would instead do something less desirable overall.

The primary purpose of this split is to make us not really have to think at all about status updating. The result is merely passed through.

Is a linter complaining about this? If so, WDYT about applying the `//nolint:<whatever>` label with an explanation along these lines.

m1kola: @joelanford I also was not super happy about this linting issue at first, but ended up obeying because I convinced myself that the code is not set in stone. It is super easy to add a new return param once we have a need for it (especially since this is not an exported method).

To me it is the same as adding input arguments into a function: we usually add new args as the need for them arises. There are exceptions (e.g. designing a public interfaces where signature changes are more painful), but it is not the case here.

But I can look into either adding nolint or disabling this linting rule altogether tomorrow. I am not a big fan of inline linting exceptions, but also do not have very strong objections against them.

joelanford (approved):
/lgtm

PR #638, opened 2023-06-15
Adds importas rules

m1kola:
To enforce common patterns such as importing `k8s.io/api/rbac/v1` as `rbacv1`.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -30,12 +30,18 @@ linters-settings:
 
   importas:
     alias:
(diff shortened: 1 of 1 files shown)


.golangci.yaml
```diff
@@ -30,12 +30,18 @@ linters-settings:
 
   importas:
     alias:
-    - pkg: k8s.io/api/core/v1
-      alias: corev1
```
m1kola, line 34: This (and many other imports) is now covered by the regex below.

PR #639, opened 2023-06-15
Removes deprecated `--progress` flag from Ginkgo

m1kola:


m1kola:
@joelanford the only difference I was able to spot is absense of the deprecation message. You can see it in previous jobs in GH actions.

before

after

PR #640, opened 2023-06-15
Adding manifests verification step for PRs

dtfranz:
Example test PR which fails this step:
https://github.com/operator-framework/rukpak/actions/runs/5284578866/jobs/9562154203?pr=641

Currently the action is not set as required so it depends on engineers paying attention to it. If the action is failing for an intentional change that should be noted in the PR's description.

### DIFF
--- a/.github/workflows/manifests.yaml
+++ b/.github/workflows/manifests.yaml
@@ -1,0 +1,34 @@
+name: manifests-diff
+
+on:
+  pull_request:
+  push:
+    branches:
+    - main
+
+jobs:
+  manifests-diff:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout target branch of repository
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.base_ref }}
+      - name: Setup Go
+        uses: actions/setup-go@v4
+        with:
+          go-version-file: "go.mod"
+      - name: Generate rukpak.yaml installation manifest from target branch
+        run: make quickstart
+      - name: Rename manifest file for later comparison
+        run: mv rukpak.yaml ${{ github.base_ref }}.yaml
+      - name: Checkout PR branch
+        uses: actions/checkout@v3
+        with:
+          clean: false # Preserve the manifest we generated earlier
+      - name: Generate rukpak.yaml installation manifest from source branch
+        run: make quickstart
+      - name: Rename manifest file
+        run: mv rukpak.yaml ${{ github.head_ref }}.yaml
+      - name: Compare generated manifests of source and target branch
+        run: diff ${{ github.base_ref }}.yaml ${{ github.head_ref }}.yaml


.github/workflows/manifests.yaml
```diff
@@ -0,0 +1,32 @@
...
+        with:
+          ref: ${{ github.base_ref }}
+      - name: Setup Go
+        uses: actions/setup-go@v4
+        with:
+          go-version-file: "go.mod"
+      - name: Generate rukpak.yaml installation manifest from target branch
+        run: make quickstart
+      - name: Rename manifest file for later comparison
+        run: mv rukpak.yaml rukpak-TODO.yaml
```
joelanford, line 24: WDYT about renaming these to help PR reviewers understand context a little better? For example `base-branch-manifest.yaml` and `pr-manifest.yaml`

joelanford (approved):

.github/workflows/manifests.yaml
```diff
@@ -0,0 +1,32 @@
...
+      - name: Checkout PR branch
+        uses: actions/checkout@v3
+        with:
+          clean: false # Preserve the manifest we generated earlier
+      - name: Generate rukpak.yaml installation manifest from source branch
+        run: make quickstart
+      - name: Compare generated manifests of source and target branch
+        run: diff rukpak.yaml rukpak-TODO.yaml
```
m1kola, line 32: I think `diff` will be unreliable as it compares line by line. So something like this will give a false positive:

```
key1: val1
key2: val2
```
vs
```
key2: val2
key1: val1
```

Files are different. But semantically - they are the same.

m1kola (commented):
> Currently the action is not set as required so it depends on engineers paying attention to it. If the action is failing for an intentional change that should be noted in the PR's description.

I'm not sure that I see the value in this workflow, but I might be missing something. Could you please explain what we gain from it?

.github/workflows/manifests.yaml
dtfranz: Agreed, I left the filename as `rukpak-TODO.yaml` and was going to have that be named after the branch.

.github/workflows/manifests.yaml
dtfranz: This is a good point - I may need to put some more work in to refine the comparison.

dtfranz:
> > Currently the action is not set as required so it depends on engineers paying attention to it. If the action is failing for an intentional change that should be noted in the PR's description.
> 
> I'm not sure that I see the value in this workflow, but I might be missing something. Could you please explain what we gain from it?

The value can be seen in a PR like #634 , where we are making changes to the manifests but don't actually want to change any of the content. It could also bring more attention to unwanted changes from a `kustomize` version bump. Similar in mission to the `go-apidiff` action.

.github/workflows/manifests.yaml
joelanford: fwiw, `encoding/json` (which I think is ultimately what is used under the hood of kustomize to encode/decode) always sorts map keys lexicographically.

Not positive, but I don't think we'd see this particular scenario. But if something moves a list item around, that would result in a non-semantic-difference diff.

.github/workflows/manifests.yaml
dtfranz: I've updated the branch so the manifest files will match the branches which generated them.

PR #644, opened 2023-06-22
Manifest rework

dtfranz:
Splits up the Rukpak manifests into two folders: `base/` and `overlays/`. As the names suggest, this allows us to apply different overlays on top of the base manifests, in this case specifically to allow us to more easily develop using alternative certificate management solutions. 

Changes from this PR result in an identical installation file generated from the `main` branch when using `make quickstart`.


ncdc:
LGTM. Could you please add a doc comment somewhere indicating that manifests/certmanager is a cert-manager specific

joelanford:
> manifests/certmanager is a cert-manager specific set of overlays on top of base, or something to that effect?

Also, `manifests/base` is inoperable on its own, right?

dtfranz:
> LGTM. Could you please add a doc comment somewhere indicating that manifests/certmanager is a cert-manager specific set of overlays on top of base, or something to that effect?

Yes, can do!

> Also, `manifests/base` is inoperable on its own, right?

Yes, that's right, since the webhooks need the certs to function.

manifests/base/provisioners/helm/resources/deployment.yaml
```diff
@@ -62,22 +57,9 @@ spec:
             - "--storage-dir=/var/cache/bundles"
             - "--http-bind-address=127.0.0.1:8080"
             - "--http-external-address=https://$(HELM_PROVISIONER_SERVICE_NAME).$(HELM_PROVISIONER_SERVICE_NAMESPACE).svc"
-            - "--bundle-ca-file=/etc/pki/tls/ca.crt"
           ports:
             - containerPort: 8080
-          volumeMounts:
-            - name: bundle-cache
-              mountPath: /var/cache/bundles
```
joelanford, line 70: Is it not possible to keep the unrelated volumeMounts and volumes in the base?

dtfranz: Yeah we can definitely leave them in. I'll move these back; I did this to maintain the list order after strategic merge but there are other ways to do that.

joelanford:
Nice work!

/approve

ncdc (approved):

PR #648, opened 2023-06-28
Bingo version updates

dtfranz:
Updates all bingo tools to use go 1.20. ~~Also uses yq for image tag replacements rather than doing sed replacement.~~


ncdc:
@dtfranz the description mentions image tag replacements via yq but I don't see that in the diff?

dtfranz:
> @dtfranz the description mentions image tag replacements via yq but I don't see that in the diff?

Yes, I was initially using it to set the image tags but I noticed when doing that it resulted in huge diffs in the CRDs. I figured I could take care of that later once we have the higher priority stuff done.

PR #649, opened 2023-06-30
Force install namespace for "other" objects

thetechnick:
OLMv0 forces all objects in the bundle image to be created in the install namespace even if left blank or filled with a different namespace.

This commit forces all "other" objects to also have their namespace populated via the converter, in the same way one would expect from OLMv0.

This behaviour was tested using the cert-manager operator hub bundle in version v1.12.2, where a Service without explicit namespace is part of the bundle image.

### DIFF
--- a/internal/convert/registryv1.go
+++ b/internal/convert/registryv1.go
@@ -321,6 +321,8 @@ func Convert(in RegistryV1, installNamespace string, targetNamespaces []string) 
 	}
 	for _, obj := range in.Others {
 		obj := obj
+		// Default the namespace of other bundled objects to the installNamespace, like in OLMv0.
+		obj.SetNamespace(installNamespace)
 		objs = append(objs, &obj)
 	}
 	for _, obj := range deployments {


joelanford:
Thanks @thetechnick!

I think this opens up an interesting discussion point about how OLMv1 handles registry+v1 bundles.

On one hand, we could try to replicate OLMv0 functionality as closely as possible in terms of the way that registry+v1 bundle manifests are transformed into on-cluster resources.

On the other, because we're using different APIs, we _could_ make rukpak suddenly support new/different functionality from the same bundle format. Without this change, certain bundles might end up being non-functional because we're not blindly overwriting the install namespace. On the other hand, blindly overwriting the install namespace that was explicitly defined in the bundle manifest seems extremely confusing from a UX standpoint.

My general preference with the new OLMv1 stuff is to get rid of the "magic" and make things much more predictable and perhaps strict.

Perhaps there is some middle ground here? If we don't want to suddenly allow bundles to install objects outside of the install namespace, WDYT about:
- Allowing a manifest to leave the namespace unset, in which case we'll set it to the install namespace (I think this already happens, but not positive)
- If the manifest namespace is set, check that it matches the install namespace. If it does not, **error out instead of silently overwriting it**.

thetechnick:
Hey @joelanford, I was out on PTO for a bit and only now catch up.
Let me close this PR and reopen a ticket about discussing ruckpak <> OLMv0 compatibility  in general.

PR #658, opened 2023-07-25
Add initial Tilt support

ncdc:
TODOs (either for this PR or as follow-ups)
- [ ] Add delve support
- [ ] See if we can add support for having Tilt build the unpack image and use it when running via Tilt

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,5 @@ dist/
 
 # Release artifacts
 rukpak.yaml
+
+.tiltbuild/
--- a/Tiltfile
+++ b/Tiltfile
@@ -1,0 +1,75 @@
+# This loads a helper function that isn't part of core Tilt that simplifies restarting the process in the container
+# when files changes.
+load('ext://restart_process', 'docker_build_with_restart')
+
+# These are the 4 binaries that make up the 4 deployments for rukpak.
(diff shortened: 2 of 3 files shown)


everettraven (approved):
Thanks @ncdc - this is super neat! I'm happy for this to go in as is and iterate to make improvements to the Tiltfile and the use of tilt as we go

ncdc:
Switched to draft mode for a bit. I forgot to add docs about a cert-manager requirement. Working on that.

ncdc:
Added

.gitignore
```diff
@@ -18,3 +18,5 @@ dist/
 
 # Release artifacts
 rukpak.yaml
+
+.tiltbuild
```
tmshort, line 22:
```suggestion
.tiltbuild/
```

tmshort (commented):
I also have similar comments to those I made in `operator-controller`

ncdc:
Thanks, updated

tmshort (approved):
/lgtm

tmshort (changes requested):
Do you need `.dockerignore`?

ncdc:
No, because the context used to build the images is bin/linux

tmshort (approved):

PR #665, opened 2023-07-31
Externalize storage package

### LINKED ISSUE #656: externalize the `storage` package
[catalogd](https://github.com/operator-framework/catalogd) is looking to implement a format similar to RukPak's internal `storage` library for storing catalog content via a filesystem and serving it over HTTP (for more information see the [RFC](https://docs.google.com/document/d/1exozVIb1dOQ08up5iXz4D6MAtzSyE3W8rV5_FKmds5I/edit?usp=sharing)). We would like to import this library and reuse existing code rather than reinventing the wheel for catalogd.

oceanc80:
Externalize storage package for use by catalogd.

Closes #656


everettraven (approved):
I think this is a good starting point and we can iteratively make improvements to the package as it begins being used by catalogd. That being said, I can see a case for only externalizing the interfaces in storage.go since everything else is pretty tailored around Rukpak and the notion of using Bundles for things. IMO this isn't a big deal as we can iterate on this to make it more generic as the need arises.

ncdc (approved):

PR #666, opened 2023-08-01
Fix: Overriding user defined labels

varshaprasad96:
This PR fixes an issue where helm's postrenderer
used to override the user defined object labels instead of merging them.

Issue:
Previously, if any manifest in the bundle had labels provided in them, they were overwritten when bundle deployment was created.

### DIFF
--- a/internal/controllers/bundledeployment/bundledeployment_test.go
+++ b/internal/controllers/bundledeployment/bundledeployment_test.go
@@ -1,0 +1,120 @@
+package bundledeployment
+
+import (
+	"bytes"
+	"encoding/json"
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"helm.sh/helm/v3/pkg/postrender"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
+	rukpakv1alpha1 "github.com/operator-framework/rukpak/api/v1alpha1"
+	"github.com/operator-framework/rukpak/internal/util"
+)
+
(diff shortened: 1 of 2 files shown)


varshaprasad96:
> Could you please add a test that fails without the fix, and passes with it?

Done. This package did not have tests for bundleDeployment controller, I've added one just to cover post renderer for now.  
Will create issue to add more add coverage to the controller.

internal/controllers/bundledeployment/bundledeployment_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+			labels := renderedPod.GetLabels()
+			Expect(len(labels)).To(BeEquivalentTo(3))
+			Expect(labels).Should(HaveKeyWithValue("testKey", "testValue"))
+			Expect(labels).Should(HaveKeyWithValue("core.rukpak.io/owner-kind", "BundleDeployment"))
+			Expect(labels).Should(HaveKeyWithValue("core.rukpak.io/owner-name", "test-owner"))
+		})
+
+		It("should add default labels to the pod", func() {
+			pod.SetLabels(map[string]string{})
```
ncdc, line 74:
```suggestion
			pod.SetLabels(nil)
```

is probably a more representative test. Or you could add another test case so we have 1 for nil, 1 for empty map.

But "should add default labels" is already covered by the previous test case, FYI.

ncdc (approved):

internal/controllers/bundledeployment/bundledeployment_test.go
varshaprasad96: I've rearranged the tests. Let me know if this looks better

varshaprasad96:
Looks like `manifests-diff` maybe failing because my branch name as it has a slash, where its trying to copy `rukpak.yaml` and is not able to find the path. Will close this and re-open the PR with a different branch name :/

PR #675, opened 2023-08-03
Add CI for tilt

ncdc:


.github/workflows/tilt.yaml
```diff
@@ -0,0 +1,40 @@
+on:
+  pull_request:
+    paths:
```
m1kola, line 3: Same [question](https://github.com/operator-framework/operator-controller/pull/319#discussion_r1284155257) as in operator-controller: Maybe `paths-ignore` is more suitable here?

m1kola (approved):

ncdc:
Proceeding to merge given the approval. Can adjust the paths config if needed in a follow up.

PR #678, opened 2023-08-08
fix: Adding namespace when converting bundle from registryv1 to plain

varshaprasad96:
Currently, if the objects being passed through the bundle do not have their namespaces set, they were being created on the "rukpak-system" namespace. This PR fixes this issue for only "registryV1" bundles. It sets the namespace to installNs if not provided by default.


awgreene:
Note: this PR updates the registry v1 provisioner to mirror OLM V0 behavior, in which the namespace is set to the installNamespace if the resource does not specify a namespace.
/approve

joelanford:
This makes sense to me because its an implicit feature of the registry+v1 bundle format that objects without `metadata.namespace` defined will be created in the install namespace.

joelanford:
@awgreene, if a bundle object "A" hardcodes its namespace to "foo" and then someone creates a subscription resulting in the installation of that bundle in namespace "bar", where does "A" end up? "foo" or "bar"?

awgreene:
@joelanford my impression was that the namespace is _always_ set to the installNamespace. Let me double check.

awgreene:
@joelanford based on the ensure methods found in the [bundle_unpacker](https://github.com/operator-framework/operator-lifecycle-manager/blob/2be5e58b153eea608b6da16a8e3a1c46bc6c5b87/pkg/controller/bundle/bundle_unpacker.go) code, such as [ensureRoleBidning](https://github.com/operator-framework/operator-lifecycle-manager/blob/2be5e58b153eea608b6da16a8e3a1c46bc6c5b87/pkg/controller/bundle/bundle_unpacker.go#L730) , the namespace is always overwritten by the installPlan namespace.

awgreene:
This is interesting @joelanford. Based on the ensure methods defined in [step_ensurer.go](https://github.com/operator-framework/operator-lifecycle-manager/blob/2be5e58b153eea608b6da16a8e3a1c46bc6c5b87/pkg/controller/operators/catalog/step_ensurer.go), some objects end up in `foo` and some end up in `bar`:
- [EnsureServiceAccount](https://github.com/operator-framework/operator-lifecycle-manager/blob/2be5e58b153eea608b6da16a8e3a1c46bc6c5b87/pkg/controller/operators/catalog/step_ensurer.go#L138) creates the object in `bar`. This applies for services, serviceAccounts, CSVs, subscriptions, secrets, services, roles, roleBindings
- [EnsureUnstructuredObject](https://github.com/operator-framework/operator-lifecycle-manager/blob/2be5e58b153eea608b6da16a8e3a1c46bc6c5b87/pkg/controller/operators/catalog/step_ensurer.go#L300) creates the object in `foo`. This applies to objects not defined in the previous bullet but allowed in bundles.

varshaprasad96:
@awgreene can correct me, but looks like the namespace of objects being installed is being [overwritten](https://github.com/operator-framework/operator-lifecycle-manager/blob/2be5e58b153eea608b6da16a8e3a1c46bc6c5b87/pkg/controller/operators/catalog/operator.go#L2385) by [IP](https://github.com/operator-framework/operator-lifecycle-manager/blob/2be5e58b153eea608b6da16a8e3a1c46bc6c5b87/pkg/controller/operators/catalog/operator.go#L1962)

awgreene:
NVM, @varshaprasad96 pointed out that the namespace for unstructured objects are set [here](https://github.com/operator-framework/operator-lifecycle-manager/blob/2be5e58b153eea608b6da16a8e3a1c46bc6c5b87/pkg/controller/operators/catalog/operator.go#L2385). All namespaced resources are created in the installPlan namespace.

internal/convert/registryv1.go
```diff
@@ -321,6 +321,9 @@ func Convert(in RegistryV1, installNamespace string, targetNamespaces []string)
 	}
 	for _, obj := range in.Others {
 		obj := obj
+		if obj.GetNamespace() == "" {
+			obj.SetNamespace(installNamespace)
+		}
```
joelanford, line 326: Based on [this](https://github.com/operator-framework/rukpak/pull/678#issuecomment-1670363234) comment, should we just forcibly set namespace to install namespace?

```suggestion
		obj.SetNamespace(installNamespace)
```

internal/convert/registryv1.go
```diff
@@ -321,6 +321,9 @@ func Convert(in RegistryV1, installNamespace string, targetNamespaces []string)
 	}
 	for _, obj := range in.Others {
 		obj := obj
```
joelanford, line 323: Should we also do the `IsSupported` check?

https://github.com/operator-framework/operator-registry/blob/87c0b28d82eb882c344ba2c7efc696063c227a67/pkg/lib/bundle/supported_resources.go#L53

varshaprasad96: Do we want to? One of the benefit of v1 was to not restrict users with the specific kind of GVKs being supported

internal/convert/registryv1.go
varshaprasad96: If we do so, the only way we allow authors to set namespaces would be through namespace annotation in CSV (as of now). Also, this would also mean that all bundle objects would be deployed in the same NS, different values cannot be configured. I'm fine with it, just want to make sure that we are intentionally making this design decision, and not doing it because of v0, since that behaviour can still be reproduced.

joelanford:
@varshaprasad96 for both of your questions, for better or worse, I think we should make registry+v1 bundles in OLMv1 behave as closely as possible to OLMv0. Reasons being:
1. There are years of expectations, assumptions, and muscle memory built up around the _behavior_ of registry+v1 + OLMv0, to the point that registry+v1 is pretty much synonymous with OLMv0.
2. Any deviation from OLMv0 behavior with respect to registry+v1 bundles has the potential to cause a breaking change in migration scenarios.
3. Any new features or improvements we put in registry+v1 is one less reason for people to migrate to a different format. If we think registry+v1 is not what we want folks using long-term, it is perhaps better to avoid adding features to it that make it more appealing.

ncdc:
@joelanford does your answer change at all (maybe only from an implementation perspective) if/when we switch away from provisioners that are tied to bundle format? i.e. unpack-process-apply?

joelanford:
I was thinking about that last night. The idea that came to mind first was to expose something like `renderRegistryV1` as a `processor` type, which would essentially call this same `Convert` function.

We probably have to stew on it more, but my instinct is that we should:
1. stick to our registry+v1 bundle validations.
2. make sure to support a simple unpack-process-apply setup that implements OLMv0 behavior to make migrations easy
3. allow BD clients to do more complex post-processing of registry+v1 bundles using more complex unpack-process-apply setups.

joelanford (approved):

Issue #667, opened 2023-08-01
manifests-diff CI workflow fails if PR branch has a / in its name


ncdc:
See e.g. https://github.com/operator-framework/rukpak/actions/runs/5729666108/job/15526872639?pr=666

```
mv: cannot move 'rukpak.yaml' to 'sre/cloud-ingress.yaml': No such file or directory
```

Branch name is `sre/cloud-ingress.yaml`

joelanford:
This is also a problem for dependabot PRs.

ncdc:
I see this was added in #640. I'd like to propose we remove this workflow instead of fixing it.

ncdc:
@dtfranz

dtfranz:
Agreed, created #679

PR #679, opened 2023-08-08
Remove Manifests Workflow

dtfranz:
The workflow is causing more trouble than it's worth at the moment, particularly for dependabot.

Addresses #667


joelanford:
Maybe the problem is solved if we name them hardcoded to /tmp/base.yaml and /tmp/head.yaml?

That avoids any possibility of filename overlap in our repo and the issue with slashes in branch names being interpreted as directories, which didn't exist.

Issue #704, opened 2023-08-28
BundleDeployment Health: Extend BundleDeployment API to support a new healthy condition


ankitathomas:
Introduce a new Health condition for the BundleDeployment CR with an enumerated set of valid states/reasons:

-   ObjectLookupFailure
-   InstallationStatusUnknown
-   Healthy
-   Unhealthy
-  InstallationStatusFalse

PR #711, opened 2023-08-29
Add new "Healthy" type condition and Reasons

jmprusi:
Adds a "built-in" health check function under a feature gate.

It returns true if all resources are healthy with nil error.
It returns false if any of the resources are not healthy; the error contains the GVK + resource name and the error message of
each unhealthy resource.

The current list of supported resources is:
- Deployments
- StatefulSets
- DaemonSets
- ReplicaSets
- Pods
- APIServices
- CustomResourceDefinitions

If the resource is not supported, it is assumed to be healthy.

closes https://github.com/operator-framework/rukpak/issues/704
closes https://github.com/operator-framework/rukpak/issues/705
closes https://github.com/operator-framework/rukpak/issues/706


everettraven (dismissed):

stevekuznetsov:
Was this PR supposed to include the implementation as well?

everettraven:
@stevekuznetsov My understanding from looking at the issues was that https://github.com/operator-framework/rukpak/issues/706 would be for the implementation

stevekuznetsov:
Seems weird but ok

ncdc:
I would prefer that API changes are merged in a single PR with their implementation. Would you be able to do that for this?

jmprusi:
> I would prefer that API changes are merged in a single PR with their implementation. Would you be able to do that for this?

Totally! Makes more sense to me.

internal/healthchecks/generic/generic.go
```diff
@@ -0,0 +1,198 @@
...
+// - StatefulSets
+// - DaemonSets
+// - ReplicaSets
+// - Pods
+// - APIServices
+// - CustomResourceDefinitions
+//
+// If the resource is not supported, it is assumed to be healthy.
+func AreObjectsHealthy(ctx context.Context, client client.Client, objects []client.Object) (bool, error) {
+	gvkErrorsMap := make(map[string]string)
```
joelanford, line 34: Do we need a map? I wonder if we can just have a `[]error` and use `errors.Join()` when we return it.

And if the order of the `objects` slice is deterministic, then the order of the returned error string will also be deterministic. Is that indirect determinism sufficient?

internal/healthchecks/generic/generic.go
```diff
@@ -0,0 +1,198 @@
...
+
+	// Create a new scheme to be able to convert the objects to their correct version.
+	s := runtime.NewScheme()
+	apiregistrationv1.AddToScheme(s)
+	apiextensionsv1.AddToScheme(s)
+	appsv1.AddToScheme(s)
+	corev1.AddToScheme(s)
+
+	for _, object := range objects {
```
stevekuznetsov, line 43: With the default controller-runtime client logic, this method will start a new cluster-wide informer for every instance of the GVR you're issuing a `Get()` for here. Is that what you want? Are these objects expected to be cached already?

internal/healthchecks/generic/generic.go
```diff
@@ -0,0 +1,198 @@
+package generic
```
stevekuznetsov, line 1: Not sure `generic` is super useful as a package name, especially when Go has generics and this code just does a `switch` on GVR.

internal/healthchecks/generic/generic.go
joelanford: Right now, rukpak is:
- setting a rukpak-wide label to every object in the manifest it applies
- starting an informer (with the rukpak-wide label selector) for each GVR it encounters in the manifest it applies

Therefore (for now at least), all of these requests will hit the cache.

internal/healthchecks/generic/generic.go
```diff
@@ -0,0 +1,198 @@
...
+	if len(gvkErrorsMap) == 0 {
+		return true, nil
+	}
+	return false, gvkErrorsMapToErr(gvkErrorsMap)
+}
+
+// toErrKey returns a string that uniquely identifies a resource based on its GVK and name.
+// This is used to create a map of "GVK + Resource Name" to error message.
+func toErrKey(resource client.Object) string {
+	return fmt.Sprintf("%s.%s.%s/%s", resource.GetObjectKind().GroupVersionKind().Kind, resource.GetObjectKind().GroupVersionKind().Group, resource.GetObjectKind().GroupVersionKind().Version, resource.GetName())
```
joelanford, line 175: This key needs to include a namespace to cover the case where a manifest has two or more objects with the same GVK and name, but in different namespaces.

internal/healthchecks/generic/generic.go
joelanford: maybe `builtin`?

I assume the future-thinking bit of this package name is that there will eventually be some extension that enables users to provide custom health check logic, so we're reserving space for that under the healthchecks directory?

internal/controllers/bundledeployment/bundledeployment.go
```diff
@@ -408,6 +457,22 @@ func (c *controller) reconcile(ctx context.Context, bd *rukpakv1alpha1.BundleDep
...
+			Reason:  rukpakv1alpha1.ReasonUnhealthy,
+			Message: err.Error(),
+		})
+		return ctrl.Result{}, err
+	}
+	meta.SetStatusCondition(&bd.Status.Conditions, metav1.Condition{
+		Type:    rukpakv1alpha1.TypeHealthy,
+		Status:  metav1.ConditionTrue,
+		Reason:  rukpakv1alpha1.ReasonHealthy,
+		Message: fmt.Sprintf("bundleDeployment is healthy"),
```
ncdc, line 473:
```suggestion
		Message: "BundleDeployment is healthy",
```

internal/healthchecks/generic/generic.go
```diff
@@ -0,0 +1,198 @@
...
+// - Deployments
+// - StatefulSets
+// - DaemonSets
+// - ReplicaSets
+// - Pods
+// - APIServices
+// - CustomResourceDefinitions
+//
+// If the resource is not supported, it is assumed to be healthy.
+func AreObjectsHealthy(ctx context.Context, client client.Client, objects []client.Object) (bool, error) {
```
ncdc, line 33: Given that 0 errors = healthy, and >= 1 error = unhealthy, I would eliminate the bool return

internal/healthchecks/generic/generic.go
ncdc: +1 to slice.

internal/healthchecks/generic/generic.go
```diff
@@ -0,0 +1,198 @@
...
+				gvkErrorsMap[toErrKey(deployment)] = err.Error()
+				continue
+			}
+
+			if deployment.Status.Conditions == nil {
+				gvkErrorsMap[toErrKey(deployment)] = "Deployment has no conditions"
+				continue
+			}
+
+			for _, condition := range deployment.Status.Conditions {
```
ncdc, line 67: Global comment: I would modify these condition checks to explicitly look for the expected condition and its expected status. For example, in this check, if there is no DeploymentAvailable condition, it is still considered healthy.

internal/healthchecks/generic/generic.go
stevekuznetsov: Write for the code you have, not the code you wish you might have in the future :)

joelanford: Yeah, this is all in an internal directory, and we're in `v0`, so double-reason not to worry about future package name changes and moving of stuff around.

I'm also fine if we just put this directly in a `healthchecks` or similar package.

internal/controllers/bundledeployment/bundledeployment.go
```diff
@@ -269,6 +270,12 @@ func (c *controller) reconcile(ctx context.Context, bd *rukpakv1alpha1.BundleDep
 			Reason:  rukpakv1alpha1.ReasonErrorGettingClient,
 			Message: err.Error(),
 		})
+		meta.SetStatusCondition(&bd.Status.Conditions, metav1.Condition{
```
stevekuznetsov, line 273: Copying this over from Slack:

Please add a singular condition - ObjectsHealthy - and the the conditions utilities to roll multiple other conditions up into a Ready or Healthy meta-condition. Having to manually set this every other time any other condition is set is fragile and error-prone.

joelanford: > and the the conditions utilities to roll multiple other conditions up into a Ready or Healthy meta-condition.

I don't disagree with the premise of having a rollup condition, but I feel like it might be out of scope for this PR.

> Having to manually set this every other time any other condition is set is fragile and error-prone.

+1 

Is the following a reasonable compromise?
- Make a function along the lines of `setInstalledAndHealthyFalse()` with whatever parameters are necessary, and call that instead of duplicating all of these lines.
- Make a follow-on issue to do a rollup, and/or figure out a way for Healthy to be fully orthogonal from Installed. (right now, and according to the RFC, if Installed=False, then Healthy must =False)

internal/healthchecks/generic/generic.go
jmprusi: done

internal/controllers/bundledeployment/bundledeployment.go
jmprusi: done

internal/healthchecks/generic/generic.go
jmprusi: done

internal/healthchecks/generic/generic.go
jmprusi: done

internal/healthchecks/generic/generic.go
jmprusi: done, moved into the healthchecks pakage and renamed to "builtin"

internal/healthchecks/generic/generic.go
jmprusi: done, move it to slice

internal/healthchecks/builtin.go
```diff
@@ -0,0 +1,216 @@
...
+			conditionExists := false
+			for _, condition := range obj.Status.Conditions {
+				if condition.Type == appsv1.DeploymentAvailable {
+					if condition.Status != "True" {
+						gvkErrors = appendResourceError(gvkErrors, obj, condition.Message)
+					}
+					conditionExists = true
+					break
+				}
+			}
```
joelanford, line 80: Can we use this helper function? https://github.com/kubernetes/apimachinery/blob/82362bc34ae0cf9ca8f904f7968b6bb257b6fb3d/pkg/api/meta/conditions.go#L74

internal/healthchecks/builtin.go
```diff
@@ -0,0 +1,216 @@
...
+			}, obj); err != nil {
+				gvkErrors = appendResourceError(gvkErrors, obj, err.Error())
+				continue
+			}
+			if obj.Status.Conditions == nil {
+				gvkErrors = appendResourceError(gvkErrors, obj, "Pod has no conditions")
+				continue
+			}
+			if obj.Status.Phase != corev1.PodRunning && obj.Status.Phase != corev1.PodSucceeded {
+				gvkErrors = appendResourceError(gvkErrors, obj, "Pod has no conditions")
```
joelanford, line 131: Copy/pasta typo? Seems like the error message should be different here.

internal/healthchecks/builtin.go
```diff
@@ -0,0 +1,216 @@
...
+				Name:      obj.GetName(),
+				Namespace: obj.GetNamespace(),
+			}, obj); err != nil {
+				gvkErrors = appendResourceError(gvkErrors, obj, err.Error())
+				continue
+			}
+			if obj.Status.Conditions == nil {
+				gvkErrors = appendResourceError(gvkErrors, obj, "Pod has no conditions")
+				continue
+			}
```
joelanford, line 129: Global comment: Do we need to check for this case of "no conditions" specifically? It seems like if we care about one or more specific conditions, we could check specifically for each of those conditions and call it a day?

internal/healthchecks/builtin.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+
+	for _, object := range objects {
+		// Convert the object to its proper type, deepCopy it to avoid mutating the original object.
+		obj, err := runtime.ObjectConvertor(s).ConvertToVersion(object.DeepCopyObject(), object.GetObjectKind().GroupVersionKind().GroupVersion())
+		if err != nil {
+			gvkErrors = appendResourceError(gvkErrors, object, err.Error())
+			continue
+		}
+		switch obj := obj.(type) {
```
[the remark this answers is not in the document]
jmprusi: before the type casting I don't get access to the `GetName()` `GetNamespace()` methods.

internal/healthchecks/builtin.go
```diff
@@ -0,0 +1,202 @@
...
+			if err := client.Get(ctx, objectKey, obj); err != nil {
+				gvkErrors = appendResourceError(gvkErrors, obj, err.Error())
+				continue
+			}
+			if obj.Status.Conditions == nil {
+				gvkErrors = appendResourceError(gvkErrors, obj, "Pod has no conditions")
+				continue
+			}
+			if obj.Status.Phase != corev1.PodRunning && obj.Status.Phase != corev1.PodSucceeded {
+				gvkErrors = appendResourceError(gvkErrors, obj, "Pod is not Running or Succeded")
```
joelanford, line 122:
```suggestion
				gvkErrors = appendResourceError(gvkErrors, obj, "Pod is not Running or Succeeded")
```

internal/healthchecks/builtin.go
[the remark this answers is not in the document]
jmprusi: done

internal/healthchecks/builtin.go
```diff
@@ -0,0 +1,216 @@
...
+// - Pods
+// - APIServices
+// - CustomResourceDefinitions
+//
+// If the resource is not supported, it is assumed to be healthy.
+func AreObjectsHealthy(ctx context.Context, client client.Client, objects []client.Object) error {
+	var gvkErrors []error
+
+	// Create a new scheme to be able to convert the objects to their correct version.
+	s := runtime.NewScheme()
```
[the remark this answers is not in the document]
jmprusi: done

internal/healthchecks/builtin.go
jmprusi: done

internal/healthchecks/builtin.go
```diff
@@ -0,0 +1,202 @@
...
+// - APIServices
+// - CustomResourceDefinitions
+//
+// If the resource is not supported, it is assumed to be healthy.
+func AreObjectsHealthy(ctx context.Context, client client.Client, objects []client.Object) error {
+	var gvkErrors []error
+
+	// Create a new scheme to be able to convert the objects to their correct version.
+	s := client.Scheme()
+	if err := apiregistrationv1.AddToScheme(s); err != nil {
```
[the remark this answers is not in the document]
jmprusi: done

internal/healthchecks/builtin.go
jmprusi: done

internal/healthchecks/builtin.go
joelanford: Ah, right. I forgot that there are some core types that don't use the conventional Condition type.

internal/healthchecks/builtin.go
```diff
@@ -0,0 +1,170 @@
...
+			continue
+		}
+
+		objectKey := types.NamespacedName{
+			Name:      object.GetName(),
+			Namespace: object.GetNamespace(),
+		}
+
+		switch obj := obj.(type) {
+		case *appsv1.Deployment:
```
joelanford, line 48: Global comment: Helm's similar case statements include multiple versions of the same GroupKind. If I'm understanding this correctly, if a bundle has an `apps/v1beta1` Deployment in it, this code will always report back "healthy" for that object.

https://github.com/helm/helm/blob/e7bb860d9a32e8739c944b8e7b7f7031d752411a/pkg/kube/ready.go#L115

internal/healthchecks/builtin.go
```diff
@@ -0,0 +1,170 @@
...
+			if conditionExists {
+				continue
+			}
+			gvkErrors = appendResourceError(gvkErrors, obj, "DeploymentAvailable condition not found")
+		case *appsv1.StatefulSet:
+			if err := client.Get(ctx, objectKey, obj); err != nil {
+				gvkErrors = appendResourceError(gvkErrors, obj, err.Error())
+				continue
+			}
+			if obj.Status.ReadyReplicas != obj.Status.Replicas {
```
joelanford, line 73: Global comment: It seems like Helm's ready checking logic is much more [thorough](https://github.com/helm/helm/blob/e7bb860d9a32e8739c944b8e7b7f7031d752411a/pkg/kube/ready.go#L357-L410). It seems like our code could return false positives and flap in a variety of common situations. Should we re-use helm's code somehow?

jmprusi: I will expand the logic.

internal/healthchecks/builtin.go
jmprusi: yes, we are checking for the v1 version only, I can add logic for the `apps/v1beta1`. Any other you can think of?

internal/healthchecks/builtin.go
jmprusi: done, please take a look.

internal/healthchecks/builtin.go
```diff
@@ -0,0 +1,200 @@
...
+func toErrKey(resource client.Object) string {
+	group := resource.GetObjectKind().GroupVersionKind().Group
+	namespace := resource.GetNamespace()
+	if group == "" {
+		group = "core"
+	}
+
+	// If the resource is namespaced, include the namespace in the key.
+	if namespace != "" {
+		return fmt.Sprintf("(%s.%s.%s)(%s/%s)", resource.GetObjectKind().GroupVersionKind().Kind, group, resource.GetObjectKind().GroupVersionKind().Version, namespace, resource.GetName())
```
everettraven, line 191: Is there a particular reason we are building our own string format for conveying the GVK? It looks like the [`schema.GroupVersionKind` type returned by `resource.GetObjectKind().GroupVersionKind()` has a `String()` function that outputs the GVK in the format `{group}/{version}, Kind={kind}`.](https://github.com/kubernetes/apimachinery/blob/bc548d1d2406dffb44e1f6ea7ec1f4e9744cf170/pkg/runtime/schema/group_version.go#L161-L163)  I wonder if it might be best to reuse that format since that is in the apimachinery package and is a format that is used by Kubernetes (and I'm certain I've seen in other places). This seems like it might be a format users familiar with Kubernetes might be more used to and inherently easier to parse than an arbitrary format specified only by rukpak.

pkg/features/features.go
```diff
@@ -0,0 +1,26 @@
...
+	// Ex: SomeFeature featuregate.Feature = "SomeFeature"
+
+	BundleDeploymentHealth featuregate.Feature = "BundleDeploymentHealth"
+)
+
+var rukpakFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
+	// Add new feature gate definitions
+	// Ex: SomeFeature: {...}
+
+	BundleDeploymentHealth: {Default: false, PreRelease: featuregate.Beta},
```
everettraven, line 19: Not sure the implications here, but I would have assumed this would be introduced as an alpha feature gate before beta (although maybe I have the meanings flipped?)

internal/healthchecks/builtin.go
```diff
@@ -0,0 +1,200 @@
...
+					if condition.Status != "True" {
+						gvkErrors = appendResourceError(gvkErrors, obj, condition.Message)
+					}
+					conditionExists = true
+					break
+				}
+			}
+			if conditionExists {
+				continue
+			}
```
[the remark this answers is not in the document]
joelanford: I already made this suggestion earlier and it isn't doable because `Deployment`s and other core types sometimes use custom `Condition` structs.

internal/healthchecks/builtin.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	return errors.Join(gvkErrors...)
+}
+
+// toErrKey returns a string that identifies a resource based on its GVK and namespace/name. This key is used
+// to identify the resource in the error message.
+func toErrKey(resource client.Object) string {
+	group := resource.GetObjectKind().GroupVersionKind().Group
+	namespace := resource.GetNamespace()
```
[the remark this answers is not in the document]
joelanford: In theory, since `resource` is an interface, the underlying implementation is unknown by the caller, so it might make sense to use the `GetNamespace` method just once and reuse the returned string everywhere.

In practice, it probably doesn't make much difference.

pkg/features/features.go
jmprusi: I will change it to alpha.

internal/healthchecks/builtin.go
jmprusi: I applied your proposed changes, now the format is: "(apps/v1, Kind=DaemonSet)(MyDaemonSet): DaemonSet is not ready"

internal/healthchecks/builtin.go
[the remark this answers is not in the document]
jmprusi: simplified, done

pkg/features/features.go
jmprusi: done

internal/healthchecks/builtin.go
jmprusi: done

internal/healthchecks/builtin.go
jmprusi: @ncdc I need help finding a way to satisfy the interface for client.Get(), can you help me understand how unstructured can be used here?

internal/healthchecks/builtin.go
```diff
@@ -0,0 +1,194 @@
...
+// - Pods
+// - APIServices
+// - CustomResourceDefinitions
+//
+// If the resource is not supported, it is assumed to be healthy.
+func AreObjectsHealthy(ctx context.Context, client client.Client, objects []client.Object) error {
+	var gvkErrors []error
+
+	for _, object := range objects {
+		// Convert the object to its proper type, deepCopy it to avoid mutating the original object.
```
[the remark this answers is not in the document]
jmprusi: client.Get() modifies the obj.

internal/healthchecks/builtin.go
ncdc: Try this
```
u := &unstructured.Unstructured{}
		u.SetGroupVersionKind(object.GetObjectKind().GroupVersionKind())
		if err := client.Get(ctx, objectKey, u); err != nil {
			
		}
```

internal/healthchecks/builtin.go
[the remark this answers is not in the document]
ncdc: Ok, I see the flow now. But if you switch to using Unstructured, you won't need this.

internal/healthchecks/builtin.go
jmprusi: Thanks @ncdc, PTAL

internal/healthchecks/builtin.go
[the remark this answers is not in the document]
jmprusi: fixed, PTAL

internal/healthchecks/builtin.go
```diff
@@ -0,0 +1,207 @@
...
+			}
+			// This logic has been adapted from the helm codebase.
+			// - https://github.com/helm/helm/blob/e7bb860d9a32e8739c944b8e7b7f7031d752411a/pkg/kube/ready.go#L357-L410
+
+			// If the statefulset is not using the RollingUpdate strategy, we assume it's healthy.
+			if obj.Spec.UpdateStrategy.Type != appsv1.RollingUpdateStatefulSetStrategyType {
+				continue
+			}
+			if obj.Status.ObservedGeneration < obj.Generation {
+				gvkErrors = appendResourceError(gvkErrors, obj, "StatefulSet is not ready")
```
ncdc, line 86:
```suggestion
				gvkErrors = appendResourceError(gvkErrors, obj, "StatefulSet is not ready (update has not yet been observed)")
```

internal/healthchecks/builtin_test.go
```diff
@@ -0,0 +1,604 @@
...
+					},
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "MyService",
+					},
+				},
+			},
+			expectedErr: false,
+		},
+		{
+			name: "resource with no conditions, return error",
```
ncdc, line 321: This needs a unique test case per kind

internal/healthchecks/builtin_test.go
```diff
@@ -0,0 +1,604 @@
...
+					},
+					Status: appsv1.DeploymentStatus{
+						Conditions: nil,
+					},
+				},
+			},
+			expectedErr: true,
+		},
+		{
+			name: "resources with conditions but not the one we are looking for, return error",
```
ncdc, line 375: This needs a unique test case per kind

jmprusi: done PTAL

internal/healthchecks/builtin_test.go
jmprusi: done PTAL

internal/healthchecks/builtin.go
jmprusi: done PTAL

ncdc (approved):

PR #738, opened 2023-10-10
Custom probing prototype

perdasilva:


internal/controllers/bundledeployment/bundledeployment.go
```diff
@@ -387,13 +389,82 @@ func (c *controller) reconcile(ctx context.Context, bd *rukpakv1alpha1.BundleDep
 			Message: "BundleDeployment is healthy",
 		})
 	}
+
+	// todo: unknown condition if there are no probes? or just no condition?
```
joelanford, line 393: If there are no probes, it seems like we would want to say Available=True, Reason=NoProbesConfigured or something along those lines.

You might find the [Kubernetes API conventions docs](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties) helpful for general questions like this though:
> Controllers should apply their conditions to a resource the first time they visit the resource, even if the status is Unknown. This allows other components in the system to know that the condition exists and the controller is making progress on reconciling that resource.
>
>  - Not all controllers will observe the previous advice about reporting "Unknown" or "False" values. For known conditions, the absence of a condition status should be interpreted the same as Unknown, and typically indicates that reconciliation has not yet finished (or that the resource state may not yet be observable).

perdasilva: nice - great call out!

PR #743, opened 2023-10-16
Add a bundle deployment reconciler to handle v1alpha2 API

varshaprasad96:
This PR does considerable refactoring to the existing code to introduce a new reconciler to handle v1alpha2 bundle deployment API. This does not remove or make any changes to the existing alpha1 code base. Once we have everything related to alpha2, we can cleanup and remove other controllers and manifests that are not required.

Major changes made:
1. A new reconciler introduced in: internal/v1alpha2/controllers/bundledeployment/bundledeployment.go
2. Store interface for creating a local fs for each bundle deployment to store unpacked contents: internal/v1alpha2/store
3. Unpacker interface and related code for unpacking multiple sources: internal/v1alpha2/source
4. Validator: internal/v1alpha2/validator
5. Deployer: internal/v1alpha2/deployer

The reconciler is not plugged into `core/main.go`. This PR splits #727 to make review process easier. All the features, not implemented here are tracked in there.


varshaprasad96:
cc: @perdasilva @ncdc

api/v1alpha1/bundledeployment_types.go
```diff
@@ -86,6 +86,9 @@ type BundleDeploymentStatus struct {
 //+kubebuilder:printcolumn:name=Provisioner,type=string,JSONPath=`.spec.provisionerClassName`,priority=1
 
 // BundleDeployment is the Schema for the bundledeployments API
+// TODO(varsha): Marking this as unserved version for now. Remove this definition during cleanup, when the all the
+// work on alpha2 is done.
+// +kubebuilder:unservedversion
```
varshaprasad96, line 91: This PR does not wire in the reconciler to the controller yet. Can make this as a storage version until that is done.

varshaprasad96:
Please review from the second commit. The first one would be rebased on #713

internal/v1alpha2/controllers/bundledeployment/bundledeployment.go
```diff
@@ -0,0 +1,110 @@
...
+// BundleDeploymentReconciler reconciles a BundleDeployment object
+type bundleDeploymentReconciler struct {
+	client.Client
+
+	Scheme   *runtime.Scheme
+	Recorder record.EventRecorder
+
+	controller        crcontroller.Controller
+	dynamicWatchMutex sync.RWMutex
+	dynamicWatchGVKs  map[schema.GroupVersionKind]struct{}
```
stevekuznetsov, line 82:
```go
sets.Set[schema.GroupVersionKind]
```

internal/v1alpha2/controllers/bundledeployment/bundledeployment.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+//+kubebuilder:rbac:groups=core.rukpak.io,resources=bundledeployments,verbs=list;watch
+//+kubebuilder:rbac:groups=core.rukpak.io,resources=bundledeployments/status,verbs=update;patch
+//+kubebuilder:rbac:groups=core.rukpak.io,resources=bundledeployments/finalizers,verbs=update
+//+kubebuilder:rbac:groups=*,resources=*,verbs=*
```
stevekuznetsov, line 97: uh oh :(

internal/v1alpha2/controllers/bundledeployment/bundledeployment.go
```diff
@@ -0,0 +1,110 @@
...
+
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
+// move the current state of the cluster closer to the desired state.
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.9.2/pkg/reconcile
+func (b *bundleDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log := log.FromContext(ctx)
+
+	existingBD := &v1alpha2.BundleDeployment{}
```
stevekuznetsov, line 106: nit: `rukpackv1alpha2`

internal/v1alpha2/controllers/bundledeployment/bundledeployment.go
```diff
@@ -0,0 +1,110 @@
...
+		log.Info("bundledeployment has been paused for reconciliation", "name", existingBD.Name)
+		return ctrl.Result{}, nil
+	}
+
+	reconciledBD := existingBD.DeepCopy()
+	res, reconcileErr := b.reconcile(ctx, reconciledBD)
+
+	// The controller is not updating spec, we only update the status. Hence sending
+	// a status update should be enough.
+	if !equality.Semantic.DeepEqual(existingBD.Status, reconciledBD.Status) {
```
stevekuznetsov, line 138: Can we use applyconfigurations? I imagine the code generators create them already; new code should be written to use them if possible.

internal/v1alpha2/store/interface.go
```diff
@@ -0,0 +1,53 @@
...
+	"errors"
+
+	"github.com/spf13/afero"
+)
+
+// The Store interface defines the required methods that must be implemented
+// by the bundle deployment store. These methods facilitate the unpacking and storage
+// of contents from all the sources.
+type Store interface {
+	afero.Fs
```
stevekuznetsov, line 30: I'm surprised you want to inline all the low-level bits from this.

internal/v1alpha2/store/store.go
```diff
@@ -0,0 +1,193 @@
...
+func (b *bundledeploymentStore) GetBundleDeploymentName() string {
+	return b.bundledeploymentName
+}
+
+func (b *bundledeploymentStore) GetBundleDirectory() string {
+	return b.baseDirectory
+}
+
+func (b *bundledeploymentStore) Create(name string) (afero.File, error) {
+	return b.fs.Create(name)
```
stevekuznetsov, line 149: If you do want to inline everything, why not just embed the `fs`?

internal/v1alpha2/source/image.go
```diff
@@ -0,0 +1,315 @@
...
+
+type logfn func(ctx context.Context, pod *corev1.Pod) ([]byte, error)
+
+func (i *image) Unpack(ctx context.Context, source *v1alpha2.BundleDeplopymentSource, store store.Store, opts UnpackOption) (*Result, error) {
+	// Validate inputs
+	if err := i.validate(source, opts); err != nil {
+		return nil, fmt.Errorf("validation unsuccessful during unpacking %v", err)
+	}
+	// storage path to store contents in local directory.
+	storagePath := filepath.Join(store.GetBundleDeploymentName(), filepath.Clean(source.Destination))
```
stevekuznetsov, line 63: You pass `store` and `source` into `i.unpack()` so you can likely inline this logic there instead of passing it in as a parameter

internal/v1alpha2/source/image.go
```diff
@@ -0,0 +1,315 @@
...
+	return i.unpack(ctx, storagePath, source, store, opts)
+}
+
+func (i *image) unpack(ctx context.Context, storagePath string, source *v1alpha2.BundleDeplopymentSource, store store.Store, opts UnpackOption) (*Result, error) {
+	pod := &corev1.Pod{}
+	op, err := i.ensureUnpackPod(ctx, store.GetBundleDeploymentName(), *source, pod, opts)
+	if err != nil {
+		return nil, err
+	} else if op == controllerutil.OperationResultCreated || op == controllerutil.OperationResultUpdated || pod.DeletionTimestamp != nil {
```
stevekuznetsov, line 72: If `i.ensureUnpackPod()` did not return Created or Updated, why is a deleting pod a pending unpack?

internal/v1alpha2/source/image.go
```diff
@@ -0,0 +1,315 @@
...
+		}
+		updatedPod, err = i.KubeClient.CoreV1().Pods(i.PodNamespace).Apply(ctx, podApplyConfig, metav1.ApplyOptions{Force: true, FieldManager: "rukpak-core"})
+		if err != nil {
+			return controllerutil.OperationResultNone, err
+		}
+	}
+
+	// make sure the passed in pod value is updated with the latest
+	// version of the pod
+	*pod = *updatedPod
```
stevekuznetsov, line 114: Probably best to return the pod instead of mutating in place - is there motivation for the side-effect approach?

internal/v1alpha2/source/image.go
```diff
@@ -0,0 +1,315 @@
...
+
+	// compare existingPod to newPod and return an appropriate
+	// OperatorResult value.
+	newPod := updatedPod.DeepCopy()
+	unsetNonComparedPodFields(existingPod, newPod)
```
stevekuznetsov, line 119: This is slightly hard to follow. If what you want is to compare `pod.Spec`, why not just do that?

internal/v1alpha2/source/image.go
```diff
@@ -0,0 +1,315 @@
...
+		return controllerutil.OperationResultNone, err
+	}
+
+	podApplyConfig := i.getDesiredPodApplyConfig(bdName, &bundleSrc, opts)
+	updatedPod, err := i.KubeClient.CoreV1().Pods(i.PodNamespace).Apply(ctx, podApplyConfig, metav1.ApplyOptions{Force: true, FieldManager: "rukpak-core"})
+	if err != nil {
+		if !apierrors.IsInvalid(err) {
+			return controllerutil.OperationResultNone, err
+		}
+		if err := i.Client.Delete(ctx, existingPod); err != nil {
```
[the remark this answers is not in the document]
stevekuznetsov: Or is this factored to start with an apply for the happy case where there's nothing to change? I imagine there's a way to compute whether or not a change has occurred on the Bundle object and determine if it's needed to update the pod without a call to the API server

internal/v1alpha2/source/image.go
```diff
@@ -0,0 +1,315 @@
...
+	}
+	return &Result{State: StateUnpackPending, Message: strings.Join(messages, "; ")}
+}
+
+func (i *image) failedPodResult(ctx context.Context, pod *corev1.Pod) error {
+	logs, err := i.getPodLogs(ctx, pod)
+	if err != nil {
+		return fmt.Errorf("unpack failed: failed to retrieve failed pod logs: %v", err)
+	}
+	_ = i.Client.Delete(ctx, pod)
```
stevekuznetsov, line 237: Likely want to ignore not found and return the error?

internal/v1alpha2/controllers/bundledeployment/bundledeployment.go
```diff
@@ -46,11 +49,26 @@ type bundleDeploymentReconciler struct {
...
+	unpacker source.Unpacker
+
 	dynamicWatchMutex sync.RWMutex
 	dynamicWatchGVKs  map[schema.GroupVersionKind]struct{}
 }
 
+// Options to configure bundleDeploymentReconciler
+type Option func(bd *bundleDeploymentReconciler)
+
+func WithUnpacker(u *source.Unpacker) Option {
```
stevekuznetsov, line 88: When is it valid not to have an unpacker?

internal/v1alpha2/controllers/bundledeployment/bundledeployment.go
```diff
@@ -97,15 +124,57 @@ func (b *bundleDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+	// result can be nil, when there is an error during unpacking. This indicates
+	// that unpacking was failed. Update the status accordingly.
+	if res == nil || err != nil {
+		setUnpackStatusFailing(&bundleDeployment.Status.Conditions, fmt.Sprintf("unpack unsuccessful %v", err), bundleDeployment.Generation)
 		return ctrl.Result{}, err
 	}
 
+	switch res.State {
+	case source.StateUnpackPending:
+		setUnpackStatusPending(&bundleDeployment.Status.Conditions, fmt.Sprintf("pending unpack"), bundleDeployment.Generation)
```
stevekuznetsov, line 162: Shouldn't we be watching pods with some unpacking label and triggering off of them changing state?

varshaprasad96:
@stevekuznetsov The unpacking and deploy logic is copied from v1alpha1 bits. Implementing direct client unpacking is not a now a part of this right now. The design for this was to allow different implementations of store, unpacker and deployer. Some of this are to be passed per bundle deployment (store), whereas unpacker, validator and deployer are passed while creating the controller. I'll look to some of your comments and address those.

internal/v1alpha2/store/interface.go
varshaprasad96: Right now, the sources use limited number of helpers just for creating/removing/verifying contents inside dir. Instead of specifying them separately, inlining of `afero.Fs` would make it easier and save us from adding more helpers to this interface. I assume we would be using this only in rukpak, that too to access localdir, so expecting an afero.Fs implementation to be passed seemed reasonable.

internal/v1alpha2/store/store.go
varshaprasad96: That's a good point. The only benefit for doing this was creating a chroot like view for the bundle deployment while creating a newBundleDeploymentStore in here: https://github.com/operator-framework/rukpak/pull/743/files#diff-f1022d4c85936c453f8a541cdd88c667367ac932b582d3dadaeb01dd547c9d8bR64. But this could be offloaded in the reconciler itself. 2 additional lines in there, can save a bunch of code here. Will refactor this out.

internal/v1alpha2/source/image.go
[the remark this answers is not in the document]
varshaprasad96: > Or is this factored to start with an apply for the happy case where there's nothing to change?
This has been taken from the current alpha1 implementation. But this is the only reason I see. 

> I imagine there's a way to compute whether or not a change has occurred on the Bundle object and determine if it's needed to update the pod without a call to the API server
This in general needs to done on a higher level, regardless of the image unpacking using a pod. Right now, even though there is no change, the design unpacks during every reconcile (whether image, git or http) which is unnecessary. Probably a caching mechanism, where we compute a hash of the bdName and bundleSource to figure out if there has been a change in the bundle content would be a better way to reduce time and resource during reconcile. I had implemented something similar in #727, but wanted to keep it separate for now. If this is implemented, then we need to patch, instead as you mentioned, delete+create as we would be sure of the change.

internal/v1alpha2/controllers/bundledeployment/bundledeployment.go
varshaprasad96: I know! But I have no idea on the intention behind this. Copied from https://github.com/operator-framework/rukpak/blob/bcae27f71c82b396dcf141b4fc557f06a089712d/internal/controllers/bundledeployment/bundledeployment.go#L160

internal/v1alpha2/controllers/bundledeployment/bundledeployment.go
stevekuznetsov: Shouldn't we just have this as a required field in the constructor instead of an option?

internal/v1alpha2/deployer/helm.go
everettraven, line 1: Is this net new or has this code been moved+modified?

internal/v1alpha2/deployer/interface.go
```diff
@@ -0,0 +1,65 @@
...
+
+// Deployer knows how apply objects on cluster.
+type Deployer interface {
+	// Deploy deploys the contents from fs onto the cluster and returns list of the deployed object.
+	// Pass the clients specific to deployer with Options. Refer to Helm deployer for example.
+	Deploy(ctx context.Context, store store.Store, bundleDeployment *v1alpha2.BundleDeployment) (*Result, error)
+}
+
+// Result conveys the progress information about deploying content.
+// TODO: Refactor to use the same result struct for unpacking and deployment.
```
everettraven, line 36: What prevents us from doing this now?

internal/v1alpha2/controllers/bundledeployment/bundledeployment.go
```diff
@@ -0,0 +1,361 @@
...
+			errs = append(errs, err)
+		}
+	}
+	return apimacherrors.NewAggregate(errs)
+}
+
+// deployContents calls the registered deployer to apply the bundle contents onto the cluster.
+func (b *bundleDeploymentReconciler) deployContents(ctx context.Context, store store.Store, bd *v1alpha2.BundleDeployment) (*deployer.Result, error) {
+	return b.deployer.Deploy(ctx, store, bd)
+}
```
everettraven, line 283: This function seems unneccessary. We could just use the `b.deployer.Deploy()` function directly

internal/v1alpha2/controllers/bundledeployment/bundledeployment.go
```diff
@@ -0,0 +1,361 @@
...
+		setReconcileStatusFailed(&bundleDeployment.Status.Conditions, err.Error(), bundleDeployment.Generation)
+		return ctrl.Result{}, err
+	case deployer.StateObjectFetchFailed:
+		setDynamicWatchFailed(&bundleDeployment.Status.Conditions, err.Error(), bundleDeployment.Generation)
+		return ctrl.Result{}, err
+	case deployer.StateDeploySuccessful:
+		setInstallStatusSuccess(&bundleDeployment.Status.Conditions, fmt.Sprintf("installed %s", bundleDeployment.GetName()), bundleDeployment.Generation)
+	default:
+		return ctrl.Result{}, fmt.Errorf("unknown deploy state %q for bundle deployment %s: %v", deployResult.State, bundleDeployment.GetName(), bundleDeployment.Generation)
+	}
```
everettraven, line 205: Is there a way we could make deployContents return the condition that should be set?

ncdc:
Have discussed with @varshaprasad96 @gallettilance @joelanford - we are going to investing using https://carvel.dev/kapp-controller/ as a wholesale replacement for rukpak. kapp-controller does almost everything rukpak currently does (it doesn't yet support registry+v1 bundles) and it will save us a lot of time if we can switch to using it.

Closing this for now. If for some reason kapp-controller doesn't work out, we can revisit.

internal/v1alpha2/controllers/bundledeployment/bundledeployment.go
```diff
@@ -0,0 +1,361 @@
...
+	if b.validators == nil || len(b.validators) == 0 {
+		errs = append(errs, errors.New("validators not provided"))
+	}
+	if b.deployer == nil {
+		errs = append(errs, errors.New("deployer is unset"))
+	}
+	return apimacherrors.NewAggregate(errs)
+}
+
+func SetupWithManager(mgr manager.Manager, systemNsCache cache.Cache, opts ...Option) error {
```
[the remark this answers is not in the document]
varshaprasad96: Yes, the intention was to have a slimmed down cache that watches objects with ownerKind label set to rukpak. That would be passed when this is set up with the core controller.

internal/v1alpha2/deployer/interface.go
varshaprasad96: Because of ResolvedSource field. This api was intended to accept multiple resolved sources for a single BD. So that would need more discussion.

grokspawn:
> Have discussed with @varshaprasad96 @gallettilance @joelanford - we are going to pause on this for the moment.
> 
> During the pause, we are going to investigate using https://carvel.dev/kapp-controller/ an alternative to continuing to restructure the rukpak APIs. kapp-controller does almost everything rukpak currently does (it doesn't yet support registry+v1 bundles) and it will save us a lot of time if we can switch to using it.
> 
> Closing this for now. We can reopen depending on the results of our research.

Haven't seen this in WG discussion.  Have I missed a thread?

ncdc:
No missed threads. This is something I've been talking about with @joelanford and a few other folks.

grokspawn:
Should we interrupt in-flight PRs once it matures to WG-wide discussion?

PR #745, opened 2023-10-19
use k8s cli-utils health checks for bundledeployment resources

ankitathomas:
Use the third party kstatus package from [sigs.k8s.io/cli-utils](https://github.com/kubernetes-sigs/cli-utils/tree/master/pkg/kstatus) for health checks for bundledeployment resources


internal/healthchecks/builtin.go
```diff
@@ -45,147 +47,38 @@ func AreObjectsHealthy(ctx context.Context, client client.Client, objects []clie
...
-				continue
-			}
-			conditionExists := false
-			for _, condition := range obj.Status.Conditions {
-				if condition.Type == apiextensionsv1.Established {
-					if condition.Status != "True" {
-						gvkErrors = appendResourceError(gvkErrors, obj, condition.Message)
-					}
-					conditionExists = true
+					isAvailable = &condition
```
everettraven, line 61: As an alternative to @tmshort 's [recommendation](https://github.com/operator-framework/rukpak/pull/745/files#r1365973287) I _think_ you could also do something like:
```suggestion
					availCond := condition
					isAvailable = &availCond
```

internal/healthchecks/builtin.go
```diff
@@ -45,147 +47,38 @@ func AreObjectsHealthy(ctx context.Context, client client.Client, objects []clie
...
-			conditionExists := false
-			for _, condition := range obj.Status.Conditions {
+			var isAvailable *apiregistrationv1.APIServiceCondition
+			for i, condition := range obj.Status.Conditions {
 				if condition.Type == apiregistrationv1.Available {
-					if condition.Status != "True" {
-						gvkErrors = appendResourceError(gvkErrors, obj, condition.Message)
-					}
-					conditionExists = true
+					isAvailable = &obj.Status.Conditions[i]
```
everettraven, line 61: That works also :)

everettraven (approved):

tmshort (approved):
/lgtm

tmshort:
CodeCov dropped... not sure we care?

everettraven:
>CodeCov dropped... not sure we care?

IMO not for the amount it dropped

everettraven:
Removed from merge queue as I saw @ankitathomas pushed just now

tmshort (approved):
/lgtm
The error checking is more specific in the tests.

internal/healthchecks/builtin_test.go
```diff
@@ -135,7 +148,7 @@ func TestAreObjectsHealthy(t *testing.T) {
 					},
 				},
 			},
-			expectedErr: false,
+			expectedErr: "",
```
everettraven, line 151: TLDR: IMO, check error strings instead of "an error occurred" is going to result in unit tests that are way to brittle when depending on a third-party library.

While I get the desire to use a string here, I think this is going to result in much more brittle test cases. There was a thread at some point in the `#olm-dev` channel about this and while I don't have a dog in the fight of "we should vs we shouldn't check error strings in tests", I do think this falls into territory where just checking if an error occurred is more appropriate specifically because the important part of the error message is going to be coming from a function call that we don't control (the third-party library). This means that a simple dependency bump could result in unit test failures if they change the values that are returned by `result.Message`  which could realistically be any string.

everettraven (approved):

tmshort (approved):
/lgtm

PR #758, opened 2023-11-06
Remove pointer to `BundleTemplate` from spec

m1kola:
`Template` field in `BundleDeploymentSpec` seems to have unnecessary pointer as we expect to always populate this field.


internal/util/util.go
```diff
@@ -67,10 +67,10 @@ func ReconcileDesiredBundle(ctx context.Context, c client.Client, bd *rukpakv1al
 
 	// check whether there's an existing Bundle that matches the desired Bundle template
 	// specified in the BI resource, and if not, generate a new Bundle that matches the template.
-	b := CheckExistingBundlesMatchesTemplate(existingBundles, bd.Spec.Template)
+	b := CheckExistingBundlesMatchesTemplate(existingBundles, &bd.Spec.Template)
```
stevekuznetsov, line 70: Probably best that the downstream consumers don't take a pointer unless `nil` is a valuable input?

m1kola: Updated

stevekuznetsov (approved):

PR #764, opened 2023-11-17
use stable hash for labels and name suffixes (aligning with OLMv0 changes)

joelanford:
See https://github.com/operator-framework/operator-lifecycle-manager/pull/3115


internal/convert/registryv1.go
```diff
@@ -279,13 +276,19 @@ func Convert(in RegistryV1, installNamespace string, targetNamespaces []string)
 
 	for _, permission := range permissions {
 		saName := saNameOrDefault(permission.ServiceAccountName)
-		name := generateName(fmt.Sprintf("%s-%s", in.CSV.Name, saName), []interface{}{in.CSV.Name, permission})
+		name, err := generateName(fmt.Sprintf("%s-%s", in.CSV.Name, saName), []interface{}{in.CSV.Name, permission})
```
stevekuznetsov, line 279: I didn't change this downstream for simplicity, but you don't need to include `in.CSV.Name` in the hash if it's in the prefix.

Also, use an anonymous struct here to make it JSON-serializable (see my changes in v0).

internal/convert/registryv1.go
```diff
@@ -279,13 +276,19 @@ func Convert(in RegistryV1, installNamespace string, targetNamespaces []string)
...
+		if err != nil {
+			return nil, err
+		}
 		roles = append(roles, newRole(installNamespace, name, permission.Rules))
 		roleBindings = append(roleBindings, newRoleBinding(installNamespace, name, name, installNamespace, saName))
 	}
 	for _, permission := range clusterPermissions {
 		saName := saNameOrDefault(permission.ServiceAccountName)
-		name := generateName(fmt.Sprintf("%s-%s", in.CSV.Name, saName), []interface{}{in.CSV.GetName(), permission})
+		name, err := generateName(fmt.Sprintf("%s-%s", in.CSV.Name, saName), []interface{}{in.CSV.GetName(), permission})
```
stevekuznetsov, line 288: Use an anonymous struct here to make it JSON-serializable (see my changes in v0).

internal/convert/registryv1.go
joelanford: Also, turns out getting rid of the CSV name means I can just pass `permission` directly without the interface slice wrapper.

stevekuznetsov (approved):

stevekuznetsov:
@joelanford e2e looks very borked

joelanford:
Looks like the bundle validation is rejecting the newly generated bundle name. Trying to remember why we chose `a lowercase RFC 1123 subdomain` validation for the name.

> Bundle.core.rukpak.io "ahoy-v8vq7-6L2Nq4N6UYLYpmq6ffZkPgbGnbi6mxZCYug8PO" is invalid: metadata.name: Invalid value: "ahoy-v8vq7-6L2Nq4N6UYLYpmq6ffZkPgbGnbi6mxZCYug8PO": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, ''-'' or ''.'', and must start and end with an alphanumeric character (e.g. ''example.com'', regex used for validation is ''[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*'')'

stevekuznetsov:
That error might be misleading, I think you pass the regex fine, but may be running afoul of the length requirement. Suggest not adding the prefixes unless you really need to?

joelanford:
It definitely looks like the regex rejects uppercase letters. I changed to base36 and now seeing the length validation failure.

> Bundle.core.rukpak.io "cincinnatiz6xhc-34402ftwebb3yugoq821w197i1988guvd9k6oqjcgm92" is invalid: metadata.name: Too long: may not be longer than 52'

joelanford:
I'm kinda leaning toward base36 + truncate at like 8 characters or something?

joelanford:
length of 52 is coming from here: https://github.com/operator-framework/rukpak/blob/4f98bdd43984bc647b05789a465a10f5cb566af6/manifests/base/apis/crds/patches/bundle_validation.yaml#L8

I think that was because we knew that we wanted to leave some extra characters available for other objects that might need to exist for the bundle (right now, that's at least the unpack pod, can't remember if there are others)

stevekuznetsov (approved):

PR #772, opened 2023-12-15
[Add] Support for other Install strategies

varshaprasad96:
This PR removes the requirement that AllNamespace mode should be supported always in a registry+v1 format.

It does so by:
1. Adding a field to specify watch namespaces in Bundle Template.
2. While converting from registryV1 to plain bundle, we now generate roles for all specified targetNamespaces which are being watched.


api/v1alpha1/bundle_types.go
```diff
@@ -56,6 +56,9 @@ type BundleSpec struct {
 	ProvisionerClassName string `json:"provisionerClassName"`
 	// Source defines the configuration for the underlying Bundle content.
 	Source BundleSource `json:"source"`
+	// +kubebuilder:validation:Optional
+	// Watchnamespaces refers to the namespaces the operator is allowed to watch.
```
ncdc, line 60:
```suggestion
	// watchNamespaces indicates which namespaces the operator should watch.
```

internal/convert/registryv1.go
```diff
@@ -274,15 +271,18 @@ func Convert(in RegistryV1, installNamespace string, targetNamespaces []string)
 		permissions = nil
 	}
 
-	for _, permission := range permissions {
-		saName := saNameOrDefault(permission.ServiceAccountName)
-		name, err := generateName(fmt.Sprintf("%s-%s", in.CSV.Name, saName), permission)
-		if err != nil {
-			return nil, err
+	for _, ns := range targetNamespaces {
```
ncdc, line 274: What should we do if this list is empty?

ncdc (changes requested):
Will you also be adding the annotation part in this PR?

internal/convert/registryv1.go
varshaprasad96: We already have validation for that in 2 places:
1. We ensure if targetNamespaces is empty, AllNs is supported: https://github.com/operator-framework/rukpak/blob/b4e9976e748299623d5e2548c7bf1b3a0c40edf1/internal/convert/registryv1.go#L147
2. If CSV supports AllNs or OwnNs, then accordingly fill it in here: https://github.com/operator-framework/rukpak/blob/b4e9976e748299623d5e2548c7bf1b3a0c40edf1/internal/convert/registryv1.go#L198
3. If its AllNs mode, then we elevate permissions to scaffold only CR and CRD and set permissions to nil: https://github.com/operator-framework/rukpak/blob/b4e9976e748299623d5e2548c7bf1b3a0c40edf1/internal/convert/registryv1.go#L274.

internal/convert/registryv1.go
```diff
@@ -189,9 +189,6 @@ func Convert(in RegistryV1, installNamespace string, targetNamespaces []string)
 			supportedInstallModes.Insert(string(im.Type))
 		}
 	}
-	if !supportedInstallModes.Has(string(v1alpha1.InstallModeTypeAllNamespaces)) {
-		return nil, fmt.Errorf("AllNamespace install mode must be enabled")
-	}
 	if targetNamespaces == nil {
```
ncdc, line 192: Would you mind changing this to check for len=0 instead please?

internal/convert/registryv1.go
ncdc: Thanks!

internal/convert/registryv1.go
```diff
@@ -165,8 +165,8 @@ func validateTargetNamespaces(supportedInstallModes sets.Set[string], installNam
 	return fmt.Errorf("supported install modes %v do not support target namespaces %v", sets.List[string](supportedInstallModes), targetNamespaces)
 }
 
-func Simple(in RegistryV1) (*Plain, error) {
-	return Convert(in, "", nil)
+func Simple(in RegistryV1, watchNamespaces []string) (*Plain, error) {
+	return Convert(in, "", watchNamespaces)
 }
```
joelanford, line 170: I think I'd just drop the `Simple` function at this point.

api/v1alpha1/bundle_types.go
```diff
@@ -56,6 +56,9 @@ type BundleSpec struct {
 	ProvisionerClassName string `json:"provisionerClassName"`
 	// Source defines the configuration for the underlying Bundle content.
 	Source BundleSource `json:"source"`
+	// +kubebuilder:validation:Optional
+	// watchNamespaces indicates which namespaces the operator should watch.
+	WatchNamespaces []string `json:"watchNamespaces,omitempty"`
```
joelanford, line 61: Hmm. I had imagined that we'd put this on the BundleDeployment, not the Bundle. But the wrinkle with needing to generate extra RBAC during the conversion means this is essential information in that step.

One possible solution here would be to make a registry BundleDeployment provisioner and move the conversion logic there. If we did that, we could probably make the Bundle provisioner super generic and reusable for registry/plain/helm/etc. Basically it would just be a "give me the filesystem with no format-specific validation" provisioner. And then the BD would be the place for registry/plain/helm to have separate provisioners and do whatever specific validation was necessary.

From an API design standpoint, I argue that configuration of the templating of the bundle belongs in the BundleDeployment.

All of this highlights the weird dichotomy between Bundle and BundleDeployment, and I think it's another good reason to move to a single API for "source/template/apply" like the Carvel App API. I think moving toward a super-generic Bundle provisioner gets us more aligned with that concept anyway, so perhaps that would be a good step in that direction?

What do you think?

ncdc: I missed this was on Bundle and not BundleDeployment. +100 to getting rid of Bundle. @varshaprasad96 if you're around this week and want help, please let me know.

varshaprasad96: @joelanford @ncdc fair. The reason for having this on template, was to be able to pass this to bundle provisioner which handles the conversion. If we move that logic to `Bundle`, there doesn't seem to be much value in having a separate API. 

> I think it's another good reason to move to a single API for "source/template/apply" like the Carvel App API

Removing Bundle would change the BundleDeployment API, we would need an alpha2? Nevertheless, I'll get a draft PR ready, and we can discuss the specifics of the API.

ncdc: Yes to alpha2

PR #774, opened 2023-12-18
Remove Bundle API and its controller

varshaprasad96:
As discussed [here](https://github.com/operator-framework/rukpak/pull/772#discussion_r1428695817) this PR intends to remove the Bundle API for purposes of having a single interface to handle bundle unpacking, templating and installation. 

The following are the changes made:
- [x]  Move the BundleSource specifications to Bundle Deployment API.
- [x] Move the statuses which were initially a part of Bundle, to Bundle Deployment.
- [x] Refactor Bundle Deployment controller to perform unpacking.
- [x] Remove Bundle controller.
- [x] Fix e2e tests.
    - [x] Fix provisioner tests
    - [x] Fix webhook tests
    - [x] Fix api validator tests  
- [x] Remove upload src, rukpakctl and relevant code. 
- [x] Fix unit tests.
- [x] Fix the crd and web hook validator, to validate the sources in Bundle Deployment instead of Bundle.
- [x] Bump the Bundle Deployment version from v1alpha1 to v1alpha2 in light of breaking API.
- [x] Re-scaffold necessary manifests (ref: https://github.com/operator-framework/rukpak/issues/777). 
- [x] Make linter happy. 

TODO in follow ups:
- [ ] Fix documentation.
- [ ] Clean up all the comments in the code base that reference bundle API instead of bundle deployments.


internal/controllers/bundledeployment/interfaces.go
```diff
@@ -19,3 +19,16 @@ type HandlerFunc func(context.Context, fs.FS, *rukpakv1alpha1.BundleDeployment)
 func (f HandlerFunc) Handle(ctx context.Context, fsys fs.FS, bd *rukpakv1alpha1.BundleDeployment) (*chart.Chart, chartutil.Values, error) {
 	return f(ctx, fsys, bd)
 }
+
+// TODO: Having two interfaces with same parameters seems unnecessary. This should ideally
+// be moved to HandleBundleDeployment. With it, we can remove the additional step of loading from
+// store.
```
varshaprasad96, line 25: TODO: Revisit the use of storage interface

internal/storage/http_test.go
```diff
@@ -17,21 +17,21 @@ import (
...
 
 var _ = Describe("HTTP", func() {
 	var (
-		ctx        context.Context
-		bundle     *rukpakv1alpha1.Bundle
-		testFS     fs.FS
-		localStore *LocalDirectory
-		server     *httptest.Server
+		ctx              context.Context
+		bundledeployment *rukpakv1alpha2.BundleDeployment
```
ncdc, line 27: Global: please make sure to use camelCase. In other words: `bundleDeployment` instead of `bundledeployment`

internal/util/util.go
```diff
@@ -177,61 +109,13 @@ func MapOwneeToOwnerProvisionerHandler(ctx context.Context, cl client.Client, lo
 	})
 }
 
-// MapBundleToBundleDeployment is responsible for finding the BundleDeployment resource
```
ncdc, line 180: Please retain+update the godoc

internal/util/util.go
```diff
@@ -177,61 +109,13 @@ func MapOwneeToOwnerProvisionerHandler(ctx context.Context, cl client.Client, lo
...
-		bd := bd
-
-		if bd.GetName() == bundleOwnerName {
-			return bd.DeepCopy()
-		}
-	}
-	return nil
-}
-
-// MapBundleToBundleDeploymentHandler is responsible for requeuing a BundleDeployment resource
```
ncdc, line 209: Please retain+update the godoc

internal/util/util.go
```diff
@@ -335,9 +178,9 @@ func GenerateBundleName(bdName, hash string) string {
 	return fmt.Sprintf("%s-%s", bdName, hash)
 }
 
-// SortBundlesByCreation sorts a BundleList's items by it's
+// SortBundleDeploymentsByCreation sorts a BundleDeploymentList's items by it's
```
ncdc, line 181:
```suggestion
// SortBundleDeploymentsByCreation sorts a BundleDeploymentList's items by its
```

cmd/core/main.go
```diff
@@ -116,7 +115,7 @@ func main() {
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
 	setupLog.Info("starting up the core controllers and servers", "git commit", version.String(), "unpacker image", unpackImage)
 
-	dependentRequirement, err := labels.NewRequirement(util.CoreOwnerKindKey, selection.In, []string{rukpakv1alpha1.BundleKind, rukpakv1alpha1.BundleDeploymentKind})
+	dependentRequirement, err := labels.NewRequirement(util.CoreOwnerKindKey, selection.In, []string{rukpakv1alpha2.BundleKind, rukpakv1alpha2.BundleDeploymentKind})
```
joelanford, line 118: Since there is no longer a `Bundle` kind, we should probably be using the `BundleDeploymentKind` in this label whenever we create other cluster objects for unpacking (haven't gotten far enough in the review to see if this is the case). But if we're doing that, we no longer need this requirement to include the `BundleKind` label value.
```suggestion
	dependentRequirement, err := labels.NewRequirement(util.CoreOwnerKindKey, selection.In, []string{rukpakv1alpha2.BundleDeploymentKind})
```

cmd/core/main.go
```diff
@@ -231,44 +229,34 @@ func main() {
...
+	if err := bundledeployment.SetupWithManager(mgr, systemNsCluster.GetCache(), systemNamespace, append(
 		commonBDProvisionerOptions,
-		bundledeployment.WithProvisionerID(plain.ProvisionerID),
+		bundledeployment.WithUnpacker(unpacker),
+		bundledeployment.WithProvisionerID(registry.ProvisionerID),
+		bundledeployment.WithBundleDeplymentProcessor(bundledeployment.ProcessorFunc(registry.ProcessBundleDeployment)),
 		bundledeployment.WithHandler(bundledeployment.HandlerFunc(plain.HandleBundleDeployment)),
 	)...); err != nil {
-		setupLog.Error(err, "unable to create controller", "controller", rukpakv1alpha1.BundleDeploymentKind, "provisionerID", plain.ProvisionerID)
+		setupLog.Error(err, "unable to create controller", "controller", rukpakv1alpha2.BundleDeploymentKind, "provisionerID", plain.ProvisionerID)
```
joelanford, line 259:
```suggestion
		setupLog.Error(err, "unable to create controller", "controller", rukpakv1alpha2.BundleDeploymentKind, "provisionerID", registry.ProvisionerID)
```

cmd/rukpakctl/cmd/run.go
```diff
@@ -73,7 +73,6 @@ one version to the next.
 			}
 			_, err := r.Run(ctx, bundleDeploymentName, os.DirFS(bundleDir), rukpakctl.RunOptions{
 				BundleDeploymentProvisionerClassName: bundleDeploymentProvisionerClassName,
-				BundleProvisionerClassName:           bundleProvisionerClassName,
```
joelanford, line 76: Remove the `bundleProvisionerClassName` variable and flag configuration for this field as well?

internal/controllers/bundledeployment/bundledeployment.go
```diff
@@ -145,21 +186,31 @@ func (c *controller) validateConfig() error {
...
+	finalizers        crfinalizer.Finalizers
 	dynamicWatchMutex sync.RWMutex
 	dynamicWatchGVKs  map[schema.GroupVersionKind]struct{}
 }
 
-//+kubebuilder:rbac:groups=core.rukpak.io,resources=bundledeployments,verbs=list;watch
-//+kubebuilder:rbac:groups=core.rukpak.io,resources=bundledeployments/status,verbs=update;patch
+// TODO: recheck rbac
 //+kubebuilder:rbac:groups=core.rukpak.io,resources=bundledeployments/finalizers,verbs=update
+//+kubebuilder:rbac:groups=core.rukpak.io,resources=bundledeployments,verbs=list;watch;update;patch
```
joelanford, line 205: Are the `update` and `patch` verbs actually necessary here? I see that we had them for `Bundles`, so maybe so? Only thing I can think of is that this is necessary to update finalizers and somehow the `bundledeployments/finalizers` permission is insufficient or not doing what we think it does?

internal/controllers/bundledeployment/bundledeployment.go
```diff
@@ -198,53 +249,79 @@ func (c *controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 // Today we always return ctrl.Result{} and an error.
 // But in the future we might update this function
 // to return different results (e.g. requeue).
-func (c *controller) reconcile(ctx context.Context, bd *rukpakv1alpha1.BundleDeployment) (ctrl.Result, error) {
+func (c *controller) reconcile(ctx context.Context, bd *rukpakv1alpha2.BundleDeployment) (ctrl.Result, error) {
 	bd.Status.ObservedGeneration = bd.Generation
 
-	bundle, allBundles, err := util.ReconcileDesiredBundle(ctx, c.cl, bd)
+	// handle finalizers
+	finalizerBundleDelpoyment := bd.DeepCopy()
```
joelanford, line 256: Do we need another deep copy of the bundle deployment here? We deep copy in `Reconcile` already.

internal/controllers/bundledeployment/bundledeployment.go
```diff
@@ -198,53 +249,79 @@ func (c *controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	}
+
+	switch unpackResult.State {
+	case unpackersource.StatePending:
+		updateStatusUnpackPending(&bd.Status, unpackResult)
+		// There must a limit to number of retries if status is stuck at
+		// unpack pending.
+		// Forcefully requing here, to ensure that the next reconcile
+		// is triggered.
+		// TODO: Caliberate the requeue interval if needed.
```
joelanford, line 296: I think the expectation here is that the controller will receive an event when the underlying unpacking process moves beyond Pending, so I'm wondering if there's a need to use `RequeueAfter`?

The Pending state is used:
- by the 'image' unpacker which uses a pod (and we get pod events when the pod moves on from pending)
- by the 'upload' source. When the upload manager receives an upload it directly updates the status of the BD, which triggers a reconcile.

internal/controllers/bundledeployment/bundledeployment.go
```diff
@@ -198,53 +249,79 @@ func (c *controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
 
-	bundleFS, err := c.storage.Load(ctx, bundle)
+		if err := c.storage.Store(ctx, bd, storeFS); err != nil {
+			return ctrl.Result{}, updateStatusUnpackFailing(&bd.Status, fmt.Errorf("persist bundle content: %v", err))
+		}
+		contentURL, err := c.storage.URLFor(ctx, bd)
+		if err != nil {
+			return ctrl.Result{}, updateStatusUnpackFailing(&bd.Status, fmt.Errorf("get content URL: %v", err))
+		}
+		updateStatusUnpacked(&bd.Status, unpackResult, contentURL)
```
joelanford, line 314: I wonder if we should just store the unpacked contents directly without modification, and then move the logic that is in the processor into the existing handler?

internal/controllers/bundledeployment/bundledeployment.go
```diff
@@ -198,53 +249,79 @@ func (c *controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	default:
+		return ctrl.Result{}, updateStatusUnpackFailing(&bd.Status, fmt.Errorf("unknown unpack state %q: %v", unpackResult.State, err))
+	}
+
+	bundleFS, err := c.storage.Load(ctx, bd)
 	if err != nil {
 		meta.SetStatusCondition(&bd.Status.Conditions, metav1.Condition{
-			Type:    rukpakv1alpha1.TypeHasValidBundle,
+			Type:    rukpakv1alpha2.TypeHasValidBundle,
```
joelanford, line 322: Now that we've moved the bundle conditions into the bundle deployment, WDYT about dropping the `HasValidBundle` condition from the bundle deployment?

internal/controllers/bundledeployment/interfaces.go
```diff
@@ -7,15 +7,28 @@ import (
...
+type HandlerFunc func(context.Context, fs.FS, *rukpakv1alpha2.BundleDeployment) (*chart.Chart, chartutil.Values, error)
 
-func (f HandlerFunc) Handle(ctx context.Context, fsys fs.FS, bd *rukpakv1alpha1.BundleDeployment) (*chart.Chart, chartutil.Values, error) {
+func (f HandlerFunc) Handle(ctx context.Context, fsys fs.FS, bd *rukpakv1alpha2.BundleDeployment) (*chart.Chart, chartutil.Values, error) {
 	return f(ctx, fsys, bd)
 }
+
+// TODO: Having two interfaces with same parameters seems unnecessary. This should ideally
+// be moved to HandleBundleDeployment. With it, we can remove the additional step of loading from
+// store.
```
joelanford, line 25: I think it probably makes sense to keep storing the unpacked contents. That way future reconciles can see that the data is already present and skip the unpack step.

But I agree that this interface is unnecessary. I think we can just store whatever comes from the source unmodified, and then any templating or modification could happen at the beginning of the Handler function.

internal/rukpakctl/run.go
```diff
@@ -67,27 +59,17 @@ func (r *Run) Run(ctx context.Context, bundleDeploymentName string, bundle fs.FS
 		return false, err
 	}
 
-	bundleLabels := map[string]string{
+	bundleDeploymentLabels := map[string]string{
```
joelanford, line 62: The fact that the BD had a bundle template with more metadata was important for this upload use case, specifically for the use case of uploading new contents for an existing BD to pivot from the old bundle content to the new bundle content).

Whenever an upload happens, the client gets a digest of the uploaded data and includes that in the bundle template metadata. That way, when the BD reconciles, it sees that the bundle template is different and creates a new Bundle object, which allows the upload web service to accept an HTTP POST for the new data.

Now, it seems like we change the BDs labels, but in the case that the BD is already unpacked, the upload manager will fail to upload new content to it.

We likely need a few more tweaks here to make things work without an underlying Bundle API.

internal/rukpakctl/run.go
```diff
@@ -99,59 +81,37 @@ func (r *Run) Run(ctx context.Context, bundleDeploymentName string, bundle fs.FS
...
 		RootCAs:                rukpakCA,
 	}
-	modified, err := bu.Upload(ctx, bundleName, bundle)
+	modified, err := bu.Upload(ctx, bundleDeploymentName, bundle)
 	if err != nil {
 		return false, fmt.Errorf("failed to upload bundle: %v", err)
 	}
 	if !modified {
-		opts.Log("bundle %q is already up-to-date\n", bundleName)
+		opts.Log("bundle %q is already up-to-date\n", bundleDeploymentName)
```
joelanford, line 89:
```suggestion
		opts.Log("bundledeployment %q is already up-to-date\n", bundleDeploymentName)
```

internal/util/util.go
varshaprasad96: This method has been removed.

internal/util/util.go
varshaprasad96: This method has been removed.

internal/controllers/bundledeployment/bundledeployment.go
varshaprasad96: Had added these verbs because they were in Bundles. Will test this again and get back.

internal/controllers/bundledeployment/bundledeployment.go
varshaprasad96: This is required because we are finalizing the deepCopy, and if there is any change, updating the finalizers: https://github.com/operator-framework/rukpak/blob/4f54625e9eec52b211fdaf1c41c00b86715b8022/internal/controllers/bundledeployment/bundledeployment.go#L271-L278

internal/controllers/bundledeployment/bundledeployment.go
varshaprasad96: Modified in fd262e028de9b2bb81082408b1a7c12406dcaa56

internal/controllers/bundledeployment/bundledeployment.go
varshaprasad96: Right now:
- Unpacking contents from source, and storing it in Fs - Handled by TypeHasUnpacked
- Load and Process the bundle contents - TypeHasValidBundle
- Installation and Health - TypeInstall and TypeHealthy.

`TypeHasValidBundle` seems ambiguous. Probably, `TypeHasProcessed`, or `HasValidBundleContents` would be better?

internal/controllers/bundledeployment/interfaces.go
varshaprasad96: Modified in https://github.com/operator-framework/rukpak/commit/fd262e028de9b2bb81082408b1a7c12406dcaa56

internal/controllers/bundledeployment/bundledeployment.go
joelanford: In both of those cases, we end up returning the finalized BD anyway though. It seems like the only thing this approach "buys" us (if it's even a benefit) is that we only use the updated finalizers and status. If the finalizer implementation changes anything else, we would ignore it with this approach.

We handle this a bit differently in operator-controller. Perhaps we copy what it does? (i.e. let code change whatever it wants, and then assert/panic if unexpected fields changed): https://github.com/operator-framework/operator-controller/blob/main/internal/controllers/clusterextension_controller.go#L87-L89

api/v1alpha2/bundledeployment_types.go
```diff
@@ -52,40 +52,29 @@ type BundleDeploymentSpec struct {
...
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
 //+kubebuilder:resource:scope=Cluster,shortName={"bd","bds"}
-//+kubebuilder:printcolumn:name="Active Bundle",type=string,JSONPath=`.status.activeBundle`
 //+kubebuilder:printcolumn:name="Install State",type=string,JSONPath=`.status.conditions[?(.type=="Installed")].reason`
 //+kubebuilder:printcolumn:name=Age,type=date,JSONPath=`.metadata.creationTimestamp`
 //+kubebuilder:printcolumn:name=Provisioner,type=string,JSONPath=`.spec.provisionerClassName`,priority=1
 
-// BundleDeployment is the Schema for the bundledeployments API
+// BundleDeployment is the Schema for the bundleDeployments API
```
joelanford, line 77: When we're talking about the API and using the plural resource name, I'd suggest leaving it all lower-case to match the resource name in the API.
```suggestion
// BundleDeployment is the Schema for the bundledeployments API
```

cmd/core/main.go
```diff
@@ -231,44 +218,32 @@ func main() {
...
-		os.Exit(1)
-	}
-
-	if err := bundle.SetupWithManager(mgr, systemNsCluster.GetCache(), systemNamespace, append(
-		commonBundleProvisionerOptions,
-		bundle.WithProvisionerID(registry.ProvisionerID),
-		bundle.WithHandler(bundle.HandlerFunc(registry.HandleBundle)),
+	if err := bundledeployment.SetupWithManager(mgr, systemNsCluster.GetCache(), systemNamespace, append(
+		commonBDProvisionerOptions,
+		bundledeployment.WithUnpacker(unpacker),
```
joelanford, line 232: The unpacker is shared between the plain and registry BD controllers, right? Move this to the `commonBDProvisionerOptions` slice?

cmd/helm/main.go
```diff
@@ -205,35 +203,22 @@ func main() {
...
-	)...); err != nil {
-		setupLog.Error(err, "unable to create controller", "controller", rukpakv1alpha1.BundleKind, "provisionerID", helm.ProvisionerID)
-		os.Exit(1)
-	}
-
-	if err := bundledeployment.SetupWithManager(mgr, append(
+	if err := bundledeployment.SetupWithManager(mgr, systemNsCluster.GetCache(), systemNamespace, append(
 		commonBDProvisionerOptions,
 		bundledeployment.WithProvisionerID(helm.ProvisionerID),
+		bundledeployment.WithUnpacker(unpacker),
```
joelanford, line 218: Same question here that I asked in `core`: put this in the common options slice?

internal/controllers/bundledeployment/bundledeployment.go
```diff
@@ -198,64 +230,85 @@ func (c *controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	unpackResult, err := c.unpacker.Unpack(ctx, bd)
+	if err != nil {
+		return ctrl.Result{}, updateStatusUnpackFailing(&bd.Status, fmt.Errorf("source bundle content: %v", err))
+	}
 
-	bundleFS, err := c.storage.Load(ctx, bundle)
+	switch unpackResult.State {
+	case unpackersource.StatePending:
+		updateStatusUnpackPending(&bd.Status, unpackResult)
+		// There must a limit to number of retries if status is stuck at
```
joelanford, line 273: Remove this comment as well. Is this a holdover TODO from when the PR had requeueAfter?

internal/controllers/bundledeployment/bundledeployment.go
```diff
@@ -198,64 +230,85 @@ func (c *controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
 			Message: err.Error(),
 		})
 		return ctrl.Result{}, err
 	}
 
 	chrt, values, err := c.handler.Handle(ctx, bundleFS, bd)
 	if err != nil {
+		// Should be a better reason than "TypeHasValidBundle", probably "TypeHasPorcessedBundle"?
+		// Cannot use HasUnpacked condition again, as it is being set to true when the contents
+		// are fetched from source and stored in Fs.
```
joelanford, line 307: What if we just fold this into the "install" process from the status conditions standpoint and say that if `Handle` fails, that means installation failed?

We can always come back and add another condition later to further disambiguate if we need to.

joelanford, line 307: What if we just fold this into the "install" process from the status conditions standpoint and say that if `Handle` fails, that means installation failed?

We can always come back and add another condition later to further disambiguate if we need to.

internal/provisioner/plain/plain.go
```diff
@@ -24,15 +26,25 @@ const (
...
+		return nil, nil, err
+	}
+
+	processedFs := fsys
+	if isregistryV1 {
+		processedFs, err = convert.RegistryV1ToPlain(fsys)
+		if err != nil {
+			return nil, nil, fmt.Errorf("convert registry+v1 bundle to plain+v0 bundle: %v", err)
+		}
+	}
```
joelanford, line 41: Ah, I see. This answers my question about whether we need a separate handler. But this raises another question.

If we have the exact same logic between registry+v1 and plain provisioners, do we need both provisioners?

internal/provisioner/plain/plain.go
```diff
@@ -50,6 +62,15 @@ func ValidateBundle(fsys fs.FS) error {
 	return nil
 }
 
+func isRegistryV1Bundle(bd *rukpakv1alpha2.BundleDeployment) (bool, error) {
+	// Should not happen as it would have been as a nil pointer error earlier in the
+	// reconciler.
+	if bd == nil {
+		return false, fmt.Errorf("unable to find bundle format through provisioner")
+	}
+	return bd.Spec.ProvisionerClassName == registryv1.ProvisionerID, nil
```
joelanford, line 71: It's a bit confusing, IMO, to have a "plain" provisioner that handles multiple provisioner IDs and switches its logic based on that.

I think we should do one of the following:
1. Have two separate handler functions: one for `plain` and one for `registry` (and they can share any shared logic via a library if necessary)
2. Have a single handler with a single provisioner ID that introspects the bundle contents and if it looks like a registry+v1 bundle, just treat it that way. Otherwise, assume `plain`.

internal/storage/http.go
joelanford, line 1: The primary purpose of the `http` storage was for BD provisioners to be able to get the contents of bundles from bundle provisioners.

The secondary purpose was to be able to more generally serve requests that other clients would make (using the `status.contentURL`.

The primary purpose is longer present with the removal of the Bundle API/provisioner concept. And it makes me wonder if the secondary purpose is important enough to keep all of this HTTP serving of bundle content around.

I think I lean toward removing it. (Which would incidentally mean that we _could_ fully remove `rukpakctl`).

internal/util/adopt.go
joelanford, line 1: Is this file/function used anywhere anymore? I think it was only used by `alpha bootstrap`. If not, should we delete it?

internal/util/util.go
```diff
@@ -260,56 +143,15 @@ func MapConfigMapToBundlesHandler(ctx context.Context, cl client.Client, configM
 // GetBundlesForBundleDeploymentSelector is responsible for returning a list of
 // Bundle resource that exist on cluster that match the label selector specified
 // in the BD parameter's spec.Selector field.
-func GetBundlesForBundleDeploymentSelector(ctx context.Context, c client.Client, bd *rukpakv1alpha1.BundleDeployment) (*rukpakv1alpha1.BundleList, error) {
+func GetBundlesForBundleDeploymentSelector(ctx context.Context, c client.Client, bd *rukpakv1alpha2.BundleDeployment) (*rukpakv1alpha2.BundleDeploymentList, error) {
```
joelanford, line 146: Do we still use or need this function?

internal/util/util.go
```diff
@@ -335,9 +177,9 @@ func GenerateBundleName(bdName, hash string) string {
 	return fmt.Sprintf("%s-%s", bdName, hash)
 }
 
-// SortBundlesByCreation sorts a BundleList's items by it's
+// SortBundleDeploymentsByCreation sorts a BundleDeploymentList's items by its
 // metadata.CreationTimestamp value.
-func SortBundlesByCreation(bundles *rukpakv1alpha1.BundleList) {
+func SortBundleDeploymentsByCreation(bundles *rukpakv1alpha2.BundleDeploymentList) {
```
joelanford, line 182: Similar question as above. This function was used in the context of multiple `Bundles` existing for a BD. Now that we no longer have a Bundle API, I wonder if we still need/use this function?

manifests/base/apis/crds/patches/bundledeployment_validation.yaml
```diff
@@ -1,10 +1,33 @@
...
+  path: /spec/versions/0/schema/openAPIV3Schema/properties/spec/properties/source/oneOf
+  value:
+  - required:
+    - git
+  - required:
+    - image
+  - required:
+    - configMaps
+  - required:
+    - upload
```
joelanford, line 20: Remove the `upload` source.

internal/controllers/bundledeployment/bundledeployment.go
varshaprasad96: I think this still holds valid. If there is an image pull error with the unpack pod, we set the status to unpack pending and keep trying. Ideally, shouldn't we have a limit to number of retires when we are sure that a reconcile in same state is not going to change anything?

.goreleaser.yml
```diff
@@ -63,17 +63,6 @@ builds:
     - arm64
     - ppc64le
     - s390x
-  - id: rukpakctl
```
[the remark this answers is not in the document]
varshaprasad96: Removed the entire command as per our earlier conversation.

internal/provisioner/plain/plain.go
varshaprasad96: Using option (1) for now, to make less user-facing changes.

internal/util/util.go
varshaprasad96: My bad, missed cleaning this up. Wasn't being used anywhere.

internal/util/util.go
varshaprasad96: Missed cleaning this up.

internal/controllers/bundledeployment/bundledeployment.go
varshaprasad96: Looks like they are not required. Fixed.

internal/controllers/bundledeployment/bundledeployment.go
```diff
@@ -105,14 +122,14 @@ func SetupWithManager(mgr manager.Manager, opts ...Option) error {
...
 	controller, err := ctrl.NewControllerManagedBy(mgr).
 		Named(controllerName).
-		For(&rukpakv1alpha1.BundleDeployment{}, builder.WithPredicates(
+		For(&rukpakv1alpha2.BundleDeployment{}, builder.WithPredicates(
 			util.BundleDeploymentProvisionerFilter(c.provisionerID)),
 		).
-		Watches(&source.Kind{Type: &rukpakv1alpha1.Bundle{}}, handler.EnqueueRequestsFromMapFunc(
-			util.MapBundleToBundleDeploymentHandler(context.Background(), mgr.GetClient(), c.provisionerID)),
-		).
+		Watches(source.NewKindWithCache(&corev1.Pod{}, systemNsCache), util.MapOwneeToOwnerProvisionerHandler(context.Background(), mgr.GetClient(), l, c.provisionerID, &rukpakv1alpha2.BundleDeployment{})).
```
ncdc, line 131: Not sure if you want to fix this here/now, but we shouldn't be passing in context.Background. We have some options, but they might be worth doing separately (?)

internal/controllers/bundledeployment/bundledeployment.go
```diff
@@ -151,15 +174,23 @@ type controller struct {
...
 
+//+kubebuilder:rbac:groups=core.rukpak.io,resources=bundledeployments/finalizers,verbs=update
 //+kubebuilder:rbac:groups=core.rukpak.io,resources=bundledeployments,verbs=list;watch
 //+kubebuilder:rbac:groups=core.rukpak.io,resources=bundledeployments/status,verbs=update;patch
-//+kubebuilder:rbac:groups=core.rukpak.io,resources=bundledeployments/finalizers,verbs=update
+//+kubebuilder:rbac:verbs=get,urls=/bundles/*;/uploads/*
+//+kubebuilder:rbac:groups=core,resources=pods,verbs=list;watch;create;delete
+//+kubebuilder:rbac:groups=core,resources=configmaps,verbs=list;watch
+//+kubebuilder:rbac:groups=core,resources=pods/log,verbs=get
 //+kubebuilder:rbac:groups=*,resources=*,verbs=*
```
ncdc, line 191: 😄 (I realize this wasn't a change from this PR)

internal/controllers/bundledeployment/bundledeployment.go
varshaprasad96: I can create a separate issue to handle in a follow up. 

Just curious, why not pass the same context? Isn't it helpful to pass deadlines, cancellations and any request scoping information across API calls?

internal/controllers/bundledeployment/bundledeployment.go
varshaprasad96: Do you mean: `//+kubebuilder:rbac:groups=*,resources=*,verbs=*`?

I guess this was to be able to handle any bundle objects. Not sure if we need wildcard for verbs, but can create an issue to discuss more on handling this permission scoping.

internal/controllers/bundledeployment/bundledeployment.go
ncdc: context.Background() never gets canceled, so passing it in here is a no-op.

A map function ideally runs and returns quickly. When using typed generated cache listers, you can traverse from the watch event to whatever you want from the listers without needing a context. But with controller-runtime's Client, you need to pass in a context (and you don't know if you're reading from the cache or using a live client unless you trace the creation of the client you're using).

Let's say instead of context.Background() here, you passed in some cancelable context. Because this call is creating a map function, whatever context you pass in is shared by all invocations of the map function. You can't easily cancel one without canceling all of them (although it's not really clear if we'd want to cancel just one).

varshaprasad96: Thanks for explaining. Have created https://github.com/operator-framework/rukpak/issues/799 for follow up.

internal/controllers/bundledeployment/bundledeployment.go
varshaprasad96: https://github.com/operator-framework/rukpak/issues/800 for follow up.

internal/controllers/bundledeployment/bundledeployment.go
joelanford: This will be fixed when we upgrade to controller-runtime v0.16.3.

ncdc (approved):

PR #794, opened 2024-01-12
:seedling: bump k8s, helm-operator-plugins, of/operator-registry, and of/api

joelanford:


go.mod
```diff
@@ -11,121 +11,126 @@ require (
...
-	k8s.io/kube-aggregator v0.26.10
-	k8s.io/utils v0.0.0-20230220204549-a5ecb0141aa5
+	helm.sh/helm/v3 v3.13.3
+	k8s.io/api v0.28.5
+	k8s.io/apiextensions-apiserver v0.28.5
+	k8s.io/apimachinery v0.28.5
+	k8s.io/cli-runtime v0.28.5
+	k8s.io/client-go v0.28.5
+	k8s.io/component-base v0.28.5
+	k8s.io/kube-aggregator v0.28.5
```
everettraven, line 29: We should probably bump the KinD version to v0.20.0 and use a 1.28 based node image to ensure we are testing against a 1.28 cluster in e2e

joelanford: sgtm

joelanford: 🤔 Something about this breaks the e2e test where we pull from a private registry. Investigating.

joelanford:
Fixes #799

go.mod
joelanford: I'll try this again after I rebase, but I'm tempted to separate this out into a separate PR if I can't get it working immediately.

joelanford: https://github.com/operator-framework/rukpak/issues/803

cmd/crdvalidator/handlers/crd.go
```diff
@@ -74,12 +75,6 @@ func (cv *CrdValidator) Handle(ctx context.Context, req admission.Request) admis
 	return admission.Allowed("")
 }
 
-// InjectDecoder injects a decoder for the CrdValidator.
-func (cv *CrdValidator) InjectDecoder(d *admission.Decoder) error {
-	cv.decoder = d
-	return nil
-}
-
```
everettraven, line 82: Any particular reason for this removal? I noticed that this made the go-apidiff check unhappy

.bingo/variables.env
```diff
@@ -12,7 +12,7 @@ BINGO="${GOBIN}/bingo-v0.8.0"
 
 CONTROLLER_GEN="${GOBIN}/controller-gen-v0.12.0"
```
varshaprasad96, line 13: Do we also want to bump this to 0.13.0: https://github.com/kubernetes-sigs/controller-tools/blob/v0.13.0/go.mod

everettraven (approved):
Aside from the go-apidiff failure this LGTM. If that is expected then 👍

varshaprasad96 (commented):
Just one question regarding controller-gen bump.
Rest looks good to me, /lgtm from my end

.bingo/variables.env
joelanford: I'll do this in a follow-up. Thanks for pointing it out!

cmd/crdvalidator/handlers/crd.go
joelanford: That was the old way controller-runtime could inject a Decoder. That interface was removed, and the function is no longer useful.

I updated the code to set the decoder in the constructor function instead.

PR #802, opened 2024-01-17
✨  Support for other Install strategies

varshaprasad96:
Follow up of #772 

This PR removes the requirement that AllNamespace mode should be supported always in a registry+v1 format.

It does so by:

1. Adding a field to specify watch namespaces in BundleDeployment Spec.
2. While converting from registryV1 to plain bundle, we now generate roles for all specified targetNamespaces which are being watched.


everettraven (commented):
I think this looks good at a glance, but I do have a question - do we currently have a way to communicate the set of namespaces to watch via the deployment?

internal/convert/registryv1.go
```diff
@@ -274,15 +267,18 @@ func Convert(in RegistryV1, installNamespace string, targetNamespaces []string)
...
-			return nil, err
+	for _, ns := range targetNamespaces {
+		for _, permission := range permissions {
+			saName := saNameOrDefault(permission.ServiceAccountName)
+			name, err := generateName(fmt.Sprintf("%s-%s", in.CSV.Name, saName), permission)
+			if err != nil {
+				return nil, err
+			}
+			roles = append(roles, newRole(ns, name, permission.Rules))
+			roleBindings = append(roleBindings, newRoleBinding(ns, name, name, installNamespace, saName))
```
everettraven, line 278: What is expected to happen if the specified namespace set is all namespaces (i.e `""` or `"*"`)?

joelanford: `"*"` is not a valid namespace name, and (correct me if I'm wrong) I don't think there is any existing code in OLM or Kubernetes that treats `"*"` as all namespaces?

If `targetNamespaces == []string{""}`, that has typically meant "please watch all namespaces"

Separate but related question: what if `targetNamespaces == []string{"", "foo"}`? Perhaps we should have a CRD schema validation that forbids empty string in this list, and document that empty/null list is the way to get "all namespaces".

everettraven: > `"*"` is not a valid namespace name, and (correct me if I'm wrong) I don't think there is any existing code in OLM or Kubernetes that treats `"*"` as all namespaces?
> 

I don't think there is - I've seen "*" used before in RBAC to mean "all apis" within a group, but I don't recall ever seeing it used in the namespace context so I'm happy to concede that I'm wrong with that being a valid "all namespaces" notation.

> If `targetNamespaces == []string{""}`, that has typically meant "please watch all namespaces"
> 

So then in this case we should do what?
- List all namespaces and create `Role`+`RoleBindings`?
- Escalate all permissions to `ClusterRole` + `ClusterRoleBindings`?
- Something else?

My reason for calling this out is that I think if `""` is specified the role+rolebinding creation in this logic will fail.

> Separate but related question: what if `targetNamespaces == []string{"", "foo"}`? Perhaps we should have a CRD schema validation that forbids empty string in this list, and document that empty/null list is the way to get "all namespaces".

+1 to validating that `""` doesn't appear in a list of specific namespaces

varshaprasad96: As Joe mentioned, we do have validation in place that checks if `len(targetNs) == 1` and if it is "", if so we set to AllNamespace mode and elevate the Roles and RoleBindings to ClusterRoles and ClusterRoleBindings respectively. 

https://github.com/operator-framework/rukpak/blob/a990f15b67ffc35453801e814979b91a8ebab715/internal/convert/registryv1.go#L265

> what if targetNamespaces == []string{"", "foo"}

In this case our validation would check if MultiNamespace mode is supported or not. If not, it would error - https://github.com/operator-framework/rukpak/blob/a990f15b67ffc35453801e814979b91a8ebab715/internal/convert/registryv1.go#L161. But a corner case is:
- Multinamespace mode is set to true
- And target ns has a list with "" in it
Will add a validation as discussed to ensure - if len(targetNs) > 1, "" shouldn't appear in the list. 

On that note, just to confirm can MultiNamespace be set to true, and the targetNs list just have a single value? 
Currently, the answer is no. Because in that case, we expect SingleNs to be set instead. I am not aware of the reasoning behind differentiating single and multiNs in this context, but just checking if that's how we are handling in v0.

everettraven: >As Joe mentioned, we do have validation in place that checks if len(targetNs) == 1 and if it is "", we set to AllNamespace mode and elevate the Roles and RoleBindings to ClusterRoles and ClusterRoleBindings respectively.

Ah, I see now - thanks for providing a link! Even though I expanded the code to look for this I guess I didn't expand it enough and missed it 🤦

varshaprasad96:
> do we currently have a way to communicate the set of namespaces to watch via the deployment?

Which deployment are you referring to? This PR introduces a spec field where watchnamespaces can specified while creating a BD ([ref](https://github.com/operator-framework/rukpak/pull/802/files#diff-a1c6e6920d8eaf59ba7a13cf6105541e11849da6eeb1a5a8ea3db999cbb2bcfaR62)). It would then be propagated to cluster extension.

internal/convert/registryv1.go
joelanford: Just to clarify, I'm suggesting that we use OpenAPI schema validation in the CRD to outright prevent `""` in the list.

api/v1alpha2/bundledeployment_types.go
```diff
@@ -57,6 +57,9 @@ type BundleDeploymentSpec struct {
 	// Config is provisioner specific configurations
 	// +kubebuilder:pruning:PreserveUnknownFields
 	Config runtime.RawExtension `json:"config,omitempty"`
+	// +kubebuilder:validation:Optional
+	// watchNamespaces indicates which namespaces the operator should watch.
+	WatchNamespaces []string `json:"watchNamespaces,omitempty"`
```
everettraven, line 62: I think we can create some CEL validation logic here to prevent the `""` value existing in this set using the `// +kubebuilder:validation:XValidation` marker, but I do wonder if we could get away with a combination of the other existing validation markers to ensure the list is unique and make sure that `""` isn't included in the list

internal/convert/registryv1.go
```diff
@@ -158,17 +158,13 @@ func validateTargetNamespaces(supportedInstallModes sets.Set[string], installNam
 			return nil
 		}
 	default:
-		if supportedInstallModes.Has(string(v1alpha1.InstallModeTypeMultiNamespace)) {
+		if supportedInstallModes.Has(string(v1alpha1.InstallModeTypeMultiNamespace)) && !set.Has("") {
```
everettraven, line 161: If we aren't going to support the `""` value at all, we should fail fast with validation logic when attempting to create the resource.

internal/convert/registryv1.go
```diff
@@ -40,7 +40,7 @@ type Plain struct {
 	Objects []client.Object
 }
 
-func RegistryV1ToPlain(rv1 fs.FS) (fs.FS, error) {
+func RegistryV1ToPlain(rv1 fs.FS, watchnamespaces []string) (fs.FS, error) {
```
ncdc, line 43:
```suggestion
func RegistryV1ToPlain(rv1 fs.FS, watchNamespaces []string) (fs.FS, error) {
```

internal/convert/registryv1.go
varshaprasad96: We do support targetNs with a single empty string element to handle watching all Ns. The CEL validation which has been added should along with this check should take care of all the cases.

api/v1alpha2/bundledeployment_types.go
varshaprasad96: Fixed with the recent changes.

internal/convert/registryv1.go
varshaprasad96: > Just to clarify, I'm suggesting that we use OpenAPI schema validation in the CRD to outright prevent "" in the list.

Fixed.

everettraven (approved):

internal/convert/registryv1.go
```diff
@@ -189,10 +185,7 @@ func Convert(in RegistryV1, installNamespace string, targetNamespaces []string)
...
 		}
 	}
-	if !supportedInstallModes.Has(string(v1alpha1.InstallModeTypeAllNamespaces)) {
-		return nil, fmt.Errorf("AllNamespace install mode must be enabled")
-	}
-	if targetNamespaces == nil {
+	if len(targetNamespaces) == 0 {
 		if supportedInstallModes.Has(string(v1alpha1.InstallModeTypeAllNamespaces)) {
 			targetNamespaces = []string{""}
 		} else if supportedInstallModes.Has(string(v1alpha1.InstallModeTypeOwnNamespace)) {
```
joelanford, line 191: Should we start requiring that for `OwnNamespace`, the targetNamespace must be explicitly specified?

varshaprasad96: I'm fine with it, will it break any existing norm that v0 has set in terms of if `OwnNamespace` is set, we watch installNs.

joelanford: In v0, the OperatorGroup explicitly lists (or matches) just the namespace that it is in, so it is explicit in that sense.

The reason this logic exists in rukpak is because (until now) there is no way to define a list of target namespaces, so they only possibilities we could even derive were AllNamespaces or OwnNamespace.

varshaprasad96: Leaving it as is for now based on the [slack](https://kubernetes.slack.com/archives/C0181L6JYQ2/p1705688983270469?thread_ts=1705669480.907989&cid=C0181L6JYQ2) discussion.

PR #816, opened 2024-01-22
:sparkles: Cleanups after removing BundleAPI

varshaprasad96:
This PR makes the follwing changes:
1. Fixes the reference to v1alpha1 APIs in the linter.
2. TypeHasValidBundle was being set to false only when the bundle load errored. Switching the condition to be TypeInstalled instead.
3. Fix documentation on provisioners.

Follow up #774

### DIFF
--- a/docs/provisioners/helm.md
+++ b/docs/provisioners/helm.md
@@ -3,7 +3,7 @@
 ## Summary
 
 The `helm` provisioner is one of the [provisioners](https://github.com/operator-framework/rukpak/tree/main/internal/provisioner) of RukPak.
-It is able to instantiate a given `helm+v3` bundle with a specified helm chart onto a cluster where it then installs the content. It does so by reconciling `Bundle` and `BundleDeployment` types that have
+It is able to instantiate a given `helm+v3` bundle with a specified helm chart onto a cluster where it then installs the content. It does so by reconciling `BundleDeployment` types that have
 the `spec.provisionerClassName` field set to `core-rukpak-io-helm`. This field must be set to the correct provisioner
 name in order for the `helm` provisioner to see and interact with the bundle.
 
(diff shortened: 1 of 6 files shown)


docs/provisioners/helm.md
```diff
@@ -166,18 +149,18 @@ EOF
...
+kubectl get bundledeployment -l app=my-ahoy
 ```
 
-Eventually the Bundle should show up as Unpacked:
+Eventually the BundleDeployment should show up as Unpacked:
 
 ```console
-$ kubectl get bundle -l app=my-ahoy
-NAME                 TYPE   PHASE      AGE
-my-ahoy-5764594dc8   git    Unpacked   33s
```
joelanford, line 180: The bundle deployment column output doesn't have a PHASE column, right? And I would assume that basically all BDs would move on from "Unpacked" to either a successful or failed installation fairly quickly. So this area might need a bit more tweaking.

docs/provisioners/plain.md
```diff
@@ -36,45 +36,29 @@ The `plain` provisioner will unpack the provided Bundle onto the cluster, and ev
...
 Then eventually, as the bundle content is unpacked onto the cluster via the defined storage mechanism, the bundle status
 will be updated to Unpacked, indicating that all its contents have been stored on-cluster.
 
 ```console
-$ kubectl get bundle my-bundle
-NAME           TYPE    PHASE      AGE
-my-bundle      image   Unpacked   10s
+$ kubectl get bundledeployment my-bundle
+NAME                    TYPE    PHASE      AGE
+my-bundle-deployment    image   Unpacked   10s
```
joelanford, line 58: Same comment here as above.

docs/provisioners/helm.md
varshaprasad96: My bad, that's right. It only has install state. Tweaked this part a bit for all 3 provisioners.

PR #823, opened 2024-02-15
Bump helm to 3.14.1

kevinrizza:
Also bumps kubernetes to 1.29 and golang to 1.21


go.mod
```diff
@@ -18,15 +20,15 @@ require (
...
+	k8s.io/apiextensions-apiserver v0.29.0
+	k8s.io/apimachinery v0.29.0
+	k8s.io/cli-runtime v0.29.0
+	k8s.io/client-go v0.29.0
+	k8s.io/component-base v0.29.0
 	k8s.io/kube-aggregator v0.28.5
-	k8s.io/utils v0.0.0-20230505201702-9f6742963106
+	k8s.io/utils v0.0.0-20230726121419-3b25d923346b
 	sigs.k8s.io/cli-utils v0.35.0
 	sigs.k8s.io/controller-runtime v0.16.3
```
varshaprasad96, line 33: Can we also bump controller-runtime to 0.17.1 - given we are bumping other k8s deps here?

go.mod
```diff
@@ -18,15 +20,15 @@ require (
...
 	golang.org/x/crypto v0.18.0
 	golang.org/x/sync v0.6.0
-	helm.sh/helm/v3 v3.13.3
-	k8s.io/api v0.28.5
-	k8s.io/apiextensions-apiserver v0.28.5
-	k8s.io/apimachinery v0.28.5
-	k8s.io/cli-runtime v0.28.5
-	k8s.io/client-go v0.28.5
-	k8s.io/component-base v0.28.5
+	helm.sh/helm/v3 v3.14.1
```
varshaprasad96, line 23: We also need helm version in helm-operator-plugins repo bumped.

m1kola:
@kevinrizza I haven't looked into this properly, but I have a feeling that the e2e failures are legit. Most likely we assert (for whatever reason) error messages which got changed.

go.mod
m1kola: There is https://github.com/operator-framework/helm-operator-plugins/pull/301 which still needs a bit of work. I'll take over from Kevin.

go.mod
m1kola: Bumped to v0.17.2

m1kola:
I'm taking over this PR (with Kevin's permissions). I updated the branch and I think we should be good now.

test/e2e/plain_provisioner_test.go
```diff
@@ -1337,7 +1337,7 @@ var _ = Describe("plain provisioner bundle", func() {
 					WithTransform(func(c *metav1.Condition) string { return c.Type }, Equal(rukpakv1alpha2.TypeInstalled)),
 					WithTransform(func(c *metav1.Condition) metav1.ConditionStatus { return c.Status }, Equal(metav1.ConditionFalse)),
 					WithTransform(func(c *metav1.Condition) string { return c.Reason }, Equal(rukpakv1alpha2.ReasonInstallFailed)),
-					WithTransform(func(c *metav1.Condition) string { return c.Message }, ContainSubstring(`the server could not find the requested resource`)),
+					WithTransform(func(c *metav1.Condition) string { return c.Message }, ContainSubstring(`no matches for kind "CatalogSource" in version "operators.coreos.com/v1alpha1"`)),
```
varshaprasad96, line 1340: Though `no matches for the kind` and `could not find the resources` return the same HTTP error code, I believe the former means that the CRD is not installed available in the cluster (API server does not know the definition), whereas the latter indicates that the specific resource endpoint does not exist. Was there a change in k8s because of which this change was needed? Just curious.

varshaprasad96 (approved):
The changes look good, and all the tests pass. 
/lgtm
/approve. 

Just a small question regarding the change in error.

PR #848, opened 2024-03-27
bump kind to 0.22.0

joelanford:


.bingo/Variables.mk
```diff
@@ -1,4 +1,4 @@
-# Auto generated binary variables helper managed by https://github.com/bwplotka/bingo v0.8. DO NOT EDIT.
+# Auto generated binary variables helper managed by https://github.com/bwplotka/bingo v0.9. DO NOT EDIT.
```
m1kola, line 1: It looks like you use bingo 0.9, but our pinned version below is 0.8. I'm ok if we merge this as is because it doesn't have any impact (other than generated comment updates).

m1kola (approved):

everettraven (approved):

PR #864, opened 2024-04-25
BundleDeployment: add and implement required installNamespace field

joelanford:
In order to make it possible to add an `InstallNamespace` field in the operator-controller project's `ClusterExtension` spec, we need to add it here first (because the embedding of rukpak logic in operator-controller is still ongoing).

The logic here configures the Helm client to use the InstallNamespace as the context for the templating engine and k8s client, the result being similar to using `helm --namespace <installNamespace>`. The exception is that rukpak continues to store the helm release secret in its system namespace.

This commit pins `rukpak` to an untagged commit on `main` in the `helm-operator-plugins` repo. I can tag that commit if reviewers want to stick with a tagged release for helm-operator-plugins.


internal/provisioner/registry/registry.go
```diff
@@ -19,7 +20,7 @@ const (
 )
 
 func HandleBundleDeployment(ctx context.Context, fsys fs.FS, bd *rukpakv1alpha2.BundleDeployment) (*chart.Chart, chartutil.Values, error) {
-	plainFS, err := convert.RegistryV1ToPlain(fsys, bd.Spec.WatchNamespaces)
+	plainFS, err := convert.RegistryV1ToPlain(fsys, bd.Spec.InstallNamespace, []string{metav1.NamespaceAll})
```
varshaprasad96, line 23: I assume removal of `watchNamespace` field would be a follow up?

varshaprasad96 (approved):
Just questions, overall the changes look good to me!
/lgtm
/approve

perdasilva (approved):
lgtm

internal/provisioner/registry/registry.go
joelanford: You mean the `RegistryV1ToPlain` function accepting a `watchNamespaces` parameter? I left that in intentionally. I assume we'll still need all of that underlying logic in order to support all install modes via the Extension API.

PR #873, opened 2024-05-01
Add a CRD Upgrade Safety preflight check to the `BundleDeployment` controller

everettraven:
**Description**:
- Adds a new `preflights/crdupgradesafety` package containing the logic for a CRD Upgrade Safety preflight check
- Updates the `BundleDeployment` controller to be able to be configured with preflight checks and runs them if they are configured


cmd/core/main.go
```diff
@@ -226,11 +228,23 @@ func main() {
 		setupLog.Error(err, "unable to create action client getter")
 		os.Exit(1)
 	}
+
+	aeClient, err := v1.NewForConfig(cfg)
```
joelanford, line 232: Nit
```suggestion
	aeClient, err := apiextensionsv1client.NewForConfig(cfg)
```

internal/preflights/crdupgradesafety/crdupgradesafety.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	relObjects, err := util.ManifestObjects(strings.NewReader(rel.Manifest), fmt.Sprintf("%s-release-manifest", rel.Name))
+	if err != nil {
+		return fmt.Errorf("parsing release %q objects: %w", rel.Name, err)
+	}
+
+	validateErrors := []error{}
+	for _, obj := range relObjects {
+		if obj.GetObjectKind().GroupVersionKind() == apiextensionsv1.SchemeGroupVersion.WithKind("CustomResourceDefinition") {
```
joelanford, line 72: Nit:
```suggestion
		if obj.GetObjectKind().GroupVersionKind() != apiextensionsv1.SchemeGroupVersion.WithKind("CustomResourceDefinition") {
			continue
		}
		// outdent the rest of the logic
```

internal/preflights/crdupgradesafety/crdupgradesafety.go
```diff
@@ -0,0 +1,100 @@
...
+				validateErrors = append(validateErrors, fmt.Errorf("validating upgrade for CRD %q failed: %w", newCrd.Name, err))
+			}
+		}
+	}
+
+	if len(validateErrors) > 0 {
+		return errors.Join(validateErrors...)
+	}
+
+	return nil
```
joelanford, line 99: Pretty sure this returns `nil` if there are no non-nil errors in the slice.
```suggestion
	return errors.Join(validateErrors...)
```

everettraven: updated in https://github.com/operator-framework/rukpak/pull/873/commits/3d4d5b5c90e98dceb74874d3796c0028167c4327

internal/preflights/crdupgradesafety/crdupgradesafety.go
everettraven: Updated in https://github.com/operator-framework/rukpak/pull/873/commits/3d4d5b5c90e98dceb74874d3796c0028167c4327

cmd/core/main.go
everettraven: updated in https://github.com/operator-framework/rukpak/pull/873/commits/3d4d5b5c90e98dceb74874d3796c0028167c4327

go.mod
```diff
@@ -1,10 +1,11 @@
 module github.com/operator-framework/rukpak
 
-go 1.21
+go 1.22
```
grokspawn, line 3: Is there a technical reason for this to bump golang also?  I think we'd prefer to keep potentially disruptive dep bumps (for e.g. k8s/controller-runtime/kubebuilder) discrete from feature/bug PRs in general.

everettraven: The version of kapp we are pulling in as a dependency uses go 1.22

internal/controllers/bundledeployment/bundledeployment.go
```diff
@@ -310,12 +332,30 @@ func (c *controller) reconcile(ctx context.Context, bd *rukpakv1alpha2.BundleDep
 		},
 	}
 
-	rel, state, err := c.getReleaseState(cl, bd, chrt, values, post)
+	// TODO: update this to return desired release as well
```
grokspawn, line 335: Is this todo going to be resolved before merging this?

everettraven: Ah, I think that should have already been addressed... I'll double check

go.mod
grokspawn: Ah.  I see with 1.21++ it [became mandatory](https://go.dev/ref/mod#go-mod-file-go):
```
The go line must be greater than or equal to the go line of all dependencies.
```

Boo.

everettraven: Yeah, there are more preflight checks we could pull in but it resulted in k8s version bumps to v0.30.z and so I refrained from going further than this PR currently does.

It looks like the tilt stuff is failing on this go bump as well...

internal/controllers/bundledeployment/bundledeployment.go
```diff
@@ -101,6 +101,12 @@ func WithActionClientGetter(acg helmclient.ActionClientGetter) Option {
 	}
 }
 
+func WithPreflights(preflights ...Preflight) Option {
```
varshaprasad96, line 104: Nit: Is this required and present by default for now? If so, it would be nice to add a validation for this in `validateConfig` func, just to avoid programmer's error since we are calling `preflight.Install` directly.

internal/controllers/bundledeployment/bundledeployment.go
```diff
@@ -156,6 +162,21 @@ func (c *controller) validateConfig() error {
 	return utilerrors.NewAggregate(errs)
 }
 
+// Preflight is a check that should be run before making any changes to the cluster
+type Preflight interface {
+	// Install runs checks that should be successful prior
+	// to installing the Helm release. It is provided
+	// a Helm release and returns an error if the
+	// check is unsuccessful
+	Install(context.Context, *release.Release) error
```
varshaprasad96, line 171: Having preflight here looks good, and not blocking this PR at all. I remember this being mentioned somewhere earlier - but just to refresh, can this be directly integrated as a part of helm-operator-plugin's `ActionInterface`: https://github.com/operator-framework/helm-operator-plugins/blob/ae087c419483d73c37155c02f142ab14b4d8c706/pkg/client/actionclient.go#L55?

There may be tweaks required to get the existing release from the cluster and passing it to action client, but this thought popped up because we *may* probably have to expand checks for Uninstall to, and could be cleaner if the helm client can do it all together.

everettraven: I originally had a PR adding it to the actionclient here: https://github.com/operator-framework/helm-operator-plugins/pull/327

There was some discussion in slack where it was decided that it would be a better approach to place the logic directly in operator-controller/rukpak: https://kubernetes.slack.com/archives/C0181L6JYQ2/p1714509603408259

internal/controllers/bundledeployment/bundledeployment.go
everettraven: It is not required. If there are no preflight checks defined neither `preflight.Install()` nor `preflight.Upgrade()` should be called.

internal/controllers/bundledeployment/bundledeployment.go
varshaprasad96: Ah Sorry! Now I realize why I remembered it! :see_no_evil:

pkg/preflights/crdupgradesafety/crdupgradesafety.go
```diff
@@ -0,0 +1,99 @@
...
+		crdClient: crdCli,
+		// create a default validator. Can be overridden via the options
+		validator: &kappcus.Validator{
+			Validations: []kappcus.Validation{
+				kappcus.NewValidationFunc("NoScopeChange", kappcus.NoScopeChange),
+				kappcus.NewValidationFunc("NoStoredVersionRemoved", kappcus.NoStoredVersionRemoved),
+				kappcus.NewValidationFunc("NoExistingFieldRemoved", kappcus.NoExistingFieldRemoved),
+				&kappcus.ChangeValidator{
+					Validations: []kappcus.ChangeValidation{
+						kappcus.EnumChangeValidation,
```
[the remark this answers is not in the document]
everettraven: Two reasons:
- A release of kapp with these change validations has not been cut (we _could_ pin on a commit)
- requires a k8s bump (I am actively queuing these up so we can add the rest of them as part of this PR)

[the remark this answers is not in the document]
rashmigottipati: Ah that makes sense!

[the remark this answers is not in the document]
everettraven: All the ones present in kapp should be added now!

go.mod
everettraven: Rebased on main which already bumped go and k8s deps.

api/v1alpha2/bundledeployment_types.go
```diff
@@ -45,6 +45,7 @@ const (
 	ReasonReconcileFailed           = "ReconcileFailed"
 	ReasonUnhealthy                 = "Unhealthy"
 	ReasonUpgradeFailed             = "UpgradeFailed"
+	ReasonPreflightFailed           = "PreflightFailed"
```
joelanford, line 48: Do we need another reason, or can we just use `InstallFailed` and `UpgradeFailed`, etc?

internal/controllers/bundledeployment/bundledeployment.go
```diff
@@ -449,27 +489,35 @@ const (
...
+		return currentRelease, nil, stateError, err
 	}
 	if desiredRelease.Manifest != currentRelease.Manifest ||
 		currentRelease.Info.Status == release.StatusFailed ||
 		currentRelease.Info.Status == release.StatusSuperseded {
-		return currentRelease, stateNeedsUpgrade, nil
+		return currentRelease, desiredRelease, stateNeedsUpgrade, nil
 	}
-	return currentRelease, stateUnchanged, nil
+	return currentRelease, nil, stateUnchanged, nil
```
joelanford, line 520: Seems like it might make sense to return `desiredRelease` even in line 520. Is there a good reason not to?

And looking at 515-518, I wonder if we should do something like this instead:
```suggestion
	relState := stateUnchanged
	if desiredRelease.Manifest != currentRelease.Manifest ||
 		currentRelease.Info.Status == release.StatusFailed ||
 		currentRelease.Info.Status == release.StatusSuperseded {
 		relState = stateNeedsUpgrade
 	}
 	return currentRelease, desiredRelease, relState, nil
```

test/e2e/plain_provisioner_test.go
```diff
@@ -1202,9 +1202,9 @@ var _ = Describe("plain provisioner bundle", func() {
 						Not(BeNil()),
 						WithTransform(func(c *metav1.Condition) string { return c.Type }, Equal(rukpakv1alpha2.TypeInstalled)),
 						WithTransform(func(c *metav1.Condition) metav1.ConditionStatus { return c.Status }, Equal(metav1.ConditionFalse)),
-						WithTransform(func(c *metav1.Condition) string { return c.Reason }, Equal(rukpakv1alpha2.ReasonInstallFailed)),
+						WithTransform(func(c *metav1.Condition) string { return c.Reason }, Equal(rukpakv1alpha2.ReasonErrorGettingReleaseState)),
```
joelanford, line 1205: Is this happening because we're now doing a dry-run install that is failing?

api/v1alpha2/bundledeployment_types.go
everettraven: Good question! I think we could get away with just using `InstallFailed` and `UpgradeFailed`.

test/e2e/plain_provisioner_test.go
everettraven: Correct

internal/controllers/bundledeployment/bundledeployment.go
everettraven: >Seems like it might make sense to return desiredRelease even in line 520. Is there a good reason not to?

No particular reason. I was just trying to keep the impact to the logic pretty minimal and didn't think it would be worth returning the desired release in the case where it wouldn't be used, but that is probably better left to the caller to decide than the function. Will tweak this a bit.

api/v1alpha2/bundledeployment_types.go
everettraven: Updated in the latest push

internal/controllers/bundledeployment/bundledeployment.go
everettraven: updated in the latest push

joelanford (approved):

PR #874, opened 2024-05-02
update image source unpacking to use a direct image registry client

everettraven:
**Description**:
- This updates rukpak's image source unpack process to be similar to catalogd's by using a direct image registry client for pulling and unpacking bundle contents. Most of the logic was copied and adapted from catalogd's implementation and as such includes:
    - bundle unpack content caching (for identifying changes)
    - unpack content cache period garbage collection (in the event any BundleDeployment deletes were missed for cleaning up the content cache)


manifests/base/core/resources/deployment.yaml
```diff
@@ -73,3 +75,5 @@ spec:
       volumes:
         - name: bundle-cache
```
varshaprasad96, line 76: I'm missing the benefit of having `bundle-cache` and `unpack-cache` here, that contain the same bundle contents. Initially the use of `localdir` store was to copy the unpacked contents from pod log (for image source at least) and store locally. 

Disregarding other sources (which we are not looking to support) - the registry client can directly unpack contents at a single location that we look at for generating charts.

internal/garbagecollection/garbagecollector.go
```diff
@@ -0,0 +1,94 @@
...
+	Logger         logr.Logger
+	MetadataClient metadata.Interface
+	Interval       time.Duration
+}
+
+// Start will start the garbage collector. It will always run once on startup
+// and loop until context is canceled after an initial garbage collection run.
+// Garbage collection will run again every X amount of time, where X is the
+// supplied garbage collection interval.
+func (gc *GarbageCollector) Start(ctx context.Context) error {
```
varshaprasad96, line 38: What's the use of garbage collector running at specific internal of time? Can this be a part of finalizer, where all cached bundles belonging to a particular BD are deleted at once. 

Running a GC could turn out to be more expensive than caching a couple more bundles belonging to the same BD with different image digests. 

Referring back to my previous comment on having unpack and bundle cache - we can also consolidate them into one, if we don't do the GC and are able to pass on the right unpack location path to the handler to do the next steps in processing and creating helm chart.

manifests/base/core/resources/deployment.yaml
everettraven: I totally agree, but my intention behind this PR is to get the functionality in place and we can optimize in the future. catalogd is going to need the same updates in the future, so ideally we can spend some time updating this implementation to be something that can be used in both catalogd and operator-controller in the future.

I'm not against making the changes in this PR, but the scope of this PR is already quite large with the necessary e2e changes and bringing it to parity with catalogd is a good starting point IMO. We can always iterate and improve in the future.

internal/garbagecollection/garbagecollector.go
everettraven: It's been long enough since I wrote this in catalogd that I can't remember the exact reason why we needed to implement a garbage collector, so it being present in this PR is a result of trying to achieve functional parity with catalogd.

I do believe catalogd uses a finalizer to also ensure cleanup is performed but I missed adding that functionality in this PR.

I don't disagree that this might be unnecessary, but in the interest of ensuring we are making these optimizations everywhere they need to be I think we should:
- Leave the garbage collector in place for this PR
- After this PR merges, create an epic/issues to update this package to work for both catalogd and operator-controller use cases
- Make any necessary optimizations to the implementation (one location for changes, benefits multiple projects)

everettraven: Additionally, the garbage collector doesn't _need_ to be copied over into operator-controller if it isn't needed. The real intention for this PR is to get the direct image registry client functionality in place and able to be used in operator-controller via the `source` pkg. 

If all we care about is getting that code in I can strip all the other changes from this PR and only add the functionality for unpacking by talking directly to an image registry and we handle everything else in operator-controller as we see fit.

varshaprasad96: I'd suggest removing the garbage collector, and keeping the PR confined to adding direct image registry client with a finalizer to clean up bundle contents - which is the bare minimum requirement to unblock currently. GC or any other improvements to maintain common code b/w catalogd and o-c/rukpak can be handled in follow ups based on future discussions.

everettraven: Garbage collection removed and finalizer logic added in https://github.com/operator-framework/rukpak/pull/874/commits/59925ff9d2361a49b8a824da8b644e9f0365ce61

internal/garbagecollection/garbagecollector.go
```diff
@@ -0,0 +1,94 @@
...
+	"k8s.io/client-go/metadata"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+
+	"github.com/operator-framework/rukpak/api/v1alpha2"
+)
+
+var _ manager.Runnable = (*GarbageCollector)(nil)
+
+// GarbageCollector is an implementation of the manager.Runnable
+// interface for running garbage collection on the Catalog content
```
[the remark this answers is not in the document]
everettraven: Removed in https://github.com/operator-framework/rukpak/pull/874/commits/59925ff9d2361a49b8a824da8b644e9f0365ce61

testdata/bundles/plain-v0/dependent/build-push-e2e-bundle.sh
```diff
@@ -0,0 +1,75 @@
+#! /bin/bash
```
dtfranz, line 1: Instead of having these separate files sharing a lot of common scripting, can we instead just have one script with arguments and have each sub-folder contain the needed yaml?

everettraven: I originally had it as a single script that did this, but since some of the testdata has specific nuances for the tests there was an increase in the single script complexity that I felt would only continue to get more complex when/if new testdata is added. Some examples:
- Two different bundle formats, `registry+v1` and `plain+v0`, have different directory requirements. This means that a common script must account for this difference and as such introduces more complex logic to ensure there are no flakes when building each bundle.
- One of the `plain+v0` testdata bundles requires inclusion of subdirectories. this requires more complex configmap creation and volume mounting logic to ensure the subdirectory is copied over to the container running the build + push operations.
- Ability to push to a registry that requires authentication requires additional volume mounts and configuration of the container to successfully authenticate with the registry.

I felt that reducing the script complexity and increasing "repetition" is likely going to be more maintainable in the long run since additional test cases can get as custom as they need to and provide all the logic to ensure the image is appropriately built and pushed to the on cluster registry.

go.mod
```diff
@@ -5,9 +5,13 @@ go 1.21
 toolchain go1.21.0
 
 require (
+	github.com/containerd/containerd v1.7.15
 	github.com/go-git/go-billy/v5 v5.5.0
 	github.com/go-git/go-git/v5 v5.12.0
 	github.com/go-logr/logr v1.4.1
+	github.com/google/go-containerregistry v0.19.0
+	github.com/google/go-containerregistry/pkg/authn/k8schain v0.0.0-20240418155129-98dd3e91704f
+	github.com/google/go-containerregistry/pkg/authn/kubernetes v0.0.0-20240418155129-98dd3e91704f
```
acornett21, line 14: These imports are confusing, `authn` package is already in `go-containerregistry` why are we pulling in 2 different packages explicitly? Or is this just happening automatically because those directories have mod files?

Either way, I think we'd want to go go `0.19.1`, since that is out and latest.

everettraven: These were pulled automatically via `go mod tidy`

everettraven: Also, authn/k8schain and authn/kubernetes have their own mod files and don't have tags. These commits align with the version of go-containerregistry in the go.mod (which for this PR is 0.19.0). I can bump it to 0.19.1 if necessary

test/e2e/plain_provisioner_test.go
```diff
@@ -279,40 +246,14 @@ var _ = Describe("plain provisioner bundle", func() {
...
 				}
 				unpackPending := meta.FindStatusCondition(bundleDeployment.Status.Conditions, rukpakv1alpha2.PhaseUnpacked)
-				if unpackPending == nil {
-					return false
-				}
-				if unpackPending.Message != fmt.Sprintf(`Back-off pulling image "%s"`, bundleDeployment.Spec.Source.Image.Ref) {
-					return false
-				}
-				return true
+				return unpackPending != nil
```
joelanford, line 256: This test no longer seems to be testing what it says it is testing.  Before, it only finished successfully when the unpack pod had "Back-off pulling image".

Now, it finishes successfully if it finds an unpacked condition with any status/message?

go.mod
acornett21: Yeah, I just noticed they have their own mod files, which seems a strange choice that I've never noticed working with that project.

test/e2e/plain_provisioner_test.go
everettraven: Ah good catch - I went through and removed all the pod based tests. Seems like I missed that this was testing for a specific unpack state. I'll fix this

testdata/bundles/plain-v0/dependent/build-push-e2e-bundle.sh
joelanford: Is there a way to reduce down to a single script and make the specific nuances encapsulated with some other mechanism? For example, it seems like the nuances are in the shape of the bundles. Can we tar.gz the root directories and then have the script deal in tar.gz files generically. Configmap data is tar.gz, and the build/push pod is changed to untar prior to running the build?

everettraven: I'm sure we could do _something_ to make it one or two scripts (likely 2 for the pushing to a registry requiring auth), but what do we gain from that vs the current approach?

I understand the desire to reduce the repetition of the scripting but I also see value in having the ability to have each "bundle" specify how it should be built rather than trying to optimize for a single common script

test/e2e/plain_provisioner_test.go
everettraven: Fixed in https://github.com/operator-framework/rukpak/pull/874/commits/9bc11577670c2a52254ae026c64f0336f6e32117

go.mod
everettraven: bumped go-containerregistry version to 0.19.1 in https://github.com/operator-framework/rukpak/pull/874/commits/414a3791d5d936f7943f5f62fee3637af7010f91

testdata/bundles/plain-v0/dependent/build-push-e2e-bundle.sh
everettraven: As discussed via Slack I created https://github.com/operator-framework/rukpak/issues/885 to track this as a follow up to this PR

joelanford: > but what do we gain from that vs the current approach?
Simplicity

In rukpak, a bundle is an `fs.FS`. So given that we're going for consistency and simplicity, it seems reasonable for us to have a Dockerfile like this that we can re-use for any bundle.

```Dockerfile
FROM scratch
ARG destBundleRoot=/
ADD bundle.tar.gz $destBundleRoot
````

And then the configmap would contain this Dockerfile entry and a `bundle.tar.gz` entry, which we could mount into a pod to do a generic build.

varshaprasad96 (approved):
This looks good to me, given we are going to discuss https://github.com/operator-framework/rukpak/pull/874#discussion_r1597071250 and https://github.com/operator-framework/rukpak/pull/874#discussion_r1592945966 in follow ups!

Thanks for working on this, @everettraven!
/lgtm

PR #876, opened 2024-05-02
Export rukpak functionality

tmshort:
In anticipation of using rukpak as a library, export some of the functionality. This primarilly meant moving files from `internal` to `pkg`.


pkg/convert/registryv1.go
```diff
@@ -291,6 +291,14 @@ func Convert(in RegistryV1, installNamespace string, targetNamespaces []string)
 	}
 
 	objs := []client.Object{}
+
+	// Add the install namespace
+	ns := &corev1.Namespace{
+		TypeMeta:   metav1.TypeMeta{Kind: "Namespace", APIVersion: "v1"},
+		ObjectMeta: metav1.ObjectMeta{Name: installNamespace},
+	}
+	objs = append(objs, ns)
```
joelanford, line 300: I _just_ removed this in the `InstallNamespace` PR. The expectation is that the install namespace will already exist.

tmshort: ?I think that wasn't true in our PoC, but it means we'll have to deal with it better during test.

joelanford: It wasn't. This change just merged yesterday in rukpak and today in operator-controller. If you're able to rebase to operator-controller `main`, you shouldn't need this.

joelanford (approved):

PR #877, opened 2024-05-02
Makefile: avoid re-evaluating vars unnecessarily; other minor fixup

### LINKED ISSUE #875: Improve Makefile to avoid unnecessarily shelling out multiple times for the same variables
See https://github.com/operator-framework/operator-controller/pull/813

joelanford:
Fixes #875

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -11,7 +11,6 @@ builds:
   - id: unpack
     main: ./cmd/unpack
     binary: unpack
-    tags: "{{.Env.GO_BUILD_TAGS}}"
     goos:
     - linux
     goarch:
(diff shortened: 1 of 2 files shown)


grokspawn:
I'm still not 100% sold on replacing ?= statements with the likes of
```Makefile
ifeq ($(origin IMAGE_REPO), undefined)
include .bingo/Variables.mk
IMAGE_REPO := quay.io/operator-framework/rukpak
endif
```

since according to gnu make these are equivalent.  Any other behavior should be considered a bug in the tool. 

However, as with the related PRs there are lots of other improvements which tangibly improve make interactions, so I'm content to agree-to-disagree on this point. 😄

grokspawn (approved):

PR #879, opened 2024-05-03
Pod Security

dtfranz:
Change bundle unpack pod security settings to allow running in restricted namespaces. Change Dockerfile to use debug-nonroot image as builder only because we only need the cp binary from it.


pkg/source/image.go
```diff
@@ -138,13 +138,12 @@ func (i *Image) getDesiredPodApplyConfig(bundle *rukpakv1alpha2.BundleDeployment
...
 				WithImage(i.UnpackImage).
 				WithImagePullPolicy(corev1.PullIfNotPresent).
-				WithCommand("cp", "-Rv", "/unpack", "/util/bin/unpack").
+				WithCommand("/cp", "-Rv", "/unpack", "/util/bin/unpack").
 				WithVolumeMounts(applyconfigurationcorev1.VolumeMount().
 					WithName("util").
 					WithMountPath("/util/bin"),
 				).
-				WithSecurityContext(containerSecurityContext).
-				WithTerminationMessagePolicy(corev1.TerminationMessageFallbackToLogsOnError),
```
tmshort, line 147: I thought @joelanford had just put these in? Why are they being removed?

tmshort: https://github.com/operator-framework/rukpak/pull/872

dtfranz: They got stomped from the file that was in op-con, fixing now :+1:

varshaprasad96:
e2e is failing because of change in base image. Looks like we are trying to list something:
```
exec: \"ls\": executable file not found in $PATH: unknown"
```

dtfranz:
> e2e is failing because of change in base image. Looks like we are trying to list something:
> 
> ```
> exec: \"ls\": executable file not found in $PATH: unknown"
> ```

I'll just add `ls` from the base image for now as we do for `cp` if you're OK with that.

varshaprasad96 (approved):
The changes look good to me!
/lgtm 
/approve

PR #884, opened 2024-05-10
:sparkles: [Add] Coverage dir for unpack pod

varshaprasad96:
This PR adds the go test coverage variable to the
unpack pod. It also creates necessary volumes
if the GOCOVERDIR env var is set.

### DIFF
--- a/pkg/source/image.go
+++ b/pkg/source/image.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"io/fs"
+	"os"
 	"strings"
 
 	"github.com/nlepage/go-tarfs"
@@ -112,6 +113,12 @@ func (i *Image) getDesiredPodApplyConfig(bundle *rukpakv1alpha2.BundleDeployment
 	// this PSA definition either configurable or workable in a restricted namespace.
 	//
 	// See https://github.com/operator-framework/rukpak/pull/539 for more detail.
+
+	// This should ideally be a persistent volume that remains even after the
+	// pod is re-created. Since unpack image process is about to change to use
+	// a registry client, and the test coverage procedure is to revisited, creating
+	// a temp solution that uses an empty dir.
+	gocoverdirEnv := os.Getenv("GOCOVERDIR")
 	containerSecurityContext := applyconfigurationcorev1.SecurityContext().
 		WithAllowPrivilegeEscalation(false).
 		WithCapabilities(applyconfigurationcorev1.Capabilities().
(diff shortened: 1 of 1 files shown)


pkg/source/image.go
```diff
@@ -139,28 +146,54 @@ func (i *Image) getDesiredPodApplyConfig(bundle *rukpakv1alpha2.BundleDeployment
...
 				WithImage(bundle.Spec.Source.Image.Ref).
 				WithCommand("/bin/unpack", "--bundle-dir", "/").
-				WithVolumeMounts(applyconfigurationcorev1.VolumeMount().
-					WithName("util").
-					WithMountPath("/bin"),
-				).
+				WithVolumeMounts(func() []*applyconfigurationcorev1.VolumeMountApplyConfiguration {
+					var volumeMounts []*applyconfigurationcorev1.VolumeMountApplyConfiguration
+					if gocoverdirEnv != "" {
+						volumeMounts = append(volumeMounts, applyconfigurationcorev1.VolumeMount().
```
varshaprasad96, line 172: A temporary fix to get e2e's running in operator controller!

pkg/source/image.go
```diff
@@ -139,28 +146,54 @@ func (i *Image) getDesiredPodApplyConfig(bundle *rukpakv1alpha2.BundleDeployment
...
-					WithName("util").
-					WithMountPath("/util/bin"),
-				).
+				WithVolumeMounts(func() []*applyconfigurationcorev1.VolumeMountApplyConfiguration {
+					var volumeMounts []*applyconfigurationcorev1.VolumeMountApplyConfiguration
+					if gocoverdirEnv != "" {
+						volumeMounts = append(volumeMounts, applyconfigurationcorev1.VolumeMount().
+							WithName("test-coverage").
+							WithMountPath(gocoverdirEnv))
+					}
```
joelanford, line 155: I don't think we need coverage in this container, right? There are no instrumented binaries here.

pkg/source/image.go
```diff
@@ -139,28 +146,54 @@ func (i *Image) getDesiredPodApplyConfig(bundle *rukpakv1alpha2.BundleDeployment
...
+					volumeMounts = append(volumeMounts, applyconfigurationcorev1.VolumeMount().
+						WithName("util").
+						WithMountPath("/util/bin"))
+					return volumeMounts
+				}()...).
+				WithEnv(applyconfigurationcorev1.EnvVar().WithName("GOCOVERDIR").WithValue(gocoverdirEnv)).
```
joelanford, line 161: Similar comment as above: this container doesn't need GOCOVERDIR set, right?

pkg/source/image.go
```diff
@@ -139,28 +146,54 @@ func (i *Image) getDesiredPodApplyConfig(bundle *rukpakv1alpha2.BundleDeployment
...
+						volumeMounts = append(volumeMounts, applyconfigurationcorev1.VolumeMount().
+							WithName("test-coverage").
+							WithMountPath(gocoverdirEnv))
+					}
+					volumeMounts = append(volumeMounts, applyconfigurationcorev1.VolumeMount().
+						WithName("util").
+						WithMountPath("/bin"))
+					return volumeMounts
+				}()...).
+				WithEnv(applyconfigurationcorev1.EnvVar().WithName("GOCOVERDIR").WithValue(gocoverdirEnv)).
```
joelanford, line 181: Can we only set this if `gocoverdirEnv != ""`?

pkg/source/image.go
varshaprasad96: That's right, init container doesn't need it. Reverted back.

everettraven (dismissed):

everettraven (dismissed):

joelanford (approved):

tmshort (dismissed):
/lgtm

pkg/source/image.go
varshaprasad96: Passing an empty configuration is causing errors. Tried various options, but looks like the only one is to create the entire container spec separately based on whether env var is present or not. 

Leaving it to be as is for now, if the env is not set it results in an empty string which doesn't cause a problem for now. Also this is just a band aid, and will replaced by direct registry client anyways.

tmshort (approved):

PR #902, opened 2024-05-21
Add optional field to BundleDeployment API to enable/disable CRD upgrade safety preflight checks 

rashmigottipati:
#### Description

Adds a new, optional field to the `BundleDeployment` API spec for enabling/disabling the CRD Upgrade Safety preflight checks. The default state should be enabled.

No logic included in this PR as the CRD Upgrade Safety preflight check as there isn't anything yet to enable/disable. So logic to enable/disable will be added in a follow-up

fixes https://github.com/operator-framework/operator-controller/issues/745
original PR in operator-controller: https://github.com/operator-framework/operator-controller/pull/845


api/v1alpha2/bundledeployment_types.go
```diff
@@ -69,8 +69,41 @@ type BundleDeploymentSpec struct {
...
+
+	//+kubebuilder:Optional
+	// Preflight defines the strategy for the preflight checks (i.e. as of now, the CRD upgrade safety checks) to be enabled or disabled when attempting to install the BundleDeployment.
+	Preflight *PreflightConfig `json:"preflight,omitempty"`
+}
+
+// PreflightConfig holds the config for the preflight checks. Currently, this is implemented for the
+// CRDUpgradeSafety preflight check and can be extended in the future to add other preflight checks.
+type PreflightConfig struct {
+	CRDUpgradeSafety *CRDUpgradeSafetyPreflightConfig `json:"crdUpgradeSafety,omitempty"`
```
everettraven, line 81: Do we need to add any of the kubebuilder validation tags to mark this field as optional as well?

rashmigottipati: Yes, good catch! Updated in [commit](https://github.com/operator-framework/rukpak/pull/902/commits/3c22f6db5cca72ff64366a7928dcc3b10ca66dc6)

everettraven (approved):
These changes look good to me.

I'll hold on merging this for a bit to give @joelanford a chance to review in case he has any additional comments

api/v1alpha2/bundledeployment_types.go
```diff
@@ -69,8 +69,42 @@ type BundleDeploymentSpec struct {
...
 	// config is provisioner specific configurations
 	Config runtime.RawExtension `json:"config,omitempty"`
+
+	//+kubebuilder:Optional
+	// Preflight defines the strategy for the preflight checks (i.e. as of now, the CRD upgrade safety checks) to be enabled or disabled when attempting to install the BundleDeployment.
+	Preflight *PreflightConfig `json:"preflight,omitempty"`
+}
+
+// PreflightConfig holds the config for the preflight checks. Currently, this is implemented for the
+// CRDUpgradeSafety preflight check and can be extended in the future to add other preflight checks.
```
joelanford, line 79:
```suggestion
// PreflightConfig holds the config for the preflight checks.
```

This comment makes it into the CRD API docs. The "Currently ...." bit is developer doc that shouldn't show up there.

api/v1alpha2/bundledeployment_types.go
```diff
@@ -69,8 +69,42 @@ type BundleDeploymentSpec struct {
 	//
 	// config is provisioner specific configurations
 	Config runtime.RawExtension `json:"config,omitempty"`
+
+	//+kubebuilder:Optional
+	// Preflight defines the strategy for the preflight checks (i.e. as of now, the CRD upgrade safety checks) to be enabled or disabled when attempting to install the BundleDeployment.
```
joelanford, line 74:
```suggestion
	// Preflight defines the configuration of preflight checks.
```

api/v1alpha2/bundledeployment_types.go
```diff
@@ -69,8 +69,42 @@ type BundleDeploymentSpec struct {
...
+	Preflight *PreflightConfig `json:"preflight,omitempty"`
+}
+
+// PreflightConfig holds the config for the preflight checks. Currently, this is implemented for the
+// CRDUpgradeSafety preflight check and can be extended in the future to add other preflight checks.
+type PreflightConfig struct {
+	//+kubebuilder:Optional
```
joelanford, line 81: Add description GoDoc here about `CRDUpgradeSafety`?

api/v1alpha2/bundledeployment_types.go
```diff
@@ -69,8 +69,42 @@ type BundleDeploymentSpec struct {
...
+	CRDUpgradeSafety *CRDUpgradeSafetyPreflightConfig `json:"crdUpgradeSafety,omitempty"`
+}
+
+// CRDUpgradeSafetyPreflightConfig is a custom struct that holds the necessary configuration values to
+// enable or disable the CRD Upgrade Safety validations. Currently, this holds Mode
+// that sets the CRD Upgrade Safety validations to either Enabled/Disabled.
+// By default, this is set to "Enabled".
```
joelanford, line 88:
```suggestion
// CRDUpgradeSafetyPreflightConfig is the configuration for CRD upgrade safety preflight check.
```

api/v1alpha2/bundledeployment_types.go
```diff
@@ -69,8 +69,42 @@ type BundleDeploymentSpec struct {
...
+type CRDUpgradeSafetyPreflightConfig struct {
+	//+kubebuilder:validation:Enum:=Enabled;Disabled
+	//+kubebuilder:default:=Enabled
+	//+kubebuilder:Optional
+	Mode *CRDUpgradeSafetyMode `json:"mode,omitempty"`
```
joelanford, line 93: Do we envision add more modes here (e.g. to enable/disable certain facets of the check)? Or are we thinking that would end up going into a separate field or fields?

joelanford: Awhile back with `.spec.paused`, I flip-flopped on "to bool, or not to bool" and I ended up feeling like a bool actually worked better. I wonder if that's the case here as well.

joelanford, line 93: `Mode` should have it's own description comment. This is where we can talk about `Enabled` being the default value.

everettraven: I was thinking to keep this an enum with the idea that this could be used for modes like Enabled, Disabled, Strict, Loose.

We had originally discussed a potential in the future to have designated Strict/Loose options but if we don't think that is something that would be worth pursuing in the long run then I would be okay swapping this back to a bool

rashmigottipati: > Do we envision add more modes here (e.g. to enable/disable certain facets of the check)? Or are we thinking that would end up going into a separate field or fields?

I think this would end up going into a separate field or a set of fields. 
Also, I renamed `Mode` to `Enabled` and it's a bool now - that either enables/disables the entire preflight check.

everettraven (dismissed):

api/v1alpha2/bundledeployment_types.go
```diff
@@ -69,6 +69,26 @@ type BundleDeploymentSpec struct {
...
+type PreflightConfig struct {
+	//+kubebuilder:Optional
+	// CRDUpgradeSafety holds necessary configuration for the CRD Upgrade Safety preflight checks.
+	CRDUpgradeSafety *CRDUpgradeSafetyPreflightConfig `json:"crdUpgradeSafety,omitempty"`
+}
+
+// CRDUpgradeSafetyPreflightConfig is the configuration for CRD upgrade safety preflight check.
+type CRDUpgradeSafetyPreflightConfig struct {
+	//+kubebuilder:default:=true
+	//+kubebuilder:Optional
```
joelanford, line 88: Let's make this required. Since the parent struct is optional, it would only make sense to specify the parent struct if you intended to explicitly specify something in the child struct.

Looking at this more closely, I think this rule should apply all the way up the stack.

So `spec.preflight` is optional, but if you specify it, the only choice right now is `spec.preflight.crdUpgradeSafety` (so let's make that required for now as well).

WDYT?

joelanford: But let's leave `CRDUpgradeSafety *CRDUpgradeSafetyPreflightConfig` as a pointer since the expectation is that we may add configurations for other preflight checks in the future.

rashmigottipati: I totally agree with this. It doesn't make sense to specify the parent struct without explicitly specifying something in the child struct. It would make it more semantically correct to apply this rule all the way up the stack.

rashmigottipati: Updated in [latest commit](https://github.com/operator-framework/rukpak/pull/902/commits/c7eb026d9d8bbbe3f06a830fe91ce7c0d5bfc4e4)

joelanford (approved):

PR #913, opened 2024-06-12
Add install mode validation

m1kola:
New validation only allows `AllNamespaces` install mode to be used.

Related to https://github.com/operator-framework/operator-controller/issues/816


m1kola:
Closing this for now as we try to keep convertor generic in case we want to support other install modes too.

PR #920, opened 2024-06-14
Update certificate support

tmshort:
Remove `rootCAs` from the `NewDefaultUnpacker` API, the argument is no longer used for HTTP transport.

Add `CertificateData` to the ImageSource struct. This is PEM-encoded data (straight from a Secret[tls.crt]) to be used to validate the certificate used to access an image regidstry (works along side the `InsecureSkipTLSVerify` option).

### DIFF
--- a/api/v1alpha2/bundle_types.go
+++ b/api/v1alpha2/bundle_types.go
@@ -67,6 +67,8 @@ type ImageSource struct {
 	// This should not be used in a production environment.
 	// +optional
 	InsecureSkipTLSVerify bool `json:"insecureSkipTLSVerify,omitempty"`
+	// CertificateData contains the PEM data of the certificate that is to be used for the TLS connection
+	CertificateData string `json:"certificateData,omitempty"`
 }
 
 type GitSource struct {
--- a/cmd/core/main.go
+++ b/cmd/core/main.go
@@ -199,7 +199,7 @@ func main() {
 		os.Exit(1)
 	}
 
-	unpacker, err := source.NewDefaultUnpacker(mgr, systemNamespace, unpackCacheDir, rootCAs)
+	unpacker, err := source.NewDefaultUnpacker(mgr, systemNamespace, unpackCacheDir)
 	if err != nil {
 		setupLog.Error(err, "unable to setup bundle unpacker")
 		os.Exit(1)
--- a/cmd/helm/main.go
+++ b/cmd/helm/main.go
@@ -189,7 +189,7 @@ func main() {
 		os.Exit(1)
 	}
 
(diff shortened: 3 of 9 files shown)


tmshort:
This is necessary to get certs work on a per-`ClusterExtension` basis for an image registry. The `CertifcateData` is PEM to avoid rukpak having to look up secrets, etc. The creator of the `BundleDeployment` needs to do the lookup, etc. The default is to use system provided CAs, and those _are_ included even if `CertificateData` is specified.

tmshort:
The code is tested in the e2e, but that doesn't count toward code coverage.

api/v1alpha2/bundle_types.go
```diff
@@ -67,6 +67,8 @@ type ImageSource struct {
 	// This should not be used in a production environment.
 	// +optional
 	InsecureSkipTLSVerify bool `json:"insecureSkipTLSVerify,omitempty"`
+	// CertificateData contains the PEM data of the certificate that is to be used for the TLS connection
+	CertificateData string `json:"certificateData,omitempty"`
```
varshaprasad96, line 71: Wondering if instead of accepting as a string - should we instead let users references a secret directly here. I understand that this a BD spec, so we could technically get a secret in CE, process it and send it here - but probably processing it in the unpacking code makes it easier, rather than polluting the CE reconciler?

tmshort: I thought about that, but that potentially meant pushing more arguments down into rukpak, and adding more code to rukpak, which will likely change (i.e. be pulled into operator-controller) in the future. We don't know how this interface will change, so I wrote this to accept a PEM which is literally just thrown into another API. This should mean less code to modify when pulling in rukpak. 
I don't really consider it "polluting" the CE reconciler; the same work would have to be done within the reconciler in some way, shape or form.

tmshort:
(As of right now, my changes are not working on the latest operator-controller main... so I need to get that working again.)

api/v1alpha2/bundle_types.go
varshaprasad96: Yeah, totally understand. So basically CE would instead create a spec that could look something like this?:

```
spec:
     connectionConfig:
                 secretRef:
                           name:
                 insecureSkipTLSVerify:
```

We immediately have to make this change in o-c. So why not make the API look consistent here too is what I'm thinking.

tmshort: Again, it requires adding complexity to the function-based API that is not necessary, and since rukpak is going to be pulled into operator controller; we likely won't be using the bundleDeployment API any more, so why add the effort into rukpak?

varshaprasad96: Resolving this based on offline discussion

PR #922, opened 2024-06-17
Add tests to make sure rukpak enforces limitations

m1kola:
Currently we do not allow:
* Bundles with webhook definitions
* Bundles which provide API Service definitions


varshaprasad96 (approved):
/lgtm
/approve

PR #940, opened 2024-08-05
WIP: (feat) dump CSV annotations into chart metadata

gallettilance:
Tried to do this in a way that wouldn't change the API at all. This is needed for the cluster olm operator to find the ocp max version


perdasilva:
Moved to https://github.com/operator-framework/operator-controller/pull/1095

